PR #6, opened 2016-04-15
GenWayAreaDat: Fix bugs

LeonYang5114:
There are two bugs that this PR fixed:

1. The way areas are written to file when either the rawWays or the nodeIds fill the block size, or the last raw way is reached. However, the last raw way might not be an way area (i.e. IsArea() returns false, or is in blacklist). If so, all the way areas added after the last write will be lost. Therefore, flushing the remaining way areas should be done outside the loop.

2. If strict area is set to true, and the first ring of the area is not simple, then the program will always give seg fault, because the area.GetType() calls the GetType() of its first ring, and in the condition mentioned above, its first ring is NULL.


Framstag:
Thanks for this bug report! These problems were up to know unnoticed to me :-/

PR #7, opened 2016-04-27
Pedantic warnings

Karry:
These patches adding pedantic warning to CMake build (when compiler is GCC or Clang) - it can show possible problems in code. Second commit fix some of these warnings...


Framstag:
Thanks for the fixes!

PR #9, opened 2016-05-01
fix mercator projection (bug #17)

Karry:
This MR fixes bug #17 ( https://sourceforge.net/p/libosmscout/bugs/17/ )

This bug is visible when we render test grid from parallels and meridians or just grid for OSM tiles. Grid shouldn't change its dimensions when moving map to north or south.
This change breaks meterInPixel and meterInMM variables, they are valid
on equator and should be recompute for latitude. Should we crate new function for it?

### DIFF
--- a/libosmscout/src/osmscout/util/Projection.cpp
+++ b/libosmscout/src/osmscout/util/Projection.cpp
@@ -126,9 +126,9 @@ namespace osmscout {
     }
 
     // Resolution (meter/pixel) of a pixel in a classical 256 pixel tile for the given zoom level
-    double resolution=tileSizeZoom0Aquator/256*cos(lat*gradtorad)/magnification.GetMagnification();
+    double equatorResolution=tileSizeZoom0Aquator/256/magnification.GetMagnification();
 
-    double groundWidthMeter=width*tileDPI/dpi*resolution;
+    double groundWidthMeter=width*tileDPI/dpi*equatorResolution;
 
     //
     // Calculation of bounds and scaling factors
--- a/libosmscout/src/osmscout/util/Projection.cpp
+++ b/libosmscout/src/osmscout/util/Projection.cpp
@@ -200,10 +200,10 @@ namespace osmscout {
     std::cout << "Magnification: " << magnification.GetMagnification() << "/" << magnification.GetLevel() << std::endl;
     std::cout << "Screen dimension: " << width << "x" << height << " " << dpi << " DPI " << std::endl;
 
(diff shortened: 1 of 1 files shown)


Framstag:
Hello, I have tried to improve your code further and have now the following code fragment:

```C++
  //< Radius of the earth in meter
  static const double earthRadiusMeter=6378137.0;
  //< Extent of the earth in meter
  static const double earthExtentMeter=2*M_PI*earthRadiusMeter;
  //< Width of a tile at the equator for zoom level 0 in meter (equal to extent of the earth at the equator
  static const double tileWidthZoom0Aquator=earthExtentMeter;
  //< DPI of a classical OSM tile
  static const double tileDPI=96.0;
```
...

```C++
    // Width in meter of a tile of the given magnification at the equator
    double equatorTileWidth=tileWidthZoom0Aquator/magnification.GetMagnification();

    // Resolution (meter/pixel) of a pixel in a classical 256 pixel tile for the given zoom level at the equator
    double equatorTileResolution=equatorTileWidth/256.0;

    // Modified resolution (meter/pixel) at the equator based on our actual DPI instead of the standard tile DPI
    double equatorCorrectedEquatorTileResolution=equatorTileResolution*tileDPI/dpi;

    // Width of the visible area at the equator
    double groundWidthEquatorMeter=width*equatorCorrectedEquatorTileResolution;

    // Resulting projection scale factor
    scale=width/(2*M_PI*groundWidthEquatorMeter/earthExtentMeter);
    scaleGradtorad=scale*gradtorad;

    // Width of the visible area in meter
    double groundWidthVisibleMeter=groundWidthEquatorMeter*cos(lat*gradtorad);

    //
    // Calculation of bounds and scaling factors
    //
    // We have three projections:
    // * Mercator projection of longitude to X-coordinate
    // * Mercator projection of latitude to Y-coordinate
    // * Projection of X and Y coordinate as result of mercator projection to on screen coordinates
    //

    // Size of one pixel in meter
    pixelSize=groundWidthVisibleMeter/width;

    // How many pixel are one meter?
    meterInPixel=1/pixelSize;

    // 1 meter on the ground is how many millimeter on display?
    meterInMM=meterInPixel*25.4/pixelSize;
```

I see that the scale factor has changed (this had to be expected) and this again has some influence on the style sheet (things got more dense). This can be corrected, but I would like to make sure, that I now have "the right" code. It also seems like the calculation for meterInMM is complete false. At least I get very strange values. I had though not yet time to take a look at this aspect. 

Please double check my code, it looks like you know what I'm doing ;-)

Karry:
It looks fine for me. Thanks. I updated MR with your changes.

PR #10, opened 2016-05-11
fix libosmscout-map-agg build on Ubuntu:16.04

Karry:
Agg library is distributed as static library in two variants on Ubuntu:
 - position dependent objects: /usr/lib/libagg.a
 - position independent ( -fPIC ) objects: /usr/lib/libagg_pic.a

Because libosmscout-map-agg is compiled as position independent,
we should search second variant of libagg first.

This patch fixes following error:

[100%] Linking CXX shared library libosmscout_map_agg.so
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/5/../../../../lib/libagg.a(agg_vcgen_contour.o): relocation R_X86_64_32S against `.rodata' can not be used when making a shared object; recompile with -fPIC

### DIFF
--- a/cmake/FindAgg.cmake
+++ b/cmake/FindAgg.cmake
@@ -23,7 +23,7 @@ FIND_PATH(LIBAGG_INCLUDE_DIRS
 )
 
 FIND_LIBRARY(LIBAGG_LIBRARIES
-    NAMES libagg agg libagg-2.5 agg-2.5
+    NAMES agg_pic libagg agg libagg-2.5 agg-2.5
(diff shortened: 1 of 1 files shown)


Framstag:
Do you also have a fix for the autoconf based build?

Karry:
It seems that autoconf detects agg library correctly

libosmscout-map-agg$ ./configure
...
checking our pkgconfig ext libs... -losmscoutmapagg -lagg_pic -lfreetype
...
and make ends without error.

PR #12, opened 2016-05-18
don't allocate value buffer in FeatureValueBuffer if it isn't necessary

Karry:
Hi, I inspect libosmscout memory usage (with pprof) and found that allocations from FeatureValueBuffer occupy around 25% heap data (76MB from 315MB in one of my samples). On deeper look, I found that many (around 16MB) of allocated value buffers are never used! You can see it when you compile my "print-unused-value-buffer" branch:

https://github.com/karry-space-with-my-second-forks/libosmscout/tree/print-unused-value-buffer

As a simple solution, I allocate value buffer lazy. But now exist some danger in GetValue method. It can return NULL. But I don't found any unchecked usage.

### DIFF
--- a/libosmscout/include/osmscout/TypeConfig.h
+++ b/libosmscout/include/osmscout/TypeConfig.h
@@ -956,7 +956,14 @@ namespace osmscout {
 
   private:
     void DeleteData();
-    void AllocateData();
+    void AllocateBits();
+    void AllocateValueBufferLazy();
+
+    inline FeatureValue* GetValueAndAllocateBuffer(size_t idx)
+    {
+      AllocateValueBufferLazy();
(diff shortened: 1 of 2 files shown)


Karry:
btw, here are two pprof snapshots converted to graph:
pprof-1-315mb
pprof-2-496mb

You can found pprof documentation here: https://github.com/gperftools/gperftools/tree/master/doc , you should read page heapprofile.html

Framstag:
Sorry for my late answer, I was away over the weekend.

One question. Is it possible that the same effect could be archived if a feature value buffer is only allocated if type->GetFeatureValueBufferSize()>0 (which means there is at least one feature that possibly could have a value)? This would be far simpler to check and handle.

Though while I think about, even if a Feature could have a value in principle there are still instances where the value is not set, because it is equal to the default or the feature is not set.

PR #14, opened 2016-05-24
Optimize memory usage of Point sequence (vector) in Way and Area

Karry:
Hi, my target device (Jolla phone) has limited amount of memory (1 GiB). When some application on foreground occupy more that ~300 MiB, kernels "lowmemorykiller" start killing background apps. Sometimes when foreground app occupy ~500 MiB and system starts email synchronization for example, even foreground application is convicted to die. For this reasons, I want to keep OSMScout (heap) memory usage < 300 MiB. Remaining available space can be used for various caches then.

My experiments with memory profiler shows that *std::vector<Point>* in Way and Area can occupy significant amount of memory:
https://cloud.githubusercontent.com/assets/309458/15384023/bbd950be-1d97-11e6-9bc5-6b5803b4f073.png

But these Point sequences just contains data read from memory mapped database files. So my idea is to create new type *PointSequence* and use it instead of std::vector<Point>. PointSequence can contains just a pointer to memory mapped file and everything else can be evaluated lazy, with minimum of memory allocations. See initial implementation for more details. What do you think about it?

This concept should be good for sequential access but can be slower with random reads. But most parts of library already using iterator-like access to PointSequence. My initial memory measurements looks great. Code itself needs to be cleaned up little bit, I think that it is not ready for merging. But I will welcome every feedback and help.


Framstag:
I'm not yet convinced of this patch, though some parts of the changes could/should get integrated.

* I do see memory consumption as crucial. So if the library consumes too much memory this has to be fixed.
* In your measurements (if I read the image correctly) the most memory is hold by optimized areas via a list. For me this looks like you are just read a rather huge amount of optimized areas - which of course will result in a huge amount of vector instances. This is a little bit strange, since there should be only a few optimized areas and these should only have a few nodes. Did you analyse this before? Some simple test shows, that I access up to 4-5000 optimized areas (and 2-3000 ways) for a rendering. Are you in the same range? Bigger? It would more sense, to first analyse if the amount of areas could not be reduced (are all 4-5000 areas actually visible, should be not drop to small areas?). It would also make sense to check how much node each area has and if a more aggressive reduction might make more sense. Low level optimizations are OK but only if algorithmic optimizations are already at their limit.
* Memory mapped files require that the sum of the size of all memory mapped files can be mapped into the OS virtual memory. For 32 bit systems this is 4GB. The importer crashes for the same reasons under Linux if process memory reaches around 3,7GB. I do not know how many memory mapped files you can actual open (and hold open at the same time) on iOS or Android. You I would not rely on memory mapped files being available.
* Memory mapped files would be of excellent use of if file representation and memory representation do match. In this case memory mapped files make thing definitely easier. This is not the case for libosmscout because I did the design decision that I trade smaller on disk file size and platform independent format in favour of data reading size (we could reduce the CPU demand and would be able to load bigger data blocks into memory if we would do same changes to the file data model). This could be discussed but I assume that resulting file size would increase by factors. As far as I can see, you fall back to normal vectors if the file is not memory mapped.
* You assume that the file is always open. I fear you will get very strange (catchable?) errors, if the file has been close in the meanwhile?
* What is the impact on the import time? This will also give some hint how high are the cost by trading low memory footprint in favour to CPU an disk IO. Did the rendering on the mobile device take longer?
* You are also moving much data in to the the PointSequence, which again results in some memory footprint.
* I have no problem with dropping standard containers and actually have though of this in the past - if the custom container behaves like a standard one. For example range based loop could be used.
* I also did see the problem with allocation many objects. For me the potential problem however was more the number of (small) objects and the resulting memory overhead. So the question would be: How much would be gain, if we just reduce the number of allocated objects? How could be reduce the number of objects? The solution would be to even more store data in tiles/cells/blocks and always load these in whole. This way I could use one big buffer for all coordinate "strips" in the tile and just address this by offset and count.
* Your solution is more expensive (CPU, IO and overall time wise), if points are visited more than once, because it has to be "loaded" every time again. Currently data is cached in the MapService. So the renderer scan data again and again. Rendering should get more expensive for your solution this way. I plan to move more caching into the renderer. I'm not sure if as a result the render must visit (all) data again after initially processing a tile. On the other side the renderer will definitely have a different data layout then actual raw data, it would not benefit from the PointSequence.

As you see, the patch would be a bigger change and I'm not yet sure if it would move things is the right direction. It definitely make some parts of the code more difficult. It would help if you give me some feedback to better judge, since I do not want to through away you changes light-heartly.

In a result, I plan to analyse the number of optimized areas to see if this can get reduced. This would solve the problem much more elegantly.

Framstag:
I'm not yet convinced of this patch, though some parts of the changes could/should get integrated.

* I do see memory consumption as crucial. So if the library consumes too much memory this has to be fixed.
* In your measurements (if I read the image correctly) the most memory is hold by optimized areas via a list. For me this looks like you are just read a rather huge amount of optimized areas - which of course will result in a huge amount of vector instances. This is a little bit strange, since there should be only a few optimized areas and these should only have a few nodes. Did you analyse this before? Some simple test shows, that I access up to 4-5000 optimized areas (and 2-3000 ways) for a rendering. Are you in the same range? Bigger? It would more sense, to first analyse if the amount of areas could not be reduced (are all 4-5000 areas actually visible, should be not drop to small areas?). It would also make sense to check how much node each area has and if a more aggressive reduction might make more sense. Low level optimizations are OK but only if algorithmic optimizations are already at their limit.
* Memory mapped files require that the sum of the size of all memory mapped files can be mapped into the OS virtual memory. For 32 bit systems this is 4GB. The importer crashes for the same reasons under Linux if process memory reaches around 3,7GB. I do not know how many memory mapped files you can actual open (and hold open at the same time) on iOS or Android. You I would not rely on memory mapped files being available.
* Memory mapped files would be of excellent use of if file representation and memory representation do match. In this case memory mapped files make thing definitely easier. This is not the case for libosmscout because I did the design decision that I trade smaller on disk file size and platform independent format in favour of data reading size (we could reduce the CPU demand and would be able to load bigger data blocks into memory if we would do same changes to the file data model). This could be discussed but I assume that resulting file size would increase by factors. As far as I can see, you fall back to normal vectors if the file is not memory mapped.
* You assume that the file is always open. I fear you will get very strange (catchable?) errors, if the file has been close in the meanwhile?
* What is the impact on the import time? This will also give some hint how high are the cost by trading low memory footprint in favour to CPU an disk IO. Did the rendering on the mobile device take longer?
* You are also moving much data in to the the PointSequence, which again results in some memory footprint.
* I have no problem with dropping standard containers and actually have though of this in the past - if the custom container behaves like a standard one. For example range based loop could be used.
* I also did see the problem with allocation many objects. For me the potential problem however was more the number of (small) objects and the resulting memory overhead. So the question would be: How much would be gain, if we just reduce the number of allocated objects? How could be reduce the number of objects? The solution would be to even more store data in tiles/cells/blocks and always load these in whole. This way I could use one big buffer for all coordinate "strips" in the tile and just address this by offset and count.
* Your solution is more expensive (CPU, IO and overall time wise), if points are visited more than once, because it has to be "loaded" every time again. Currently data is cached in the MapService. So the renderer scan data again and again. Rendering should get more expensive for your solution this way. I plan to move more caching into the renderer. I'm not sure if as a result the render must visit (all) data again after initially processing a tile. On the other side the renderer will definitely have a different data layout then actual raw data, it would not benefit from the PointSequence.

As you see, the patch would be a bigger change and I'm not yet sure if it would move things is the right direction. It definitely make some parts of the code more difficult. It would help if you give me some feedback to better judge, since I do not want to through away you changes light-heartly.

In a result, I plan to analyse the number of optimized areas to see if this can get reduced. This would solve the problem much more elegantly.

Karry:
Too many questions :-) We definitely need some reproducible and automated test for memory consumption. I will try to extend PerformanceTest.cpp by memory measurement (for Linux for the beginning) and prepare some statistics next week. 

Meanwhile just few notes: 
 * Memory mapped PointSequence expect that file is open (mapped) all time when sequence is used. Exceptions are not handled now.
 * I know that mmap is limited to < 4 GiB on 32 bits systems (usual arm phones), so vector based implementation is used as fallback.
 * I dont measure performance impact (rendering or import) yet, but I don't feel any rendering slowdown on my phone. Moreover, It is more stable because lowmemorykiller have no work.

Framstag:
Having more reproducable and automated tests - yes, there we definitely agree :-) The problem with all automated tests for OSM is, that the data is constantly changing. One would need a self-made *.osm or *.osm.pbf file to make sure that the test base data does not change. Freezing an special OSM export for a region and a date is an option, but makes it difficult to change the data if you want to add additional test cases.

I would ask you to check if my checkin fixing calculation of optimized areas does ease your memory consumption problems. In either case it would be helpful to make a more in deep analysis of the data before applying a "big strike" solution.

I do make some memory consumption check using valgrind, these did not show that massive memory consumption. Likely because the bug was introduced later. Automatic test cases would sure help here and this is one of the reason why I started implementing automatic builds as a base for further work.

Karry:
Hi. Your recent fixes improves database size and memory footprint too. Czech republic database occupy 464 MB on disk now. It was 563 MB before fix. 

I used improved PerformanceTest ([MR 16](https://github.com/Framstag/libosmscout/pull/16)) to measure memory footprint. I used following arguments:
```
./build/Demos/PerformanceTest \
  $database \
  ./stylesheets/standard.oss \
  50.088 14.418 50.086 14.421 \
  0 20 \
  256 256 \
  Qt
```
...used area is Prague center that contains many building polygons and street ways.

I tested two library branches (master and my point-sequence) with old and new database. [See attached logs](https://github.com/Framstag/libosmscout/files/293930/perf-logs.zip). New database is better with both implementations. Memory usage is worst with maximum zoom level (20) and it is smaller with point-sequence (231 MB vs. 196 MB). Point sequence implementation is even faster! I thing that it is caused by caching of GeoBox (not measured). I think that cost of m-mapped memory (complications) is too high for 15% of memory improvement. What do you think?

I will try to find more elegant replacement of `vector<Point>`.

Framstag:
Hi. I do not think that the around 10% memory improvement do not justify the far more complex approach. IT is possible that these 10% could for example already be saved by a better file format for some datafiles or by more aggressive optimizations during import. Those improvements would have no impact on the actual loading code.

An indicator for the "quality" of the individual data files could be the compression factor of e.g. the zip tool. I also recently though about compressing individual files using zip, if zip is able to act as a transparent layer regarding file positioning (FileScanner of course would have to be changed).

I'm happy that my fix improved the memory situation. I however still wonder why the memory consumptions is maximum at zoom 20? The individual tile should contain less data as some tiles with lower zoom. As far as I remember the *number* of object peaks around levels 9-13 and sinks again (depending of the actual style sheet of course).

The repeated calculation of GeoBox is of still an interesting optimization. Where do the repeated call come from. MY plan is to allow the MapPainter to build up a (backend specific) cache for the tile data. Here it store for example bounding box sizes, label sizes etc... That will increase memory but will reduce (re)render time. Would this possibly a place where the GeoBox could get cached, too?

Karry:
> I however still wonder why the memory consumptions is maximum at zoom 20
I will try to evaluate it. When I start test with zoom level 19~20, after last tile heap occupy 151 MB. But when I repeat test just for zoom level 20, it occupy only 141 after last tile. It don't makes sence for me. (Test load 110 tiles on level 20 and cache size is default: 25.) It looks like some memory leak, but I can't see it with valgrind.

> Where do the repeated call come from.
Look commit: https://github.com/Framstag/libosmscout/pull/14/commits/0d2f3e9e912b30e9de8425491ac0fe8728384505
`PointSequence::bbox()` is called more than once for each rendering... For example from `MapPainter::PrepareAreas` is called by method `MapPainter::IsVisibleArea` and then (if area is visible) by `Area::Ring::GetBoundingBox`

> Would this possibly a place where the GeoBox could get cached, too?
I am not sure. Cache label size and position makes sence for me, but I can't imagine how backend specific cache should look like.

Karry:
Hi Tim. 

I updated rest of modules to work with `PointSequence` container. Here are updated results with fixed [memory leak](https://github.com/Framstag/libosmscout/commit/dfc5cd919608eb4d63d47c9d8dae23c79f66806b):
 - Import process was not slowdown by PointSequence usage 
    - Czech republic import tooks 5 hours, 26 minutes with master, 5:24 with PointSequence branch
 - `PointSequence` has better results with `PerformanceTest`, see [scout-perf.zip](https://github.com/Framstag/libosmscout/files/311304/scout-perf.zip) logs. For example for zoom level 16:
    - avg. memory usage 164 MB -> 126 MB (after loading tile data)
    - avg. DB load 860 ms -> 411 ms
    - avg. Map Rednering 575 ms -> 314 ms

I don't have `PerformanceTest` results from my phone yet, because It don't contains some Qt components like `QDesktopWidget`. It require addition work to compile it.

Karry:
Added results from Jolla phone (ARMv7): [scout-perf2.zip](https://github.com/Framstag/libosmscout/files/313187/scout-perf2.zip). Memory consumption and database times are lower with PointSequence, like on desktop. But render performance is worse little bit...

Framstag:
I see the  principle reduction in memory usage, but my arguments regarding the implementation are IMHO still valid?

Reduction of memory usage allows us to make use of more memory at other points of the code, e.g. caching. But I'm just not that happy with the approach (which is partly subjective I admit). Do you have an idea for an implementation that has the positive effects without the (that much) negative effects?

What are the consequences? How to continue?

Karry:
Hi. I tried this memory mapped *PointSequence* on my Rasberry PI (model B+, armv6 cpu) and its performance was terrible - worse than with original *std::vector* implementation. So I can't suggest to merge it in current form. 

How to continue? I will try another solution (array of points in continuous memory?) that may have better performance and save some memory too...

Framstag:
I now entered issue #55 to draft how an initial proof of concept could work. I have assigned it to me, but I admit that I currently do not have too much time. So if you like to take that one (because you have more time), just tell me. There are enough issues for everbody on the list, and I have still more on my private TODO list ;-)

Karry:
I am closing this outdated MR... It helps with memory footprint, but solution is not elegant (rely and access to mem-mapped files from data objects). I will continue provide these patches for my app separately, but it should not be merged to upstream library...

Issue #22, opened 2016-06-15
Travis Linux GCC cmake build fails


Framstag:
Currently the Travis build for libosmsocut in the combination gcc/cmake/Linux fails:

```
...
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.

Please set them or make sure they are set and tested correctly in the CMake files:

GLUT_Xi_LIBRARY (ADVANCED)

    linked by target "osmscout_map_opengl" in directory /home/travis/build/Framstag/libosmscout/libosmscout-map-opengl

PROTOBUF_PROTOC_LIBRARY (ADVANCED)

    linked by target "osmscout_import" in directory /home/travis/build/Framstag/libosmscout/libosmscout-import
...
```

Either some installation packages are missing or the cmake check is not correct (cmake should only fail on required dependencies missing, optional dependencies missing should never make the build fail).

I need someone with more cmake knowledge to help me.

Framstag:
Fixed with the referenced pull request!

PR #26, opened 2016-06-18
Cmake fixes

Karry:
It should solve CMake/Linux Travis build https://github.com/Framstag/libosmscout/issues/22


Framstag:
GitHub/Travis/Appveyor do also build pull request. So you can already see, that this indeed fixes the Travis build (the Appveyor build though is still running) :-)

Thanks for this patch! I'll close the bug also.

PR #36, opened 2016-06-21
Projection improvements

Karry:
Hi. I see that `*Projection::GeoToPixel` method is relative expensive in callgrind results. It is better when SSE optimizations are enabled (TileProjection), but it is still one of hot spots in rendering code. It is even worse on slow arm processor without SIMD optimizations. So I prepared two new projections with simple optimization:
```
Vertical (y) coordinate [pixels] in Mercator projection is computed
as atanh(sin(latitude)). For small render areas with bigger magnification
can be this function aproximated by tangent in projection center. 
Horizontal coordinate computation can be simlified just to multiplication 
of latitude and scaled latitude fn. derivation.
It should be faster than `sin` and `atanh` computation.
```
I setup minimum magnification level for use this optimization to 10 now. Y coordinate inaccuracy should be around 1% (without math prove, tested by small amount of experiments). Error is slightly visible when magnification is setup to 10 and map is moved vertically...

Rendering for magnification level > 10 is about **5 ~ 10% faster** on armv7 and x86_64 (without SSE optimizations). You can see in *PerformanceTest* logs  from my phone (rendering was repeated 100x):
[scout-perf-projection.zip](https://github.com/Framstag/libosmscout/files/326013/scout-perf-projection.zip)

Btw, I will try to use SIMD optimizations on arm later...

### DIFF
--- a/Demos/src/PerformanceTest.cpp
+++ b/Demos/src/PerformanceTest.cpp
@@ -288,10 +288,10 @@ int main(int argc, char* argv[])
   }
 #endif
 
-  osmscout::TileProjection      projection;
-  osmscout::MapParameter        drawParameter;
-  osmscout::AreaSearchParameter searchParameter;
-  std::list<LevelStats>         statistics;
+  osmscout::ApproximateTileProjection   projection;
+  osmscout::MapParameter                drawParameter;
+  osmscout::AreaSearchParameter         searchParameter;
+  std::list<LevelStats>                 statistics;
 
   searchParameter.SetUseMultithreading(true);
 
--- a/Demos/src/PerformanceTest.cpp
+++ b/Demos/src/PerformanceTest.cpp
@@ -288,10 +288,10 @@ int main(int argc, char* argv[])
   }
 #endif
 
-  osmscout::ApproximateTileProjection   projection;
-  osmscout::MapParameter                drawParameter;
-  osmscout::AreaSearchParameter         searchParameter;
-  std::list<LevelStats>                 statistics;
+  osmscout::TileProjection      projection;
+  osmscout::MapParameter        drawParameter;
+  osmscout::AreaSearchParameter searchParameter;
+  std::list<LevelStats>         statistics;
 
   searchParameter.SetUseMultithreading(true);
 
@@ -360,6 +360,7 @@ int main(int argc, char* argv[])
                        tileHeight);
 
         projection.GetDimensions(boundingBox);
+        projection.SetLinearInterpolationUsage(level >= 10);
 
         osmscout::StopClock dbTimer;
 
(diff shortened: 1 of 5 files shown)


Framstag:
It does make sense to try to improve the projection code. I up to now  however keep my hand of this, because I have read the the standard c library already would do some optimisations (for example aproximate atanh by some polygon or by using SSE) and thus feared that there are no optimisations left.

Did you try some compiler flags, that might trigger such compiler optimisations first? I did try to check with general math flags but did not see any obvious improvements.

It is possible that SEE only improves calculation time if you are doing calculations in a batch. On the other hand - if your findings result in no improvement in this case - I would like to remove or simplify the batch handling code. The indirections are rather difficult to follow.

I also though about moving some parts of the calculation into some lookup table (if you do a polygon approximation perhaps all possible values of the first part could be precalculated) if this would help.

So all tricks are allowed as long as:
* The result is in principle correct ;-)
* You can switch them of if the native implementation is faster
* The upper error limit can be calculated and is OK (who cares for one pixel precision if you have 400DPI).

I would suggest to modify the code to move the optimisations into the Mercator Projection itself and trigger it via some additional constructor, if this does not result in measurable performance loss.

Having too many projection classes makes things more difficult than needed. Replacing the MercatorProjection in OSMScout2 for example is/was not as easy as it looks like.

OK?

Framstag:
Note that caching of the result will not help in general, if your projection result is not independent of the actual screen position. So one first optimisation should perhaps be to check if the mercator projection and the projection onto the screen can be separated. This way, as long as a tile is visible most parts of the calculation could be retrieved from cache.

Looking at  
```C++
  void MercatorProjection::GeoToPixel(const GeoCoord& coord, double& x, double& y) const
```

this might in fact be possible?

Note, that I plan to cache most of such calculations in a render based cache anyway...The question is, do we have enough memory for this?

Karry:
Ok, I moved approximated calculation to original projections and remove new ones. 
I disabled approximations by default, it should be enabled explicitly when you know what you are doing :) 
It makes sense on small areas, but when you try to calculate something bigger, mercator bounding box for example, you don't get square and it may lead to some weird bugs (I was affected by one)...

Btw, I render functions in KmPlot for beter imagination:
mercator-latitude-fn-derivation

Karry:
Removing batch processing is another topic. Yes, batching makes code more complicated, but please wait little bit with its removing. I plan to experiment with it. It may boost conversions when compiler gets change to vectorize loops and do computation on SIMD fpu. 

And renderer cache for computed coordinates, bounding boxes (...), it is another big task... Can you create issue for that and summarise your ideas how it may work?

Framstag:
Manually merged with small corrections.

PR #51, opened 2016-06-23
Add explicit 'using Projection::GeoToPixel' in derived projections

Karry:
Add explicit 'using Projection::GeoToPixel' in derived projections to avoid name hiding

It fixes compile error when:

    osmscout::MercatorProjection projection;
    double x;
    double y;
    projection.GeoToPixel(osmscout::GeoCoord(lat, lon), x, y);

fails with:

      error: no matching function for call to ‘osmscout::MercatorProjection::GeoToPixel(osmscout::GeoCoord, double&, double&)’

---

If I understand this C++ area correctly, implementation of `GeoToPixel(double lon, double lat ...` in derived classes (`MercatorProjection`) hides `GeoToPixel(const GeoCoord& coord ...` from `Projection`. It leads to compile problem when MercatorProjection is created as local variable...

### DIFF
--- a/libosmscout/include/osmscout/util/Projection.h
+++ b/libosmscout/include/osmscout/util/Projection.h
@@ -377,6 +377,8 @@ namespace osmscout {
     bool PixelToGeo(double x, double y,
                     double& lon, double& lat) const;
 
+    using Projection::GeoToPixel;	// make every GeoToPixel available
+
     void GeoToPixel(double lon, double lat,
                     double& x, double& y) const;
 
@@ -487,6 +489,8 @@ namespace osmscout {
 
     bool PixelToGeo(double x, double y,
                     double& lon, double& lat) const;
+
+    using Projection::GeoToPixel;	// make every GeoToPixel available
 
     void GeoToPixel(double lon, double lat,
                     double& x, double& y) const;
@@ -601,6 +605,8 @@ namespace osmscout {
 
     bool PixelToGeo(double x, double y,
                     double& lon, double& lat) const;
+
+    using Projection::GeoToPixel;	// make every GeoToPixel available
 
     void GeoToPixel(double lon, double lat,
                     double& x, double& y) const;


Framstag:
Yes, I was aware of the principle problem, but since I did not get a compiler error, I did the change nevertheless.

If you have multiple functions with the same name but different signatures and you overwrite one in a derived class, you must overwrite them all - or only the overwritten ones are visible in the derived class.

I'll take a look...

Karry:
This problem don't exists when you call `GeoToPixel(osmscout::GeoCoord ...` on reference to `Projection`, like here: [MapPainter.cpp:322](https://github.com/Framstag/libosmscout/blob/master/libosmscout-map/src/osmscout/MapPainter.cpp#L322) in library.

But it fails on stack-located `MercatorProjection` variable here : [DBThread.cpp:516](https://github.com/Karry/osmscout-sailfish/blob/master/src/DBThread.cpp#L516) in OSMScout for Sailfish OS.

Sorry, I don't have enough C++ theory knowledge, so I don't know why :-(

Framstag:
The base class defines both methods. So if you call the method using a variable that is of type Projections, everything works as expected.

The derived class however only defines one variant, so if you have a variable that instantiates the derived class directly, compiler will complain.

IMHO the best solution is to get rid of the variant with the two doubles altogether. Methods should always only work an GeoCoord. I'll try to fix the corresponding code as soon as I have some time.

Framstag:
I have now pushed a possible fix to the github repository. Let's see if it builds. This changes some APIs and calling code to use GeoCoord instead of lon/lat. 

However more changes could be done, even in Projection.cpp/h - bit this changes should at least fix above problem.

Karry:
Yes, using just one variant of this function solves the problem :-) But your recent changes contains small bug, I created another MR for this: https://github.com/Framstag/libosmscout/pull/52 . 

I am closing this one...

PR #59, opened 2016-06-27
fix build: with HAVE_LIB_GPERFTOOLS mallinfo undefined

Dushistov:
I got compilation error on Linux/amd64/gcc 5.3 that here:
```
#if defined(HAVE_LIB_GPERFTOOLS)
        if (heapProfile){
            std::ostringstream buff;
            buff << "load-" << level << "-" << x << "-" << y;
            HeapProfilerDump(buff.str().c_str());
        }
        struct mallinfo alloc_info = tc_mallinfo();
#else
```

mallinfo undefined, this PR fix this issue for me.

### DIFF
--- a/Demos/src/PerformanceTest.cpp
+++ b/Demos/src/PerformanceTest.cpp
@@ -40,6 +40,7 @@
 #if defined(HAVE_LIB_GPERFTOOLS)
 #include <gperftools/tcmalloc.h>
 #include <gperftools/heap-profiler.h>
+#include <malloc.h> // mallinfo
 #else
 #if defined(HAVE_MALLINFO)
 #include <malloc.h> // mallinfo


Karry:
Hi. It looks good for me. I am just curious why this problem occurs, can you provide your `gperftools/tcmalloc.h` please? I compiled gperftools manually (version 2.5) and it includes `malloc.h` from `tcmalloc.h`... I hope that any distribution don't provide gperftools without `tc_mallinfo` function :-) If yes, we should detect it.

Dushistov:
> $ grep include /usr/include/gperftools/tcmalloc.h 
> #include <stddef.h>                     // for size_t
> #include <sys/cdefs.h>   // where glibc defines __THROW
> #include <stdlib.h>   // for struct mallinfo, if it's defined
> 

I used google-perftools-2.0

> evgeniy@localhost ~ $ grep include /usr/include/stdlib.h  | grep malloc
> evgeniy@localhost ~ $ 
> 

So authors for tacmalloc.h were wrong, and stdlib.h not include malloc.h

[tcmalloc.h.zip](https://github.com/Framstag/libosmscout/files/335465/tcmalloc.h.zip)

Issue #58, opened 2016-06-27
"Where am I?" functionality. LocationService::DescribeLocation don't work for me


Karry:
Hi tim, I want to add "where am I?" functionality to OSMScout for Sailfish. I tried to use DescribeLocation demo, but it give me empty results. I want something like "you are in Czech Republic, Prague, near the street Biskupcova". Maybe something similar your described in feature request https://github.com/Framstag/libosmscout/issues/45

When I print found administrative regions in `LocationService.cpp:1264` it gives me some description where this position is, but why this information is not returned from `DescribeLocation` function?

    for (auto &adminRegion: adminRegionVisitor.adminRegions){
        AdminRegionRef regRef = adminRegion.second;
        std::cout << "admin region: " << regRef->name << std::endl;
    }
 
Am I looking for wrong function?

Karry:
Or just ask differently. Can I change LocationService::DescribeLocation to fit my and https://github.com/Framstag/libosmscout/issues/45 requirements? Or should I create another function?

Framstag:
The LocationDescription demo (I assume you mean that one :-)) uses the DescribeLocation method of the location service. This method is designed to return multiple description variants for the given geo location based on information from the database.

The LocationService allows the API to be extended internally by new description methods. Currently there are only two descriptions implemented.
* Description based on the geo location (this is 1:1 the input)
* In relation to a coordinate

The examples from the code comment:

```bash
$ src/LocationDescription ../maps/nordrhein-westfalen 51.57162 7.45882
Coordinate: 51.57162 N 7.45882 E
Is at address: In den Hüchten 29, Dortmund
```

```bash
$ src/LocationDescription ../maps/nordrhein-westfalen 51.57251 7.46506Coordinate: 51.57251 N 7.46506 E
7.6m S of address: Am Birkenbaum 6, Dortmund
```

Further implementations of location description are planed: See #43, #44, #45

I assume that in your case no address was found, because there is a limit regarding the  bounding box, the address is searched in:

```C++
    GeoBox           box100=GeoBox::BoxByCenterAndRadius(location,100);
```

The existing implementations are rather straight forward. So if you need another description this should be easy to add on your own.

There is should be no need for a new API, just extend the existing one by adding a new Description class.

Framstag:
I have added a small "how to implement" comment on all three above mentioned issues, to describe a possible implementation.

Karry:
Ok, I will try to extend current method. 

[...]

PR #66, opened 2016-07-06
 Extend LocationService::DescribeLocationByAddress to work with address nodes

Karry:
This pull request extend `LocationService::DescribeLocationByAddress` method to work with address nodes as discussed in issue https://github.com/Framstag/libosmscout/issues/58

The main change is in commit https://github.com/Framstag/libosmscout/commit/798be9e48c6693884198f35f0e445b383dd40484 , other commits contains small fixes and improvements related with addresses. 

I am not sure about commit https://github.com/Framstag/libosmscout/commit/7d48e698b53fbd5b7b6cd5ae069087673cfdc14b (find location by name, case insensitive)
I send question to *talk-cz* OSM mailing list, but I am not sure if we will be able to fix OSM data by some automatic script. I found around 10k these errors just in Prague! (All errors are imported from national address registry UIR-ADR.) I think that this problem is so common in Czechia that we should solve it in library. And similar problems can occurs in other countries too...

Edit: one user from OSM mailing list just suggest to ignore character size. This problem is common as geofabric validator shows: http://tools.geofabrik.de/osmi/?view=addresses&lon=14.49293&lat=50.09081&zoom=17&opacity=1.00&overlays=buildings,buildings_with_addresses,postal_code,street_not_found,nodes_with_addresses_interpolated,interpolation,interpolation_errors,nearest_points,nearest_roads
 and addresses are imported in monthly period from national database (not UIR-ADR, but RUIAN) and Czech cadastre don't want to fix street names... :-(

### DIFF
--- a/Demos/src/LocationDescription.cpp
+++ b/Demos/src/LocationDescription.cpp
@@ -100,13 +100,16 @@ int main(int argc, char* argv[])
     }
     
     if (place.GetPOI()){
-        std::cout << "    POI:     " << place.GetPOI()->name << std::endl;
+        std::cout << "    POI:      " << place.GetPOI()->name << std::endl;
     }
     if (place.GetAddress()){
-        std::cout << "    address: " << place.GetAddress()->name << std::endl;
+        std::cout << "    address:  " << place.GetAddress()->name << std::endl;
+    }
+    if (place.GetLocation()){
+        std::cout << "    location: " << place.GetLocation()->name << std::endl;
     }
     if (place.GetAdminRegion()){
-        std::cout << "    region:  " << place.GetAdminRegion()->name << std::endl;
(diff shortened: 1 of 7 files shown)


Framstag:
I have created issue #67 for this case.

Karry:
I just print debug message if location name don't match exactly:

> Using address 'XXX' instead of 'xxx'

[LocationIndexGenerator::FindLocation:1512](https://github.com/karry-space-with-my-second-forks/libosmscout/blob/7d48e698b53fbd5b7b6cd5ae069087673cfdc14b/libosmscout-import/src/osmscout/import/GenLocationIndex.cpp#L1512)

Framstag:
Manually merged with small changes

PR #82, opened 2016-08-11
Client Qt lib

Karry:
Hi. I prepared new library - libosmscout-client-qt, merge generic code from OSMScout2 and StyleEditor to it and setup build dependencies.

Autoconf based build seems little bit strange to me, but I don't know how to do better (libosmscout-client-qt  build is configured by qmake like OSMScout2 and StyleEditor...)

I plan to merge generic code from my Sailfish OS port to *client-qt library in future. It should help to everyone who will be creating some app based on libosmscout/Qt/QML.


Framstag:
The idea is good :-) I would however do the actual integration and use of the library in two steps:

* First create the new lbrary and get its autoconf and cmake build running.
* Than switch the OSMScout2 and StyleEditor application to use the library

Thus I will locally get the autoconf build running and see, if I get the cmake build running, too. For this the central builds have to be "green". I'll then will switch OSMScoutw to make use of the library and if this works StyleEditor, too.

The actual pull request I will initially reject, since it contains both steps in one go :-/

Karry:
Hi. I see that you already done first step. Should I continue with second one or you will finish that? :-)

Framstag:
Yes, you can ;-)

I only tested, that the library builds. It is possible that there are problems with using the library (for example because of not exported symbols). So you have to check this. Also it would be nice, if the library is handled as all other libraries. For example it should be detected using pkg-config by OSMScout2/StyleEditor. This way the build is more robust. It is also likely, that some scripts like setupAutoconf.sh  regarding central builds must be extended (library path, pkg-config lookup). I can fix this, too, but it would be nice if you handle the obvious stuff.

Framstag:
Closing, because there is already a new pull request for the same change.

PR #84, opened 2016-08-13
Client Qt integration

Karry:
Link StyleEditor and OSMScout2 with client-qt library, remove duplicate files.


Framstag:
Near perfect, had only to change setupMSYS2.sh. Merged :-)

PR #86, opened 2016-08-15
add style definition for railway_narrow_gauge

Karry:
Hi. I added style definition for railway_narrow_gauge. It is already defined in import definition, just render style was missing. This type of rail is rare in Czechia, but usual in Switzerland. It is weird when you go to rail station, open map there are no rails :)

I am just thinking if we should render rails with property "rack=yes" with different style?
railway_with_rack


Framstag:
It is of course possible to make two different types from it. The alternative is to define a feature for it. I'm however not sure, if style sheet filtering by features already works.

PR #87, opened 2016-08-15
style parser should not fail (just warn) when style contains types missing in database

Karry:
Hi, I was experimenting with new map types. I added new type to import definition (map.ost) and then I created style (.oss) for it. Just for check that style file is correct, I tried to open it. And it fails, because it contains types that are not in my old database!

It is not optimal. I can imagine situations when user generate or download database and then use it for months. When he update application (and style) it should work even that database types are outdated.

This commit change semantic errors just to warnings. I tested it with described usecase and it works. Can be some hidden problems here that I missed?


Framstag:
You should not change the Scanner/Parser sources directly. These are generated from the *.atg file. If I change the grammar I will regenerate them and their changes get overwritten.

The tool used is Coco/R for C++: http://www.ssw.uni-linz.ac.at/Coco/

I have adapted the *.atg file, so your changes will stay :-)

PR #98, opened 2016-08-24
import: workaround optimized way and area index generation problem

Karry:
Problem is that zero cell offset is valid offset now. But zero for reader means
that this cell don't contains data! This cause that data in first cell are
never visible.

This workaround move data start by one byte. It creates little bit larger output file.
But generated format is bacward compatible and all cells with data are visible to
reader code.

TODO: when data format will be changing, consider usage ones (0xFF..FF)
as empty placeholder

Following screenshots (DrawMapQt moutput) shows that now are optimized data visible:

trunk import with bug
map-1-bad
fixed optimized ways
map-2-fixed-ways
fixed ways and areas
map-3-fixed-ways-and-areas


Framstag:
Thanks for analysing this problem. In fact I made the same error in another index a long while ago (and solved it by +1, too). So I should have known better...

PR #110, opened 2016-09-06
POC: split long ways - it is based on count of nodes (300) or real length (30 km)

Karry:
Hi. We know that labeling of long ways has nasty performance impact (at least with Qt backend). You recently added limit for merging ways to keep node count < 300. But some ways are so dense that it may have real length 150 km just with 100 nodes on map. 

It was proven with one ferry line here: https://github.com/Karry/osmscout-sailfish/issues/8#issuecomment-244596229

Because of that, I created method that split long ways (based on real length or node count) in import process. I tested it and it improves rendering time ten times (from 12s to 2.4 s).

This is just a fast proof of concept. It need better log messages, progress reporting, enable OpenMP for loops, and configurable parameters instead of constants...

### DIFF
--- a/libosmscout-import/include/osmscout/import/GenWayWayDat.h
+++ b/libosmscout-import/include/osmscout/import/GenWayWayDat.h
@@ -89,6 +89,10 @@ namespace osmscout {
                    std::list<RawWayRef>& ways,
                    std::multimap<OSMId,TurnRestrictionRef>& restrictions);
 
+    bool SplitLongWays(Progress& progress,
+                       std::list<RawWayRef>& ways,
+                       CoordDataFile::ResultMap& coordsMap);
+
     void WriteWay(Progress& progress,
                   const TypeConfig& typeConfig,
                   FileWriter& writer,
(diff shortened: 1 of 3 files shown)


Framstag:
For me it would be interesting to know how many ways are problematic. If we split only a small percentage of ways the negative impact on data size and routing performance is likely neglectable. If not, a geometric solution during rendering might be more helpful.

Karry:
I tried this modification with Estonia. Here is [part of import log](https://gist.github.com/Karry/77592a344a1fc4bb1d4ea8be097304b5). There was 175102 ways at all (step 10), some of them was merged, count was reduced to 152208 (diff 13% - 22894) [1]. Then, 690 of long ways was splitted to 2213 new segments. So if I count correctly, there should be 153731 ways after import step 10.

Disk size: When I count only export files, version with splitted ways is bigger just by 16 KiB (74532 KiB vs. 74516 KiB - less than 1%).

1)
```
$ cat import_splitted.log | grep "Reduced ways" | awk 'BEGIN{before=0 ; after=0}{before+=$6; after+=$8}END{print (before - after)}'
22894
```

PR #113, opened 2016-09-10
letting to install private includes

rinigus:
While trying to install libosmscout libraries and use them for development, I stumbled on a problem that some files need private/CoreImportExport.h and others to compile. To reproduce:

```
mkdir build
cmake -DCMAKE_INSTALL_PREFIX:PATH=<_PREFIX_> -DOSMSCOUT_BUILD_MAP_OPENGL=OFF -DOSMSCOUT_BUILD_IMPORT=OFF -DOSMSCOUT_BUILD_MAP_AGG=OFF -DOSMSCOUT_BUILD_MAP_CAIRO=OFF -DOSMSCOUT_BUILD_MAP_SVG=OFF -DOSMSCOUT_BUILD_MAP_IOSX=OFF -DOSMSCOUT_BUILD_TESTS=OFF -DOSMSCOUT_BUILD_DEMOS=OFF -DOSMSCOUT_BUILD_BINDING_JAVA=OFF -DOSMSCOUT_BUILD_BINDING_CSHARP=OFF -DOSMSCOUT_BUILD_DOC_API=OFF -DOSMSCOUT_BUILD_CLIENT_QT=OFF -DOSMSCOUT_BUILD_TOOL_OSMSCOUT2=OFF -DOSMSCOUT_BUILD_TOOL_STYLEEDITOR=OFF -DGPERFTOOLS_USAGE=OFF ..
make 
make install
```

Try to compile your program with corresponding _prefix_. It would fail due to missing includes. 

This PR fixes the problem for Qt map drawing. If its OK, the same approach should be applied to other parts.

### DIFF
--- a/libosmscout-map-qt/CMakeLists.txt
+++ b/libosmscout-map-qt/CMakeLists.txt
@@ -35,5 +35,5 @@ install(TARGETS osmscout_map_qt
         RUNTIME DESTINATION bin
         LIBRARY DESTINATION lib
         ARCHIVE DESTINATION lib)
-install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/osmscout DESTINATION include FILES_MATCHING PATTERN "*.h" PATTERN "private" EXCLUDE)
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/osmscout DESTINATION include FILES_MATCHING PATTERN "*.h" PATTERN "private/Config.h" EXCLUDE)
 install(FILES ${OSMSCOUT_BASE_DIR_BUILD}/include/osmscout/MapQtFeatures.h DESTINATION include/osmscout)
--- a/libosmscout-map/CMakeLists.txt
+++ b/libosmscout-map/CMakeLists.txt
@@ -45,5 +45,5 @@ install(TARGETS osmscout_map
         RUNTIME DESTINATION bin
         LIBRARY DESTINATION lib
         ARCHIVE DESTINATION lib)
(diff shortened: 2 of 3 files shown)


Karry:
If some (private) headers are necessary for building another programs, I suggest to move it to API include dir (installable) instead of install all private headers by default...

Can you provide more info about includes of your app? What error gives you compiler?

rinigus:
Sorry, I guess I formulated it wrong. The headers are used from libosmscout include files. 

I was just trying to compile something similar to Demo/DrawMapQt.cpp:

```
In file included from ../../../tmp/osm/include/osmscout/Database.h:31:0,
                 from main.cpp:7:
../../../tmp/osm/include/osmscout/Types.h:23:47: fatal error: osmscout/private/CoreImportExport.h: No such file or directory
 #include <osmscout/private/CoreImportExport.h>
```

cheers

Framstag:
Private in this case has two meanings: Either the headers are really private in that they should not be visible for "clients" and thus should not be installed. On the other hand some headers are private in the sense, that they must be visible during client build but should not be imported directly by the client.

Config.h is *really* private since it contains defines that are likely redefined by other libraries, too. For headers and clients CoreFeatures.h should be used instead.  CoreImportExport.h contains symbol import/export code and must be visible to the client to correctly link - but should not be included directly by the client. In this case "internal" is possible a better name (the same scheme exists for the other libraries).

I merge the pull request anyways since without it some things won't work.

rinigus:
Yes, that's important to separate these two 'private' types of headers and probably the simplest would be to use _internal_, as you proposed. Note that even after this PR some other backends could be broken. But I presume you plan to fix it via introducing internal and, when its done, reverting this commit.

Thanks for merging it,

cheers,

rinigus

PR #119, opened 2016-09-18
use addr:place when addr:street is not available for address object

Karry:
In OSM data can be used "addr:place" instead of "addr:street" for addresses that don't contains street. http://wiki.openstreetmap.org/wiki/Key:addr:place It is usual for small villages. From library view, there is no difference. It is just a address :)

I have small question with this change: I am not able to open old imports now. If fails with error:
```
Requested and actual tag id do not match
```
It is possible to make these changes backward compatible? If I understand it correctly, `TagID` is used just for import right? When database is open for read, there is no need for these checks...? When I remove these check, loading of old database works without problem.

### DIFF
--- a/libosmscout/include/osmscout/TypeFeatures.h
+++ b/libosmscout/include/osmscout/TypeFeatures.h
@@ -277,6 +277,7 @@ namespace osmscout {
   private:
     TagId tagAddrStreet;
     TagId tagAddrHouseNr;
+    TagId tagAddrPlace;
 
   public:
     /** Name of this feature */
@@ -342,6 +343,7 @@ namespace osmscout {
   private:
     TagId tagAddrHouseNr;
     TagId tagAddrStreet;
+    TagId tagAddrPlace;
 
   public:
     /** Name of this feature */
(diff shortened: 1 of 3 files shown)


Framstag:
Yes, you are right. With the existence of Features, the tag ids are not required anymore. At least tag ids are not stored in one of the *.dat or *.idx files.

However TypeConfig writing/loading is also used if you restart parts of the import (using "-s" and "-e" options). In this case it must be made sure that the import runs with the original configuration. Though even this is not required anymore. Tags are currently only relevant in the initial Preprocess Code and the ids are only temporary.

I assume the actual problem occurs, because some tags are implicitly registered in the TypeConfig and thus the order is changed (or better, new tags are injected within the old order).

We should however check if TagId is used outside the initial import step. If not, it could be completely dropped from the TypeConfig serialization code (changing the file format once again ;-))?

Karry:
Thanks for answer. I will remove these check from my "backward compatibility" branch ;-)

Framstag:
First look: It seems like it is only used during loading of OST file and preprocessing.

Karry:
Hmm, this change added house numbers to map, but location index don't work for these objects. Import log is full of errors like 
```Street of address 'Moldava' '93' cannot be resolved in region 'Moldava'```
It is obvious that "Moldava" region (village) don't have any location (street) with name "Moldava". How to fix it? I tried to add `LOCATION` flag to `place_village` and `place_hamlet` type, but it don't work. Why? Should we add fake location to region with region name?

PR #120, opened 2016-09-22
fix place index building when address don't have street, but just place

Karry:
In recent PR https://github.com/Framstag/libosmscout/pull/119 , I added support for addresses that don't have street (addr:street) but these are defined in place (addr:place). It added house numbers to map, but location index don't work for these objects. Import log is full of errors like 
```Street of address 'Moldava' '93' cannot be resolved in region 'Moldava'```

In such case, index generator create virtual address with same name as region. It is not beautiful solution but it is simple and works. With this change we add many new nodes to location index...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 2.8.11)
 project(libosmscout C CXX)
 set(CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS ON)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/" "$ENV{QTDIR}" "${QTDIR}" "$ENV{QTDIR}/lib/cmake" "${QTDIR}/lib/cmake")
@@ -20,6 +20,8 @@ if(APPLE AND NOT ANDROID)
   set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++11")
   set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++")
 endif ()
+
+add_definitions(-std=c++11)
 
 include(CheckTypeSize)
 CHECK_TYPE_SIZE("void*" OSMSCOUT_PTR_SIZE BUILTIN_TYPES_ONLY)
--- a/Demos/src/LocationDescription.cpp
+++ b/Demos/src/LocationDescription.cpp
@@ -55,6 +55,9 @@ void DumpLocationAtPlaceDescription(osmscout::LocationAtPlaceDescription& descri
 
   if (place.GetPOI()) {
     std::cout << "  - POI:      " << place.GetPOI()->name << std::endl;
+    if (place.GetObjectFeatures()){
+      std::cout << "  - type:     " << place.GetObjectFeatures()->GetType()->GetName() << std::endl;
+    }
   }
 
   if (place.GetAddress()) {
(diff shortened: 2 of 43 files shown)

### CODE REVIEW

**Comment by Framstag** [comment]
I assumed that if you use addr:place instead of addr:street there must be some area close or the object must be within an area with have the given name. This area could either be some highway=pedestrian, though it could also be an island a village or similar.

Such objects should/could be marked - if not already marked as ADMIN_REGION - LOCATION. If we resolve addr:place we should first look for a location with the same name (in the same region) and then look for a region with that name. In that case we would have to extend the data format to have addresses directly placed add region and not indirectly via a location (which in turn is placed at a region).

Does this not work? A certain error rate is expected because of spelling mistakes prefixed like "city", "city of" and similar. This IMHO are mapping errors.

Interestingly enough I did not find one (in my opinion) correctly mapped case in my region. The occurrences of addr:place also have a valid addr:street tagged and addr:place named the city containing the object. Hmmm.....

**Comment by Karry** [comment]
Actually, add LOCATION to place_village was first thing that I tried
```
  TYPE place_village
    = NODE AREA ("place"=="village")
      {Name, NameAlt}
      ADMIN_REGION LOCATION
```
but it don't work. Location list for my test region (village) is still empty. I don't investigate it yet. Do you have some idea?
Here is test extract: https://osmscout.karry.cz/kokorinsko.pbf

in center of village is point with place=village (it is "center" part of boundary relation): http://osm.org/node/302373718

Import log contains many unresolvable addresses in "Bosyně" village...
```
Street of address 'Bosyně' '70' cannot be resolved in region 'Bosyně'
````

Karry:
It is possible that node can't be location? In `LocationIndexGenerator`, I see `IndexLocationAreas` and `IndexLocationWays` but no method for nodes...

Karry:
Please review this commit: https://github.com/karry-space-with-my-second-forks/libosmscout/commit/90a3530864b6ab1e0114a9f59c37ccba94ef10a6

Do you like this solution more? :)

Framstag:
I always assumed that a location is either a way or an area since a location is something like a small region that can have close addresses. Like a place (houses around an area) or a road (houses a long the way). So yes, I did not add support for nodes being locations. I assume that you now have the problem that a place can also be represented by a node? That sounds like bad mapping but I could image that mapping reality sees it different (especially for smaller villages which do not have any region mapped).

Framstag:
I'll take a look at it over the weekend. Its getting to late for today...

Framstag:
OK, in principle yes. Though you have added LOCATION now to node types that are already ADMIN_REGION. The ADMIN_REGION hcode handles such nodes already. I would have expected that it is enough to add LOCATION to things like highway_pedestrian (though this would not make much sense as node). I'm still not sure if you are doing duplicate work. addr:place either references a location (which could be a node which node type not has ADMIN_REGION but LOCATION) or a region (which should already have ADMIN_REGION). Note that nodes which are of type ADMIN_REGION where later on a area with similar name was found is handled as RegionAlias (AddAliasToRegion()). The region they are an alias to is either the region itself (in this case there was also an area for this region) or the parent region (in the case where there is no area for this region). This is assured because we are first looking for admin_boundaries and areas and then later for nodes. For the addr:place == region case would suggest to add a list of RegionAddress directly to Region since this is semantically what is meant (for the first case you still have a Location to assign the address to).

Framstag:
After thinking about it (again), in case where you assign a address to a region, you at least must clearly identify the alias, else you might loose the information about the actual place, island etc. name

Karry:
Can you show me some example with region alias please? I probably dont understand what you mean with it.

Karry:
We are talking about this commit right? https://github.com/karry-space-with-my-second-forks/libosmscout/commit/90a3530864b6ab1e0114a9f59c37ccba94ef10a6 

As alias, you mean when in relation regions (relation `boundary==administrative`) contains more nodes with `place==village` with different names? It should be works as well... or not?

Framstag:
The location index uses containment to assign location to regions. So we need an area for a region. In some case we only have nodes though - normally because the mapper did not have any knowledge about the area "owned" by the region. So in this case we look for the parent region and give it the child region (without any area) as an alias. This is OK in cases where the parent region and the arealess child region do not have locations with the same name in common. It is the best we can do (dropping them would be worse).

So I think the short answer to your question is "yes" :-)

Karry:
And long answer? :-) Is here some case that is not covered by this solution? https://github.com/karry-space-with-my-second-forks/libosmscout/commit/90a3530864b6ab1e0114a9f59c37ccba94ef10a6 As region nodes (villages) are added to region as Location (and region name aliases), correctly defined addresses with addr:place should be resolved... Or I missed something?

Framstag:
Ah, the joys of written communication :-) Yes, it will likely work but I found it "not optimal" if you are handling e.g. place_village as location if it is already detected as region. It is not necessary to add it as location (though it would keep the database simpler), it would be enough to add a address to the existing region or region alias (which makes the file format more complex).

This of course is different if we are speaking of highway_pedestrian, which can act as a location and IMHO is a "real" location. In this case adding LOCATION to the map.ost is completely correct and (after adding support for nodebase locations) should "work".

I think this models the situation better than adding a fake location in this case. I admit I haven't yet though about how much it makes the implementation more difficult ;-)

Karry:
Ah, ok. I understand now :) In case that region have more aliases, we will need to assign address directly with region alias (otherwise we can't represent address correctly). Yes, it will complicate database :-( I wanted to do it simple. And I don't think that this simple solution is so bad. When you have some town and near village in its administrative region, this village is real "location" in my point of view, not a fake one :-)

Framstag:
I'm not sure if the double entry for region based location might be a problem later on.

Suggestion:
We drop the need to annotate ADMIN_REGION with LOCATION, too, if possible. We do add region as location if required for addr:place and see how it works out (as you planed).

This way the *.ost file is semantically clean, we try the simple solution first and we are still free to change the internal format if we run into problems later on without having the need to change the externally visible, too, in that case.

OK? Does that make sense?

Karry:
Ok. I added another commit that lookup region aliases it its names matches to place (location). https://github.com/Framstag/libosmscout/pull/120/commits/6bde3723429b0833e185c546d8e329559b89cad7 It helps to resolve another locations...

PR #121, opened 2016-09-22
Packaging and compiling on Sailfish

rinigus:
This is mainly request-for-comment at this stage. The submitted PR allows to package libosmscout Qt version for Sailfish. As a result of the package building script, 2 RPMs are generated leading to libosmscout-qt and libosmscout-qt-devel packages. For development of apps, the both packages have to be installed in SDK. For end user, libosmscout-qt is sufficient.

The problems with this PR:

1. I had to modify CMakeLists.txt to reduce cmake version and add c++11 definitions. Maybe it has to be done differently.

2. Jolla Store does not allow libraries to be submitted. So, I would have to use an alternative channel (OpenRepos) to distribute the work. I, personally, don't find that this is a problem. @Karry has done a great job in integrating libs with his app to be able to submit into the store, but I would prefer to have libraries installed as such and share them between applications. It also makes development of applications easier when you don't have to jump through such number of obstacles at once.

3. Its stated on libosmscout page that there are no releases / packages yet. This PR changes it, at least in part. 

Would be great to know what you think, @Framstag and @Karry. I am sure that this packaging script is not the best. It works, but maybe there is a better way to package it using cmake / rpm somehow.

cheers,

@rinigus


Framstag:
Looks like the central builds are still OK. So the changes to the cmake files did not broke anything obvious. A comment regarding the rest of the changes is up to Karry.

Karry:
I don't see problem with changes. I just suggest to package **libosmscout-client-qt** too. It may help when someone will preparing Qml UI.

To distributing libraries - I understand why Jolla not permits installing libraries to system locations (`/usr/lib/`), It have to be installed to application specific location (`/usr/share/<appname>/lib`), otherwise it will ends with dependency hell and Jolla QA team will be sentenced to be linux distro maintainers. I don't see problem with it, when usual user have installed around thirty applications on mobile, that occupies ~200 MiB on 16 GiB storage. Of course, you can distribute it via OpenRepos and move this nasty work to experienced users... but I think that creating some *spec* file template for new applications will be better...

I think that applications in Harbour (Sailfish OS store) has better change to address users than applications in OpenRepos. We should help to another developers to meet strict Harbour rules.

rinigus:
@Karry: let's address issues one-by-one.

Harbor vs OpenRepos: as a ported device owner, I can see that there are two parts in SFOS community with one of them having Harbor access and one that has access to OpenRepos only. Most of the ported devices are excluded from Harbor [exceptions are Nexus 4 & 5 and FP2, I think]. Its promised that the situation will change "real soon", but that promise has been there for a while as well :). Now, those who are majority in Harbor have an official Jolla device. As a result, they have access to Android apps through dalvik and can use offline apps, such as Here. To get attention of these users, you would have to have a product that competes with Here or similar apps.

Users of OpenRepos that are coming from ported devices do not have access to Android map/location programs (sfdroid does not support GPS yet) and are looking for SFOS-based offline solution. I presume that these users would be very interested in getting something that works [hence my interest as well]. So, I would suggest to distribute your app there in addition to Harbor. 

Libs embedded or not: Rules in openrepos are much simpler and allow to distribute the programs/libs in more Linux-like way, taking advantage of splitting libs and apps when needed. For rapidly developing lib, such as libosmscout, I would prefer to have library distributed separately. For example, even by following your development for less than a month, there was one change in database format that rendered old map files useless. Would be good to keep at least maps available for all libosmscout using apps.

I agree that the space used by the libs is usually not an issue (this spec leads to ~40MB libs). 

Ease of development: 
For me, its easier to develop an app when I can just link to the lib and don't have to drag full (or large part of) libosmscout with app rebuilding. It allows me to use Sailfish SDK from qt creator easily, for example.

Now assuming that we have only one app using libosmscout installed, the ability to incorporate lib into the app is attractive from distribution purposes. Its only one package to maintain and you could get to the Store as well.

In the end, it seems to me that the choice on whether to use separate lib and app packages or incorporate them into the same package is a personal choice. 

I'd suggest to use a spec file and the build instructions for osmscout-sailfish as an example on how to do single-package installation with all libs included.  The packaging proposed in this PR is not covering one-package-including-app scenario and could be used by developers that are preferring to have the lib installed separately.

re libosmscout-client-qt: Sure, I could add that as well. I'll see how large this lib is and then either put it into separate package or add into the main package.

Best wishes,

rinigus

Karry:
@rinigus thanks for response. I expected that Harbour is available on all unofficial devices too! Your view makes more sence now. 

>So, I would suggest to distribute your app there in addition to Harbor. 

Added to my todo list ;-)

Btw, offline Here maps (Android app) are officially available on Intex Aqua fish only as I know. It is stupid licensing issue. But it gives me motivation for developing native map application.

rinigus:
I'll extend then the scripts to include libosmscout-client-qt as well and will come back when its ready.

rinigus:
@Framstag: Thank you for merging it!

I'd like to polish it a little bit more. Namely, I have added the libosmscout-client-qt to the spec and since this lib part is small, it would be great to include it.

In addition, there is one part in spec that I don't like: I had to make a build directory (build-rpm) where cmake is run. I don't know yet how to treat that directory correctly (it should be deleted, I guess, after build). @Karry, maybe you have an idea? Please look at 

https://github.com/Framstag/libosmscout/blob/master/packaging/sailfish/libosmscout-qt.spec#L35

When the new spec is ready I'll push it then separately through the next PR.

Karry:
Hard to say how to write spec file correctly. I have just a little experience here. I guess that it should be built in some /tmp directory inside SDK VM and it should be deleted after build... But I do exactly same - create build directory as project subfolder - it speedup next builds, when already compiled objects are skipped...

rinigus:
@Karry: thank you! I'll look into it a bit later. I'd like (at least) to have full path for directory name and not relative, as it is now.

PR #124, opened 2016-09-29
Print object type and its features in LookupText demo

Karry:
This demo now demonstrate how to retrieve objects by its type and file offset and print useful informations for it. 

LookupText requiere complete database now...

```
$ ./Demos/LookupText /media/karry/data/osm/osmscout/kokorinsko
* Searches are case-sensitive
* Displays up to 10 unique text results
* Displays up to 5 file offsets for each result
* Input at least 3 characters or 'q' to quit


Enter a search term:
Bosyn
"Bosyně" -> 
 * N:4125
   - type:     highway_bus_stop
   + feature Name: Bosyně
 * A:31755
   - type:     boundary_administrative
   + feature Name: Bosyně
   + feature AdminLevel
```

### DIFF
--- a/Demos/src/LookupText.cpp
+++ b/Demos/src/LookupText.cpp
@@ -19,7 +19,9 @@
 
 #include <iostream>
 #include <osmscout/TextSearchIndex.h>
+#include "<osmscout/Database.h>"
 #include <algorithm>
+
 
 void badInput()
 {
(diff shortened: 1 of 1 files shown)


Framstag:
I prefer the use of range based for loops over containers (See http://en.cppreference.com/w/cpp/language/range-for) if possible, else it looks OK.

PR #127, opened 2016-10-01
initialize maxSpace in LabelLayouter

rinigus:
Fixes an issue with many labels rendered on SFOS


rinigus:
Tim, I messed up. It should be maxSpace not maxScale in the file. Have to run now, will be able to update in few hours

rinigus:
I could see you fixed it for me. Thank you and sorry again for introducing such bug.

Framstag:
No problem. Thanks for the analysis, this would have cost me much time and I would not have found the cause.

PR #132, opened 2016-10-09
Packaging sfos marisa cairo

rinigus:
These changes in RPM spec:

* add marisa as dependency
* add cairo (while commented out, its easy to move code back in)
* add client_qt library to the package

Mainly MARISA is used in the server, but other changes maybe handy one day.

Best wishes,

rinigus


Framstag:
Since I do not use the RPM: @Karry, OK for you?

Karry:
looks good for me

PR #138, opened 2016-10-12
add true|false to delete-temporary-files/delete-debugging-files

rinigus:
Specifying that a Boolean is needed after delete options in Import


Karry:
LGFM

PR #144, opened 2016-10-25
Index tram and bus stops as POI

Karry:
Hi Tim, we discuss with @rinigus that stops should be indexed as POI https://github.com/rinigus/osmscout-server/issues/17
What do you think?


rinigus:
@Karry - thank you for adding this! I hope it would get merged :)

PR #159, opened 2016-11-16
Qt Client enhancements

Karry:
Hi All. 

Last months, when I worked on port for Sailfish OS, I always keep in mind that my changes and improvements should be merged to upstream library one day. But I postpone it many times, because it was not perfect. Code is not perfect neither today. But with recent changes, it starting to be unmaintainable in separate repository... So I created this huge merge request to `libosmscout-client-qt`. It will break some functionality, but I hope that benefits are bigger than disadvantages...

main changes:
 - DBThread is abstract now and rendering functionality is moved to `TiledDBThraed` and `PlaneDBThread`. Tiled rendering combine offline database with online OSM tiles. Plane rendering use similar approach as old DBThread, but **asynchronous data loading is broken now**. I will try to fix it later.
 - `MapWidget` supports **multitouch input** by custom gesture recognizing and smooth animations when map is moving. But it **don't support map rotating** now. It is directed to north always. 
 - DBThread support multiple databases at same time. It has some limitations - routing don't work when it crossing database boundary and rendering is not perfect when databases are overlapping...
 - `SearchLocationModel` is asynchronous, so searches don't block UI thread. It support free text search now.
 - `LocationInfoModel` is Qt abstraction for query location description.

Sorry that it is so huge, I understand that it is hard to review. Some functionality can't be splitted to smaller parts.


Framstag:
There are two feature you say that are broken, that I would definitely miss:
* Asynchronous loading - since it is IMHO a major performance/experience improvement. Is it broken in both DBThread implementations?
* Rotating - if it is broken I cannot test, if it is still working. I do not need any multitouch support for it, but rotating via keyboard would be nice for testing.

It is possible to reintegrate this features?

I still need to review the individual commits.

Karry:
Hi. 

I want to fix asynchronous loading for `plane` variant. Tiled variant upscale/downscale available tiles from nearest zoom levels, it is better to wait for complete tile then display almost empty tile just with ways.

With rotating, every map transformation will be more complicated. But I can try to bring it back...

Karry:
Hi. I fixed asynchronous data loading and make Plane DB thread more smooth in recent commit...

libosmscout-client-qt/src/osmscout/MapWidget.cpp
```diff
@@ -1,406 +1,421 @@
...
-
-  This library is distributed in the hope that it will be useful,
-  but WITHOUT ANY WARRANTY; without even the implied warranty of
-  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-  Lesser General Public License for more details.
-
-  You should have received a copy of the GNU Lesser General Public
-  License along with this library; if not, write to the Free Software
-  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307  USA
+ OSMScout - a Qt backend for libosmscout and libosmscout-map
```
Framstag: You are moving here from LGPL to GPL, while the cod eis still in the client-qt-library.

Framstag:
That pull request is HUGE. In cannot do any reasonable code reviews on that. Ill wil checkout the branch and just see how it works.

What I have seen though:
* Some includes use "" instead of <>. I would suggest to at least aways use <> within the library.
* Some license header use GPL instead of LGPL in the library. It seem like you fixed some, possibly not all
* The formatting is a mess, but it was before. We should just decide if we use standard Qt formatting or the formatting of the libosmscout libraries. Currently it is a mix (with some flavour by you now, too ;-)). That we can fix though after applying the patch.
* Else, some more comments would be helpful now, since the code got more complex. We crossed the border of demo application size, though most people still would look at it first to see how it should be done.
* It broke the MSVC build, since marisa is not available there:

```
In file included from osmscout/DBThread.cpp:26:0:
C:/projects/libosmscout/libosmscout/include/osmscout/TextSearchIndex.h:29:20: fatal error: marisa.h: No such file or directory
```

Karry:
Hi. Sorry for the size. I postponed merge too long...

 - I updated includes and use <> (commit db1801dd7042781803fc8b43fd824a3a4da6fa1f )
 - All sources are under terms of LGPL in branch head.
 - Free text search is disabled when Marisa is not available ( a3214d123c8b63b0a763044d5f9c02b953242d77 ). Thanks for this warning.

Formatting is not unified, I know :-( For me it is hard to switch my mind between scopes. I am writing code in Scala/Java in my work, then osmscout library with Qt API... I prefer to use Qt formatting when class is child of `QObject` or using some Qt API. When I look to https://wiki.qt.io/Qt_Coding_Style , I don't respect some recommendations :( But I can try to fix it. What do you think?

My goal is to create platform independent library with various functionality for creating map application in Qt/QML. Beside current functionality I would like to implement map downloader, turn-by-turn voice navigation, manager of favorite places... It goes beyond simple demo application. I will try to write API documentation later...

Framstag:
Yes, the formatting in the Qt examples are (partly) different, because I in between used the Qt developer suite to work on these files. The idea was to use Qt formatting style for Qt sources. Though now I'm using Clion for everything this is hard to keep. I'm unsure how to proceed regarding formatting.

The marisa problems sees to be fixed, but if you look at https://ci.appveyor.com/project/Framstag/libosmscout/build/365, there is still an import to sys/socket.h, which does not exists under Windows (perhaps disable tile downloading under Windows by dropping this files under Windows? I assume that is the cause?). There are also a number of warnings.

I try to keep the compile warning free under gcc and clang, could you please take a look at it?
Could you also merge in my Routing API changes (see mail)?

Thanks for the fixes.

Yes, additions for the mentioned features would be cool, however the problem is, that it would be better, if these features are not bound to Qt. On the other hand I'm aware that Qt offers some functionality that may make it simpler to offer such features (for example network IO,...).

Karry:
Regarding to warnings, we have compiler option `-Wno-unused-parameter` in CMake. It suppress majority of those warnings. We can add this option to autoconf build too or define some UNUSED macro [1] that will suppress this warning and mark such parameters explicitly with that... 

1) http://stackoverflow.com/questions/15763937/unused-parameter-in-c11

Karry:
socket.h was not needed, I removed it. It was added by my IDE (Netbeans) probably and I overlook it when commiting...

Framstag:
Regarding "unused" warnings: I would suggest (and have practised this in libosmscout), to just comment out the name of function parameters. This is the simplest solution IMHO.

Karry:
Nice and clear solution. I learn something new every day :-)

Framstag:
I now build the path locally:
* OSMScout2 on start does not show the center of the map anymore. The initial coordinates were far of the map I loaded. Have you changed that?
* To get the map to show, I searched for a location. This worked and map focused on that location. However I was not able to leave the search bar anymore. Without leaving the search bar I'm not able to use the shortcuts anymore. I especially use the cursor movement buttons and "+" and "-" for navigation in the map.
* It looks like you are rendering much more "frames" or are calculation intermediate images by bitmap scaling. This is a very useful addition, but on my desktop it looks like it is rendering slightly too much images. So zooming is smooth but with minimal lack. Is this possible?
* It looks like the center of zooming is not the center of screen anymore but the position of the mouse?

Karry:
Hi. 

I added Seetings (singleton) and QmlSettings (qml element) that may persist `MapView` between sessions. Recent view (position, magnification...) is loaded on OSMScout2 start by this code:
```
Component.onCompleted: {
                map.view = settings.mapView;
            }
```
as side effect that initial value points to projection center  (lat=0, alt=0) is that map bounding box is not displayed on first startup :-( For that reason, I added button with "*" to right bottom corner to `recenter` to database bounding box. I know, it is not obvious. I may fix it somehow.

___

Sorry for that focus bug :-) I tried to fix it in recent commit.
___

Yes. My goal was to provide smooth animations for zoom and move map. Target is 60 fps. View changes are controlled and emitted to `MapView` from Input handlers (`MoveHandler`, `DragHandler`...). `DBThread` instance then paint map from cached tiles or transformed `finalImage` and render path asynchronous. Code changed a lot in rendering pipeline. `TiledDBThread` is relative simple, `PlaneDBThread` is little bit tricky... Abstract `DBThread` don't contains rendering, just database handling, searching and routing...
___

Yes. Zoom is controlled by `MoveHandler::zoom` or `MultitouchHandler` in case of `QTouchEvent`. It tries to keep location under mouse cursor on the same screen position before and after zoom. My goal was to provide similar behaviour as web Google maps.

Framstag:
Look better now. The focus problems witht he search field do not appear anymore. 

However I get a crash if opening the about dialog and closing it. I'm not sure, if that problem was there before?

```
[tim:~/projects … Scout/OSMScout2] karry-space-with-my-second-forks-tiled-client-qt+* 139 ± LC_ALL=C debug/OSMScout  ../maps/nordrhein-westfalen ../stylesheets/standard.oss 
Loading online tile providers from  ":/resources/online-tile-providers.json"
Reported screen DPI:  143.907
Map DPI override:  143
qml: Gain focus, update popup
qml: Gain focus, update popup
TextFontSize: 13 px
Average char width: 7 px
qml: Position changed:
qml: Positioning is true
qml: Last error 3
Initialize
Initialize
found database:  "/home/tim/projects/OSMScout/maps/nordrhein-westfalen"
Database::Database()
TypeConfig::TypeConfig()
Opening TypeConfig: 0.001
StyleConfig::StyleConfig()
Failed to set Geoclue positioning requirements. Geoclue error: org.qtproject.QtDBus.Error.InvalidObjectPath
Opening StyleConfig: 0.028
BoundingBox: [49.92181 N 5.58645 E - 52.70659 N 9.64789 E]
Opening BoundingBoxDataFile: 0.000
MapPainter::MapPainter()
Start data loading...
Opening OptimizeAreasLowZoom: 0.000
Opening OptimizeWaysLowZoom: 0.000
Opening AreaAreaIndex: 0.000
Opening AreaWayIndex: 0.000
Opening AreaNodeIndex: 0.000
Opening NodeDataFile: 0.000
Opening AreaDataFile: 0.000
Start rendering timer: 10 ms
Opening WayDataFile: 0.000
DrawMap()
Draw: [51.50220 N 7.45460 E - 51.52622 N 7.47775 E] 19567x/14 480x800 143.00000 DPI
ERROR while loading image 'scrub'
Paths: 4299/748/748/169 (pcs) 0.004/0.018/0.006/0.011 (sec)
Areas: 1098/291/582 (pcs) 0.003/0.007/0.002 (sec)
Nodes: 184+0/184 (pcs) 0.008/0.000 (sec)
Labels: 20/1/21 (pcs) 0.002 (sec)
Start rendering timer: 200 ms
Start data loading...
DrawMap()
Draw: [51.50220 N 7.45460 E - 51.52622 N 7.47775 E] 19567x/14 480x800 143.00000 DPI
Paths: 4722/751/751/169 (pcs) 0.003/0.012/0.000/0.009 (sec)
Areas: 1098/291/582 (pcs) 0.003/0.002/0.000 (sec)
Nodes: 184+0/184 (pcs) 0.002/0.000 (sec)
Labels: 20/1/21 (pcs) 0.002 (sec)
focus-out event
Speicherzugriffsfehler (Speicherabzug geschrieben)
```

All that smooth scrolling is irritating for me. Especially since one now sees, how the layouter for street contour labels and plates moves them around during zooming. Here are more robust layouter seem to be required (but that is not your fault ;-)). Also the debug output is now likely too much.

It it possible to switch of (configurable) the smooth zooming effect, so I get one refresh per zoom, movement? While smooth zooming/zooming is fancy it possibly will not help for debugging, so in this case it would be good to temporary disable it (by flag, define,...).

Framstag:
Yes, the crash in the about dialog was there before :-/

Framstag:
Regarding logging: I added now to master methods to query the current logging state. Also by default debug info info is now off, but can be  toggled using Control+Shift+D and Control+Shift+I.

Regarding initial positioning of map on open. To store the position is a good thing. I would however add the name of the map as additional key. So that for every key, you store a separate location. If there is no position for current map, it should default to the solution in master (centred with some medium zoom level).

Karry:
Animations are not configurable now. Animations can be suppressed by simple change in `MoveHandler`.

To logging: Do you want to use osmscout::Logger even in Qt code? In Qt world, there are `qDebug()` and `qWarning()` methods that can be controlled by [system environments](https://woboq.com/blog/nice-debug-output-with-qt.html). Btw, in your recent changes, I have nothing against the idea, but I don't like logging control in MapWidget, it has nothing to do with map. I suggest to move it to Settings (QML entry) or to new element.

I will take a look to initial position. I agree that display empty window is just wrong :-)

Karry:
Do you agree to move toggleDebug / Info to Setting object in this branch?

Framstag:
It does not make sense to two logging frameworks. With the Log framework as part of libosmscout you have the change to still using external APIs for logging (ConsoleLogger is exchangeable). So yes, it make sense to use the libosmsocut logging framework in Qt applications, too. This still allows you to redirect output, even using Qt debugging functions.

Moving the Debugging/Info toggle to the MapWidget was just a quick decision, two show you how it could work. You can move the code of course somewhere else. It may even make sense to define another component for this (because it isn't setting related either).

Karry:
Ok, I agree. Just to recap my TODO list:

 - improve code for initial possition
 - use osmscout logging everywhere, setup redirection to Qt specific functions
 - add basic documentation to classes
 - take a look if we can support map rotation with current code

I will be back in few days...

Framstag:
Just to clarify. TODO list sound a little bit harsh. I'm appreciate your patch, I'm sure it improves OSMScout2  and just want to make sure that we do not break somehow things on that way. OSSMScout2 is the one major example and showcase we currently have :-/

Karry:
Thanks :-) Btw, is doxygen documentation generated in autoconf build?

Framstag:
Yes and no. All directories to be scanned have to be added to the doxygen configuration file.

Issue #23, opened 2016-06-15
Appveyor Windows/MinGW/cmake Build fails


Framstag:
Above combination currently fails, it looks like there is some Appveyor-specific problem with from time to time appears:

```
CMake Error at C:/msys64/mingw64/lib/cmake/Qt5Qml/Qt5Qml_QTcpServerConnectionFactory.cmake:2 (add_library):
  add_library cannot create imported target
  "Qt5::QTcpServerConnectionFactory" because another target with the same
  name already exists.
Call Stack (most recent call first):
  C:/msys64/mingw64/lib/cmake/Qt5Qml/Qt5QmlConfigExtras.cmake:4 (include)
  C:/msys64/mingw64/lib/cmake/Qt5Qml/Qt5QmlConfig.cmake:172 (include)
  cmake/features.cmake:110 (find_package)
  CMakeLists.txt:108 (include)
```

I'm not sure if this is something we have some influence on. Nevertheless a fix or work around would be more than helpful.

PR #161, opened 2016-11-21
Fix av build

rinigus:
This PR should fix https://github.com/Framstag/libosmscout/issues/23 . It seems we had issues caused by 

Qt5 buggy package (similar issues were in Alexpux/MINGW-packages)
String.cpp was missing few constants required in msys2 64-bit

Please check that I used correct constants in String.cpp


Framstag:
Thanks for fixing this long standing bug. The fix might be small, but the failing builds troubled me for a long time :-)

rinigus:
You are welcome! Now, with the builds in green, you could finally sleep during nights :)

PR #163, opened 2016-11-25
Deploy Import tool with every release

rinigus:
This PR adds support for deploying compiled Import tool as a part of each release.

Patches include:

* add a new configuration option for cmake to tailor the compilation and linking for exportable Import tool

* In Windows and Ubuntu, xml2 library is compiled with lzma and iconv (windows only?) dependency. By adding these libraries to the link command, we can fully support OSM imports, not just PBF. In addition, they are linked statically as well - no requirement to deploy these libraries (.so or dll). 

* Travis and Appveyor deployment fields are added. **NB! Some parameters (secret keys and repo) have to be specified for each fork using it.**

* README instructions added for ci to simplify use of the modifications

An example release using these scripts is at https://github.com/rinigus/libosmscout/releases/tag/0.0.git.20161118

Now, the only major missing part, is the driver script for Windows importing. The request for help was sent to the mailing list, but there have been no response yet.

OSX imports are missing as well. Since I don't have any access to that OS, I left it out.


rinigus:
In these commits I added "set -e" to install and build scripts. As a result, they catch errors that were ignored earlier. While cmake installation (while already installed) should not cause failures in OSX anymore, there are errors in compilation that are caught now. 

See the failed tests and compare them with "passed" tests in Travis.

Framstag:
I see that a travis builds fails. Though I do not see, how this correlates?

rinigus:
Let's compare the builds that are failing (https://travis-ci.org/Framstag/libosmscout/jobs/178788218) and successful (https://travis-ci.org/Framstag/libosmscout/jobs/178681295). They both end on something like 

```
  CXX      WStringStringConversion.o
WStringStringConversion.cpp: In function ‘int main()’:
WStringStringConversion.cpp:49:34: error: expected unqualified-id before ‘&’ token
   catch (const std::runtime_error& e) {
                                  ^
WStringStringConversion.cpp:49:34: error: expected ‘)’ before ‘&’ token
WStringStringConversion.cpp:49:34: error: expected ‘{’ before ‘&’ token
WStringStringConversion.cpp:49:36: error: ‘e’ was not declared in this scope
   catch (const std::runtime_error& e) {
                                    ^
WStringStringConversion.cpp:49:37: error: expected ‘;’ before ‘)’ token
   catch (const std::runtime_error& e) {
                                     ^
make[1]: *** [WStringStringConversion.o] Error 1
make[1]: Leaving directory `/home/travis/build/Framstag/libosmscout/libosmscout/tests'
make: *** [check-am] Error 2
Build end time: Thu Nov 24 19:53:50 UTC 2016
```

However, since I added `set -e` in ci/travis/build.sh, the builds are reported as failing after my changes while they were considered OK before. Since its a clear compilation error, handling its as an error in Travis builds is probably correct.

Framstag:
OK, I see. Should have looked into it in more detail :-/ I'll try to fix the build error first.

rinigus:
No problem! I should have written better. And its more efficient if only one really digs into the logs :)

Framstag:
If I understand correctly. I cannot just merge the pull request but I should shortly afterwards add the necessary keys and configurations as described in your README? I'm still looking for 2-3 hours in the afternoon, where I can work through your documentation and fix any resulting problems. I would hate to break the build too long. Please keep patient :-)

rinigus:
No worries, I can wait :)

Regarding the PR - it should all be fine when you merge. Its just additional configuration is needed when you want to publish a release and deploy. All other changes are made to make it possible, but should not interfere in some destructive way.

In Travis and AppVeyor, the deployment is only when you publish new release or set a tag. So, you have time to set the keys and specify the repos until you make such release. I'll be happy to reply to all the questions to make it as easy as I can for you.

As for automatic builds, you would just get an extra check in AppVeyor that would build an importer with every commit.  

For fully configured example, see https://github.com/rinigus/libosmscout . As you could see, there are already few releases and the latest one is now advertised to the users for their importing needs. Not much feedback yet, but the one I was getting so far was positive.

rinigus:
Morning!

great news - we have a script that can be used for importing in Windows! The script has been developed by Michael Nenninger and he will submit is as an PR from https://github.com/DerKleinePunk/libosmscout/tree/windows_import . I have incorporated his commit into my sailfish branch for testing and it worked perfectly.

The changes in README and build script for importer are already in this PR. Since the script will come to you separately, my PR will fail a test on import building in windows until we would have that script in a tree. While its more complicated to have the functionality split into two PRs, I think its better to keep authorship and appreciate the work done by Michael. So, I asked him to submit the PR separately. 

Hopefully its all OK.

Best wishes,

rinigus

OgreTransporter:
Nice! But I'm building libxml2 without iconv and lzma. ;-)

I don't know Travis and AppVeyor, but on my Jenkins CI I pack the CMake output files - Import.exe is already included. Also it is possible to use CPack as part of CMake script to create release packages.

rinigus:
@OgreTransporter: I guess there are many ways to slice a cat :) . I used Travis and AppVeyor since they are a part of CI in libosmscout. And it works quite well...

As for libxml2: would my changes break anything in your build? As far as I understand, libxml2, by default, needs iconv and lzma... By adding them as I did, we can have almost static importers in Linux and Windows. Here, the idea is to have as small number of host PC requirements as possible.

OgreTransporter:
@rinigus: :+1: 
The modification of ```libosmscout-import/CMakeLists.txt``` will break my build, because LZMA is optional but Iconv is required. Maybe I could add Iconv to my CI but I try to drop as many as possible dependencies to reduce size and increase speed. Btw, only Iconv is a default  dependency on Windows:
```
libxml2 version: 2.9.4
Created Makefile.
Created config.h.

XML processor configuration
---------------------------
              Trio: no
     Thread safety: native
        FTP client: yes
       HTTP client: yes
    HTML processor: yes
      C14N support: yes
   Catalog support: yes
   DocBook support: yes
     XPath support: yes
  XPointer support: yes
  XInclude support: yes
     iconv support: yes
     icu   support: no
  iso8859x support: no
      zlib support: no
      lzma support: no
  Debugging module: yes
  Memory debugging: no
 Runtime debugging: no
    Regexp support: yes
    Module support: yes
      Tree support: yes
    Reader support: yes
    Writer support: yes
    Walker support: yes
   Pattern support: yes
      Push support: yes
Validation support: yes
      SAX1 support: yes
    Legacy support: yes
    Output support: yes
XML Schema support: yes
Schematron support: yes
   Python bindings: no

Win32 build configuration
-------------------------
          Compiler: msvc
  C-Runtime option: /MD
    Embed Manifest: no
     Debug symbols: no
    Static xmllint: no
    Install prefix: .
      Put tools in: $(PREFIX)\bin
    Put headers in: $(PREFIX)\include
Put static libs in: $(PREFIX)\lib
Put shared libs in: $(PREFIX)\bin
      Include path: .
          Lib path: .
```
If you like to take care about all dependencies you also have to detect BZIP2 for FreeType. ;-)

rinigus:
Re LZMA: this is odd that it breaks the build. I added tests on what to do if LZMA is found and what to do if its not (see  libosmscout-import/CMakeLists.txt). LZMA seems to be defult in mingw-provided xml2, if I remember correctly.

Would you mind sending the link with the error message?

As for FreeType - its not required for Importer, see the PR subject :)

OgreTransporter:
LZMA is optional - which is not a problem. But iconv is set as a requirement, which breaks my build - because I don't use iconv in libxml2. Is it possible to leave the original code and adding LZMA and Iconv only if available?

rinigus:
@OgreTransporter: yes, it should be possible to have iconv optional as well. Let me add that to the cmake files.

OgreTransporter:
@rinigus: Thanks a lot! :+1:

rinigus:
@OgreTransporter: would you mind checking out if all is OK now with respect to iconv in your builds?

rinigus:
Tim, please note that AppVeyor build failure is due to the missing import.cmd that will come from https://github.com/Framstag/libosmscout/pull/178

OgreTransporter:
@rinigus: I've testet ca19d49 on my Jenkins with Visual Studio (VS2013-x86, VS2013-x64, VS2015-x86, VS2015-x64). VS2013 failed, because bbcb94e is missing. VS2015 is complete!

```cl : Befehlszeile warning D9002: Unbekannte Option "-std=c++11" wird ignoriert.```

Framstag:
Note that libzlib is also used in the pbf preprocesing code. pbf file format also supports bzip2 and lzma, but this currently not used and will result in an error (See preprocessPBF.cpp).

rinigus:
I guess we'll deal with libzlib when the need comes. Thank you for merging it!

rinigus:
Tim,

I can see that clang and website builds are not passing on travis. The errors on which its failing are the same as in the earlier builds (that were marked passed), just now the bash returns error status if something goes wrong. As a result, the test is now correct.

However, to make it work with the corrected tests, we should address them. Do you need help with fixing these errors?

Framstag:
Yes, any help would be ... helpful :-)

I'm currently looking at the web page generation problem. I see that there is also a problem with the agg build parts.

Looking at appveyor (which is slower in builds) there also seems to be a problem with building the importer.

rinigus:
Sure. Appveyor issue will go away as soon as it would start processing
windows import script PR. It fails on copy operation of that script due to
its absence.

I'll look into agg build part.

rinigus

rinigus:
agg build error comes from the agg library itself. I don't think we can fix it easily. I suggest to disable Agg backend building with clang on Linux. How do you disable a part of the library building in autoconf? With cmake its easy, but I don't see what should I replace with "make full". I guess I can make something like

make libosmscout ...

but that does not look too handy.

Framstag:
The build deactivates itself automatically if the configure check fails. We have to extend the configure so that it fails.

rinigus:
OK, I see. I'll try :)

Framstag:
Funny enough the build already breaks while processing the agg includes. The initial check for fonttype support is not different, it just does not include as many agg imports.

rinigus:
Yes, we probably just need to add more headers and it will be all fine. Do you want to take over? Or should I look into how to craft it (haven't done autoconf for a while)

AppVeyor build is passing with your first baked importer at https://ci.appveyor.com/project/Framstag/libosmscout/build/431/job/hnfos38f9rmaxtf7/artifacts

Framstag:
I'm finished for today. Its already 22:30+ here. And Amazon didn't deliver the the 6.12. parcel for my daughter. So we have to improvise **now**.

rinigus:
NP, I'll try to finish this agg business. Good luck!

rinigus:
Tim, I think I solved the agg case. I'll run the tests till the end and then submit the pull request later tonight.

PR #165, opened 2016-11-27
Type definition for pistes, new map style for winter sports

Karry:
Are you prepared for winter? For long pistes in Alps, chilly weather and adrenaline in blood? I finished one idea that I have some weeks in my mind. It was to prepare winter map for OSM Scout, similar to [opensnowmap.org](http://opensnowmap.org).

Type definitions are simple. I created one type for each common difficulty (easy/blue, intermediate/red, advanced/black). I decided to create new map stylesheet because I needed to change some colors, priorities, sizes... And it would be difficult to maintain it in standard stylesheet just with switches and conditions... 

Here are some examples:
winter_map_albrechtice
winter_map_kitzbuheler_horn

### DIFF
--- a/stylesheets/map.ost
+++ b/stylesheets/map.ost
@@ -521,6 +521,22 @@ TYPES
       {Name}
 
   //
+  // Winter sports
+  //
+
+  TYPE piste_downhill_easy
+    = WAY AREA ("piste:type"=="downhill" AND "piste:difficulty"=="easy")
+      {Name}
+
+  TYPE piste_downhill_intermediate
+    = WAY AREA ("piste:type"=="downhill" AND "piste:difficulty"=="intermediate")
+      {Name}
+
+  TYPE piste_downhill_advanced
+    = WAY AREA ("piste:type"=="downhill" AND "piste:difficulty"=="advanced")
+      {Name}
+
+  //
   // Water transport
   // 
 
--- a/stylesheets/winter-sports.oss
+++ b/stylesheets/winter-sports.oss
@@ -482,7 +482,7 @@ STYLE
 
     [MAG city-] {
       [TYPE natural_beach] AREA { color: @beachColor; }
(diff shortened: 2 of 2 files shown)


Karry:
Btw, is there some way to setup area priorities, like for paths? In case that piste is defined as area and it overlaps with another area (small grassland for example), it is overlapped. This is not something that I want in this case... It should be something like overlay for all areas.

It is controlled just by `AreaSorter` in `MapPainter.cpp` ?

Framstag:
Thanks for this addition. I'll merge it. The question however is, how big is the delta? I have some fear that improvements to the standard.oss will not added to this style sheet, too. That is one fo the reason why I added support for variables and conditionals. It sounds like this is not enough. The next step would be to add imports. So that the style sheet can be devided into smaller reusable parts? The alternative would be to manage patch sets for diverting style sheets.

What do you thing? What would be needed to get stylesheets "closer"?

Yes, currently the AreaSorter decides the order of areas drawn. Normally this is defined by containment. (Much) earlier version tried to define a global order but this fail. Example There are situations, where a wood is "on top of" water (island in a lake) and situations where the opposite is true (lake in a wood).

There were wishes to change that, but I would always suggest to enhance AreaSorter locally and play with it to check if the changes are really stable and helpful. I would claim that such things are normally more mapping error and shortfalls in the style sheet.

Karry:
Hi. I checked size delta on Austria (with lots of pistes) and it is less than one percent:

> before = 426567680 #(406 MiB)
> after  = 427079680 #(407 MiB)
> (after - before) / (before / 100) # = 0.12

I don't think that having more separate stylesheets will be problem. From my point of view, it is more simpler to manage various simple stylesheets that duplicates some parts, than one complex style with many conditions. I think that conditions are great for some style nuances like "night theme" or "car in tunel". But for different usecases when you want different maps (tourism, car navigation, winter sports, yacht navigation...) should be different stylesheets...

Issue #168, opened 2016-12-01
Importing large files is not working (Windows)


OgreTransporter:
I tried to import germany-latest.osm.pbf (2.85 GB) but the import tool always claimed that this file does not exist. When debugging, it has come out that under Windows the function stat (osmscout :: ExistsInFilesystem) does not work with such large files.

OgreTransporter:
#169

Framstag:
As this is solved, I'm closing it.

PR #169, opened 2016-12-01
Issue #168 stat is not working under Windows with large files

OgreTransporter:
I tried to import germany-latest.osm.pbf (2.85 GB) but the import tool always claimed that this file does not exist. When debugging, it has come out that under Windows the function stat (osmscout :: ExistsInFilesystem) does not work with such large files. To solve this issue I've changed the preprocessor order in that function to force Windows using GetFileAttributes instead. Another option is to use stat64.


Framstag:
Thanks! I always tested the Windows build with smaller files, so I did not see that.

OgreTransporter:
No problem! Normally, I also test only smaller parts, but now I have a route from NRW to BY therefore I needed the whole map of Germany.

PR #171, opened 2016-12-02
 Add DirectX map render

OgreTransporter:
I've created a basic map renderer for Direct2D.


OgreTransporter:
I had to add some corrections for cmake.

Framstag:
That sound interesting! A DirectX backend would make Windows even more a first class citizen. I see though, that the MinGW build still fails. Does have MinGW support DirectX libraries and headers at all? If not, it is of course OK to disable this part of the build for MinGW.

rinigus:
@OgreTransporter: I have just submitted PR dropping this flag. Sorry for introducing this compiler-specific option - my mistake.

I tested on Sailfish and Linux, all compiled well without it.

OgreTransporter:
@rinigus: Don't worry, I've made a solution to disable this flag for MSVC. I'll change that topic aggain, CMake has a built in support for the right flag. Also the flag should be added to CMAKE_CXX_FLAGS and not as a definition.

@Framstag: I'll disable the default build option for mingw. It's working well with MSVC 2013 and 2015 with DirectX SDK and Windows SDK.

rinigus:
@OgreTransporter, I am sorry if I wasn't clear. Yes, I did see your
solution in your PR. I don't worry - I want to get rid of this flag for
all. See https://github.com/Framstag/libosmscout/pull/173 for proposed
solution. :)

OgreTransporter:
I agree, that
```set_property(TARGET NumberSetPerformance PROPERTY CXX_STANDARD 11)```
is the best solution, but this requires CMake 3.1.x ;-)

Framstag:
Looks like the build is OK now, but there is still a conflict in the CMakeList.txt (likely because of the other C++11 pul request). Cann you please fix the conflict? I will then merge the changes.

Some questions because I cannot quickly test it myself):
* Is the backend feature complete or is it misisng features?
* How fast is it (in comparison to the other backends)?
* How difficult to implement was it? IMHO implementing a new backend is rather easy if you know the backend API. Do you agree?
* It is possible to use the backend also using MinGW or are there principle problems? If the backend should work with MinGW to, I would make an issue to create corresponding autoconf files for it.

OgreTransporter:
> Is the backend feature complete or is it missing features?

It's not complete because my intension was learning how to use Direct2D. DrawContourLabel and DrawContourSymbol are not implementet yet and some street names are not visible at the moment. If I find some time, I'll complete it. Or maybe someone else.

> How fast is it (in comparison to the other backends)?

I've measured DrawMapCairo (343200 µs), DrawMapDirectX (187200 µs) and DrawMapQt (811200 µs) for one frame on 800x600.

> How difficult to implement was it? IMHO implementing a new backend is rather easy if you know the backend API. Do you agree?

I've looked at the existing map painters. It is not that difficult, because you just have to overwrite the given functions of osmscout::MapPainter. The other map painter are showing the data handling, so it is not necessary to learn and understand the whole code in the background.

> It is possible to use the backend also using MinGW or are there principle problems? If the backend should work with MinGW to, I would make an issue to create corresponding autoconf files for it.

I don't know! I'm using MSVC on Windows and GCC or CLANG on Linux, I've never used MinGW. Also, I build everything with CMake, because CMake is working on Windows and Linux. Autoconf is just for Linux. IMHO you should use an optimized matching compiler for the target platform (MSVC => Windows, GCC/CLANG => Linux) and not workarounds like MinGW or Cygwin to compile Windows applications with Linux tools.

PR #173, opened 2016-12-03
drop specific compiler flag c++11

rinigus:
I tested without this definition using new Sailfish SDK and Linux. This flag is not needed. If it has to be introduced, it should be done properly using CMAKE macros. Sorry for this mistake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,8 +24,6 @@ if(APPLE AND NOT ANDROID)
   set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++")
 endif ()
 
-add_definitions(-std=c++11)
-
 include(CheckTypeSize)
 CHECK_TYPE_SIZE("void*" OSMSCOUT_PTR_SIZE BUILTIN_TYPES_ONLY)
 if(OSMSCOUT_PTR_SIZE EQUAL 8)


Framstag:
The C++11 flag is needed for older gcc versions. Depending on the version of the compiler it defaults to different versions of the language specification. Older versions by default using some pre C++11 standard and C++11 must explicitly get activated else you get compiler errors. newer version (likely 6.x?, C++11 is 5 years old!), do not need the flag anymore, they default to C++11 anyway - however the do not error on it, so there is no harm in keeping it there.

For VisualStudio there is likely a similar but differently named flag.

rinigus:
As far as I could see, we have CMAKE set that should do it properly. For
example:

set_property(TARGET NumberSetPerformance PROPERTY CXX_STANDARD 11)

So, specifying it as I did is wrong, we have to use CMAKE macros for it.
They will get the corresponding compiler flag. When we compile right now
without this flag, I am getting -std=gnu++11 as a compiler flag (make
VERBOSE=1)

OgreTransporter:
It's simple to solve: https://github.com/Framstag/libosmscout/pull/171/commits/446d6bd49f2399b1af659403c35ea88dfdf83c22

rinigus:
Its just not needed. We can drop the flag and there should be no problem
with it since we should have all required CMAKE set's in the other
CMakeLists files that make it correctly. I don't remember why exactly did I
add that. At present, I cannot reproduce errors in compilation without this
flag. So, its not needed as far as I can see. If something does not
compile, we should use CMAKE command to specify C++11 requirement, not this
flag (didn't know that earlier)

Karry:
Well, it seems that CXX_STANDARD property is supported from cmake 3.1. But on Debian Jessie is cmake 3.0.2, so this change breaks build in "debian_jessie_gcc_cmake" docker image...

rinigus:
Lukas,

I guess we would have to make a hack that checks CMAKE_MAJOR_VERSION & CMAKE_MINOR_VERSION, compiler and then sets -std=c++11 . Indeed, seems that the correct way is available from 3.1

rinigus

Framstag:
Seems like we have a problem with the long release cycles of debian and the resulting old versions of some artefacts. I do not have a problem with solving this specific problem with the version of cmake by using a nerwer version from debian backports. I assume, that most people would do this anyway.

See https://packages.debian.org/search?keywords=cmake&searchon=names&section=all&suite=jessie-backports

Karry:
it should be fixed by this MR: https://github.com/Framstag/libosmscout.git

PR #179, opened 2016-12-05
configure: test for compilation with agg headers

rinigus:
Tim,

I have not tested it on Travis. This PR maybe faulty and we should just cancel it if needed.

I am sending this as PR to trigger autobuilds. Let's see how it will go. I am off for today as well, getting late.


rinigus:
didn't work, closing it. I will have time to check it in the evening

rinigus:
This should put Travis in green, except for the web page. I fixed also CoordinateCalculations.cpp bug that got complaints from one of the builds. 

There are a bit too many commits. If you wish, I can make a new request with just the final changes. But that would have to wait for longer. Have to run, hopefully all the builds will turn out fine...

Framstag:
The Webpage problem is also strange. I have to use a local docker image to try to reproduce the problem. On my local system I have no such problems with similar packages (Arch Linux).

rinigus:
But do you get the messages like 

https://travis-ci.org/Framstag/libosmscout/jobs/181707375#L2095

in Arch?

rinigus:
We can always cheat with this lexer error by changing corresponding line in a script to

hugo --verbose || true

At least then the webpage will be up-to-date. Problem with this solution - we'll happily forget about it and  this error will stay for a while.

PR #183, opened 2016-12-06
Fix unknown compiler option (MSVC)

OgreTransporter:
Visual Studio below 2015 doesn't have such a flag. Since 2015 the flag is set to newest standard as default (/std:c++latest).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,9 +27,11 @@ if(APPLE AND NOT ANDROID)
 endif ()
 
 # CXX_STANDARD property is supported from cmake 3.1, we have to define -std with old cmake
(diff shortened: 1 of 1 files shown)


Karry:
LGTM. Sorry for complications. Another option is to drop support for old cmake (< 3.1). But this workaroud is still ok from my point of view...

OgreTransporter:
My fist CMake version was set to 3.1, but it has been changed to 2.8 for sailfish. I don't care about the version as long as I can build it ;-)

Please, do not add global compiler flags! Each compiler is using specific flags.

Framstag:
Yes, instead of a !MSVC should a GCC or CLANG check not be better (saver)?

OgreTransporter:
That's right. I don't know anything about the flags of other CMake supported Compilers (Intel, Borland, etc.). But it's not possible to setup compiler settings for one of these compilers without an installation of them.

PR #185, opened 2016-12-11
Map management

Karry:
Hi Tim, sorry for this huge MR (again), but single class don't give complete picture...

**Short story:** These changes adds ability to view available maps on server and download it to local computer. Picture better thousand words:
map_downloader

## Long story:
new main classes:
**AvailableMapsModel**
Tree model with maps available by configured providers (see Settings::GetMapProviders).
Every map provider have to expose list of maps by json.
**FileDownloadJob**
Simple utility class for download single file over http. It don't support downloading restart when connection is interrupted.
**MapDownloadJob**
Utility class for downloading map database described by AvailableMapsModelMap over http. When connection is interrupted, downloading will be retried after some backoff period.
**MapManager**
Manager of map databases. It provide database lookup (in databaseDirectories) and simple scheduler for downloading maps.
**MapDownloadsModel**
QML list model with currently downloaded maps. It provide methods (invocable from QML) for starting new map download.


Framstag:
Is it possible that some builds break because you did not extend the autoconf based Makefiles with the new source files?

Karry:
Build should be ok now. I have to add one `#if` for older Qt versions...

libosmscout-client-qt/src/osmscout/DBThread.cpp
```diff
@@ -124,15 +131,21 @@ DBThread::DBThread(QStringList databaseLookupDirs,
...
 DBThread::~DBThread()
 {
   osmscout::log.Debug() << "DBThread::~TiledDBThread()";
 
   for (auto db:databases){
-    db->mapService->DeregisterTileStateCallback(db->callbackId);
+    db->close();
   }
+  databases.clear();
+  delete mapManager;
```
Framstag: See comment above. This code could be dropped then.

libosmscout-client-qt/src/osmscout/AvailableMapsModel.cpp
```diff
@@ -18,6 +18,51 @@
...
+size_t AvailableMapsModelMap::getSize() const
+{
+  return size;
+}
+
+QString AvailableMapsModelMap::getSizeHuman() const
+{
+  double num = size;
+  QStringList list;
+  list << "KiB" << "MiB" << "GiB" << "TiB";
```
Framstag: We already have an libosmscout function for this. See String.h ByteSizeToString().

Framstag:
Just some minor comments, see inline comments to code changes.

Karry:
Thank you for comments. I will update this branch tomorrow.

Karry:
Hi Tim. I updated code and OSMScout2 readme: https://github.com/karry-space-with-my-second-forks/libosmscout/blob/635bc2d8cfd8af7174c6f16d8554ffc867325345/OSMScout2/README.md

Btw, `AvailableMapsModel` is still using pointers for internal model, but it is "recomended" way how to implement Qt models and its model object pointers are not available outside model... (You should not access index.internalPointer() outside the model :-) )

PR #186, opened 2016-12-11
Fix two errors in qt client lib

OgreTransporter:
Error 1: Missing marisa include directory for the project. DBThread.cpp includes TextSearchIndex.h which includes marisa.h
Error 2: M_E is not declared on all systems. Including cmath is only working in combination with _USE_MATH_DEFINES


Karry:
LGTM. Thanks you for testing. You are using MSVC compiler right?

OgreTransporter:
No problem! Yes, Win7+VS2013 and Win10+VS2015. I'm building x86 and x64 configurations on both systems with Jenkins-CI.

Issue #189, opened 2016-12-19
Render sea/land on database polygon only


Karry:
Hi. Current client-qt code tries to combine more map sources. Multiple osmscout databases or even online map (tiled DB thread). It is using database bounding box to determine what source covers given area. This aproach is good approximation, but it fails when you use databases for neighboring countries, like Czechia and Austria...

It would be great when would be possible query to database polygon (Covers database this bounding box?) and sea/land will be rendered inside this polygon only.

It is possible to store such information to water index without changing format? I have limited knowledge about this data file, but it seems to be right place to me...

Framstag:
The WaterIndex offers cells for various zoom levels where each cell has either the state "water", "ground" or "unknown".

What you are likely are interested is the state "unknown" that is either set, if there is no data, it is also returned outside the bounding box of the database (which is a simple rectangle). It can also be returned if there are big holes in the data (the "is this cell water or ground" has some configurable limits).

If you do not render "unknown" tiles or only if there is no database covering this cell, the result should already be better.

For implementation of cross-database routing we need similar information. Cross-database routing will only work well, if there is some certain overlap between database borders. For To know in which database to look for a routing node, we need some coverage map. I'm currently thinking of a simple bitmap, where a bit signals that there is at least one routing node in this cell. So if we are currently following the A* routing graph we can check in which database we have to look for (partial) routing data for a given (route)node.

Only open question is, how big should the bitmap cells be (zoom level) to fulfil the following requirements:
* Cells should not be too big, so that we do not always query all routing databases but only in border regions.
* Cells should not be too small, because the bitmap needs a certain disk space

After these basic calculations the implementation of such index should be simple. It though would only return information, if the cell contains route nodes. A similar index with more general information is though possible, too.

Does this help you?

Framstag:
Can we close this now after we merged?

Karry:
Sure, we can.

PR #190, opened 2016-12-28
Import data polygon, crop waterindex by it

Karry:
Hi Tim.

This work is still in progress, please don't merge it yet. I just want to discuss it in early development stage.

I want to solve multi-database rendering issues described here https://github.com/Framstag/libosmscout/issues/189 As a solution, I want to read database polygon from ***.poly** file, write polygons to new `rawdatapolygon.dat` file (similar with `rawcoastline.dat`) and use it in `WaterIndexGenerator` for clipping land/sea just for area where database is defined...

Is this idea correct? waterindex can be used for test if some point belongs to specific database then. It may helps to multi-database routing in the future.

Currently only `PreprocessPoly` class is implemented. Here is rendered test area with style for "datapolygon" area:
datapolygon


Framstag:
Evaluating polygon files is a very good idea. The code should however have a fallback if polygons are not available (not all *.osm or *.osm.pbf sources will offer one) and then try to estimate the polygon by looking at the existing data (outer hull).

Note though that the Water Index has a limited set of states (4). Further states would require additional bits to get allocated and since you then need at least 3 bits which then does not divide good in relation to the 8 bits of a byte.

Water index also builds the index for multiple levels (though only for a subset of all possible levels). 

I'm not sure if a separate index would not be more helpful. IT would also be interesting if it would be helpful to approximate the polygon by a (small) number of polygons. These could be loaded into memory completely and would be a much faster index to look up then. Bit a bitmap based index would still be very fast to access.

Karry:
To separate index - I think that it would be great to have one general index for "huge areas", like land/sea, datapolygon, administrative areas... Something that will allow us fast rendering of country boundaries or coastlines on all zoom levels. But I have only fuzzy idea how to implement such data structure...

Framstag:
Hmmm, you are (possibly) mixing a number of things:
* You are currently working an a index (I assume), that will tell you, if a region is covered by the database or not. This is a binary index (yes or no), which is not bound to any types in the map.ost.
* Yes, it makes sense to have a special index for big objects. The OptimizedArea and OptimizedWay index though already have this role, offering reduced shapes for such big objects. Here the problem is, how to we get an import that only contains "big objects" to get an world-wide overview? And how fill these index structure sbehave if they index the whole world?
* Coastlines are already handled by the WaterIndex. It has special data structures to render coastlines in any zoom level fast. Here one needs an additional data file that contains the coastlines for the whole world to get continental shapes and inject this file into the importer. You could also parse foreign shape files but you likely will get problems with alignment with the OSM data in this case. Again: How will the WaterIndex behave in this case?
* Regarding borderlines and administrative boundaries: For these (and other similar types like postal code areas)  we could use a similar approach like it was used for the coastlines in the WaterIndex. There is some differences though: WaterIndex supports only one type (coastline), the new index must support any number of types. It must also handle at least disputed boundaries and some more attributes like name of the neighbouring regions. For administrative boundaries one must be aware that there are multiple levels. The index should be able to return only the "highest" boundary for a given region.

In all cases I can give help. A bitmap index for simple data coverage is simple to implement. I can help you with this one. Injecting external sources for coastline should not be too complicated. I can help with this, too. A borderline, administrative boundary index is complex. There are parts of the WaterIndex code that are not "obvious" and I would have to work on them again, to at least document them better.

rinigus:
Just want to add few bits into the discussion. I think that having these polygons is vital for getting the multi-database support working. Most of our OSM imports are coming from Geofabrik, I think, which provides such polygons. For the outside cases, we may be fine, for the time being, just if we generate the polygon on the basis of the bounding box. If there is a big fat warning during import stating that the performance would be sub-optimal without polygon and the polygon is generated then the users should notice it.

rinigus:
While I was working on planet chopping with osmosis, I stumbled upon a page stating that there is a program that can calculate POLY file on the basis of PBF. Somehow, I cannot find it anymore. I'll try to look for it again. However, that means that we should be able to assume that POLY file is always available and we can use it not as an optional, but expected part in the future formats.

Karry:
Hi. After many, many night hours, I thing that it is ready for testing finally! Sorry that it is so huge, but I had to do many changes in `WaterIndexGenerator`. I hope that it is more readable now...

Italy before these changes:
italy original

After changes:
italy 4

There are remaining problems with coastlines, I will create separate issue for it...

aquiles2k:
That looks very promising! I assume this only clips coastlines and not yet e.g. administrative borders like country borders? Nevertheless great work!

Karry:
Yes, it clips only way coastlines and prevent to fill water tiles outside data polygon. If there is complete island area outside data polygon, it is keep.

Karry:
sorry for spam, but I have to share these screenshots... when sea/land is rendered on data polygon only, there is finally possible to combine online tiles with offline map with just a few visual errors...

czech-german boundary by osm mapnik:
cze-de-mapnik
offline osmscout:
cze-de-osmscout
combined:
cze-de-both

Framstag:
So, how to procede? I sugvest

Framstag:
Sorry, should not use mobile 😔. How to procede? Announce it on the list and request testing before merge?

rinigus:
Would be great to know the plans and how does it fit into the big picture. Namely, I presume that this PR is required for properly drawing multiple maps but does not fully solve it. Which pieces are required in addition when this will land into the master? I am planning to look into rendering of the labels with the hope of reducing the maze of labels written on top of each other - let's see if I can make it though.

Karry:
Well, there will be always something to improve (with multiple databases). This MR improves rendering on top of online map, and multiple databases in "layers". Next step may be merging of MapData before rendering...

@Framstag: I plan to regenerate Europe countries, we can start bug hunting then :-)

rinigus:
@Karry, does it mean that you want to test this merge request via European import?

Would be great to get some pointers regarding how to use this approach to render multiple maps for those who don't use qt client, but libosmscout-map(-qt) directly. Or is it covered in documentation already?

Karry:
Yes, I want to fix this one: https://github.com/Framstag/libosmscout/issues/282, then process Europe and wait a few days if someone will found some another coastline error... Or it can be merged to the master and bugs may be solved later... I hope that it don't creates any regression. What do you think Tim?

To documentation, I believe that best documentation is the code ;-) I am looking to the documentation when code is not clear for me. But if you think that it should be documented somewhere, I am not against it.

rinigus:
Agreed regarding documentation vs code. Except, would be good to know what do the slots and methods do in TiledDBThread / PlaneDBThread. Doxygen is of great help and adding help strings to the methods / slots would simplify code reading a lot. For example, its not clear which method to look for implementation: DrawMap?

Karry:
I found another coastline bug in Albania... I will try to fix it after weekend...

Karry:
Update: problems Albania and Croatia coastlines was fixed. There is last (I hope) corner case that is not handled - coastline path that is fully contained in cell... I will continue tonight...

rinigus:
@Karry, great work! I stumbled on description of how OSM is rendering their maps and, in particular, coastlines. See https://wiki.openstreetmap.org/wiki/Coastline , Rendering in Standard tile layer on openstreetmap.org. From this description I conclude that they don't trust auto-generated coastlines and check the imports before publishing them. This makes me wonder whether the issues will be coming up regularly if you try to detect-fix them automatically.

Karry:
Hi @rinigus . It depends on what kind of errors will arrive. `WaterIndexGenerator` is now able to fix and handle non-simple coastlines (path intersects itself), but expect that two coastlines don't intersects each other (island coastline don't itntersect another island/land)... There is infinity ways how coastlines can be broken :-)

Karry:
Hi Tim. It seems that noone test it or there are no significant errors anymore :-) I suggest to merge it. The last commit fixes last major problems that I found in Russia and India. 

There will be always something to improve, but I would like to close this topic and move my energy to routing/navigation...

Framstag:
OK. Can you summarize in a few sentences what you did? Do we need to add some documentation somewhere to actively make use of the feature? From the comments of the individual check in comments it does not sound like you already have added documentation and I assume that documentation would be helpful to make use of the new feature :-)

A possible location for documentation would be http://libosmscout.sourceforge.net/tutorials/importing/

Karry:
You are correct, I added one paragraph to Import tutorial... 

What I did: 

 * Import tool may read `*.poly` file as input now, if defined, it is used as part of coastlines processing in `WaterIndexGenerator`. Water and land boundaries are cropped by this (multi) polygon then.
 * While generating water index, there is new check that coastline (simplified  for low zoom levels) is still simple - don't intersects with itself. There is even check that simplified coastlines of islands don't intersects with land. These checks avoids major coastlines errors, because following algorithms rely on the fact that coastlines don't intersects with each other.

Framstag:
You should take the honour and advertise your new feature and its merge on the mailing list :-)

PR #194, opened 2017-01-03
Qt: drop too much squiggly contour labels, flips upward labels

Karry:
squiggly threshold is hardcoded to 45° of angle variance

drop-squiggly-labels


Karry:
second commit flips upward labels

avoid-upwards-labels

And fixes rendering of right-to-left contour labels (used in Israel). But it don't fix bi-direction labels (for example mixed latin with hebrew) https://github.com/Framstag/libosmscout/issues/196 How common are these labels in OSM?

libosmscout-map-qt/src/osmscout/MapPainterQt.cpp
```diff
@@ -480,10 +484,41 @@ namespace osmscout {
 
     QTransform tran;
 
+    // precompute character widths
+    std::vector<qreal> characterWidths(string.size());
+    for (int i=0; i<string.size(); i++) {
+      characterWidths[i]=metrics.width(string[i]);
+    }
+
     while (offset<p.length()) {
```
Framstag: Isn't this possibly too complicated? We already know the length of the way, we know the length (width) of the string. Isn't there a simpler way to check for "no angle for the next  x pixels"? This would help come around the character!=glyph problem. Also calculating the width of a character is already a medium expensive thing to do.

Karry:
Hi Tim. I implemented contour labels rendering with using of `QTextLayout::glyphRuns`, it handles bi-direction strings correctly (https://github.com/Framstag/libosmscout/issues/196). This branch includes rendering test from https://github.com/Framstag/libosmscout/pull/198, you can close that request and merge only this one.

libosmscout-map-qt/src/osmscout/MapPainterQt.cpp
```diff
@@ -430,6 +431,9 @@ namespace osmscout {
     QFontMetricsF metrics=QFontMetricsF(font,painter->device());
     QString       string=QString::fromUtf8(text.c_str());
     double        stringLength=metrics.width(string);
+    if (string.isEmpty()){
```
Karry: Ok, I removed this check

libosmscout-map-qt/src/osmscout/MapPainterQt.cpp
```diff
@@ -480,10 +484,41 @@ namespace osmscout {
 
     QTransform tran;
 
+    // precompute character widths
```
Karry: Resolved with `QTextLayout::glyphRuns`...

libosmscout-map-qt/src/osmscout/MapPainterQt.cpp
Karry: I move this logic to `SimplifiedPath::TestAngleVariance`. It may be rewrited when this method will drop too many labels...

Issue #196, opened 2017-01-04
Qt: contour labels with bi-directional text are rendered incorrectly


Karry:
This example is right-to-left (Hebrew?) label, it should be fixed by https://github.com/Framstag/libosmscout/pull/194 Top part is osmscout, bottom is Mapnik... (somewhere in Israel) Can be labels in OSM bi-directional?

![bad-right-to-left-labels](https://cloud.githubusercontent.com/assets/309458/21628631/317f99d4-d221-11e6-8c12-71680530ef41.png)

Bidirectional text rules are complicated, we should use some library function for render labels correctly. Node labels should be rendered correctly (QPainter should support it), QPainter source code may provide some hints howto implement it: https://gitlab.com/pteam/korvins-qtbase/blob/a718a99438aaca7d4cd4379726a8e131d3c4bf89/src/gui/painting/qpainter.cpp#L5825

Framstag:
I urge to make sure that the rendering backend text engine (Pango, Qt,...) should handle this. For complex paragraph rendering the text engine must know if a part is left-to-right or right-to-left. If we handle as bounding box to engine should just fo "the right thing". Important for this is that we handle text always a (one line) paragraph and not as a sequence of single characters.

Karry:
Hi Tim. I agree, complexity of non-latin text rendering can be crazy, this is not something that we (at least I) want to solve :-) But when you write that we should not handle it character by character, how do you want to solve contour labels? You certainly know that Qt don't have API for text rendering around curve. I browse Qt documentation and source code and can't found easy way how to do it correctly with public Qt API. `QTextEngine::bidiReorder` is something what we need to use, but `QTextEngine` is private API :-( 

`QTextLayout::glyphRuns` may be used for it, but this note is scary:
> This is an expensive function, and should not be called in a time sensitive context.
Should I create benchmark / test demo code for that?

Karry:
Side question: how many labels in OSM data uses bi-direction text? We are rendering righ-to-left street names incorrectly and noone reported that. I suggest to add simple support for "all righ-to-left" labels and keep this issue open to signalize that we know about this problem...

Framstag:
[...]

PR #198, opened 2017-01-06
add test that demostrate rendering of contour labels (Qt)

Karry:
It compute and prints fps (rendering frames per second)
Two methods are supported:

 - simple, that render character by character as it is in string
 - bidirectional, that supports bi-directional strings

bi-direction-contour-rendering

___
I think that we can close this https://github.com/Framstag/libosmscout/issues/196#issuecomment-270466014 discussion (for Qt). This test demonstrates bi-direction text rendering "on curve" with Qt. It is even a bit faster than current implementation. I will update pull request https://github.com/Framstag/libosmscout/pull/194 morning (or at Saturday).


Karry:
Tim, can you help me with autoconf build? I don't know where can be the problem...

When I run this test and inspect it with `perf top`, it seems that most expensive parts are `QPainterPath` methods, not text rendering! I will try to optimise it later. I may helps to rendering speed.

Framstag:
Regarding autoconf: For adding headers you can use the ..._HEADERS Makefile variable. See libosmscout/include/Makefile.am

I assume that the build fails though because you use a 
```
#include <DrawWindow..h>
```

include. but there is no automatic search path for the include directory. You can either move the include in the same directory as the source is and use "DrawWindow.h" or must explicitely set a include path like it is done for libosmscout:

```
AM_CPPFLAGS = $(LIB_CXXFLAGS)  -I$(top_srcdir)/include
```

I haven't actually tested it though.

Framstag:
If the performance is better and there is not strong limit regarding the to be used Qt version the change is of course OK for me.

Thanks for working on this. More performance an fat the same time better language support. Cool!

Issue #105, opened 2016-09-04
Improve clipping contour label rendering


Framstag:
Currently all contour labels are rendered in case the bounding box of a way intersects with the visible area.

Since for big ways the bounding boxes may interleave though the way is not visible at all, because it is just meandering around the visible area but not crossing it, we should find a better algorithm for intersection and label clipping.

Karry:
I think that we can close it. At least for Qt we have implementation that check possible visibility for each glyph before renderign.

Framstag:
Yes, of course. All backends should use a similar trick. It would be nice, if the required functionality would be part of the base class, so it is much easier for backends to make use of this feature.

Karry:
I will try to rewrite label layouter to use glyph as base unit for all backends (except OpenGL and DirectX probably).

Karry:
I am closing this issue. All glyphs outside visible  area (layouter viewport more precisely) are skipped in `Layouter::RegisterContourLabel` and are not rendered now.

PR #200, opened 2017-01-08
Optimise contour labels

Karry:
Hi. Until I have this part in my memory, I looked to open issues related with contour labels. I found this one: https://github.com/Framstag/libosmscout/issues/105 For Qt, it is now simple to drop glyphs that can't be visible. This is done in first commit. 

When I inspect rendering with `perf top`, I saw `SimplifiedPath::segmentBefore` method with ~5% cpu usage. This is improved by second commit that avoid iteration over all segments...

Then I used updated `Demos/PerformanceTest` (rendering was repeated 100x) to measure how was performance changed.

```
zoom level => improvement
16 => 2%
17 => 16%
18 => 38%
19 => 46%
20 => 37%
```

used arguments:
```
./build-$branch/Demos/PerformanceTest \
  ../europe-czech-republic-20161203-225207 \
  stylesheets/standard.oss   \
  50.088 14.418 50.086 14.421   \
  16 20  \
   256 256   \
  Qt
```

### DIFF
--- a/libosmscout-map-qt/include/osmscout/SimplifiedPath.h
+++ b/libosmscout-map-qt/include/osmscout/SimplifiedPath.h
@@ -39,6 +39,7 @@ namespace osmscout {
   private:
     qreal length;
     QVector<Segment> segments;
+    QVector<int> offsetIndex; // segment offset by length 100
     qreal minSegmentLength;
     QPointF end;
     qreal endDistance;
--- a/libosmscout-map-qt/src/osmscout/MapPainterQt.cpp
+++ b/libosmscout-map-qt/src/osmscout/MapPainterQt.cpp
@@ -525,12 +525,19 @@ namespace osmscout {
           auto pos=glypRun.positions().at(g);
           indexes[0]=index;
           positions[0]=QPointF(0,pos.y());
+          QRectF boundingRect=glypRun.rawFont().boundingRect(index);
 
           qreal glyphOffset=upwards? (offset+stringWidth-pos.x()) : offset+pos.x();
           if (glyphOffset>pLength)
             continue;
 
           QPointF point=p.PointAtLength(glyphOffset);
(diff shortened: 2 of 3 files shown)


Framstag:
That sounds great. I'm just waiting for the patch to cleanly build...

Some more thoughts on contour labels in general:
We currently draw labels from the start of the way to the end. For each individual label instance (and now if I understand the patch correctly for each glyph) we check if this is visible.

This is algorithmically always an non-optimal approach, because for a very long way we do many unnecessary  checks. A human we always look at the rectangle bounding box of the visible area and would draw labels only within the visible part. This is though too simple, because the human is very good at handling the case where the way enters and leaves the visible rectangle multiple times. And he will likely do not correctly calculate the offset of the label at the point where the way enters the visible rectangle (it is wrong to start drawing the label at the entry point). So human is implicitly dropping much of the actual complexity.

Still this is a hint that there might be a totally different approach, maybe only for fallback situations where the way crosses the bounding box at all.

Also we currently do not render contour labels nicely when zooming in/out. Labels in this case are jumping back and forth because the label position is defined by the size of the way on map, which is changing during zooming in/out.

To have more stable labels we would have to "pin" the labels somehow. So labels always have the same position on the map and overlapping labels during zoom in are just dropped.

When I thing of "pin" I'm always thinking about a grid that is placed on the map and where the labels position themselves in relation to. Since the size of the grid changes with zooming in/out labels would still move somehow, but perhaps the visual effect is more pleasing for the eye.

Fixing labels in relation to some grid would also help regarding clipping. I then would always only look at the cells that are visible.

This is in now way thought through and only some intuition based thought. But nevertheless, perhaps It creates an idea for further optimizations ;-)

The only thing it will definitely break: Beautiful and artful map rendering where ways are just show by the rendered label on them (with no spacing and big fat letters).

PR #205, opened 2017-01-14
Rendering improvements with Qt client

Karry:
Hi, these changes improves rendering of multiple databases for both methods - plane and tiled.

It is not perfect still, area around boundaries may be hidden, but I hope that it will be improved with https://github.com/Framstag/libosmscout/pull/190

multiple-dbs


Karry:
Do you mean the last commit? It is just a optimisation to avoid "unknown" area on the view corner when you drag the map... Yes, render just a background (sea/land/unknown) is also possible. But this solution is much simpler and should not be so expensive. 

Do you want this "crossover factor" (in percent or in pixels) configurable?

Framstag:
I was unsure from the quick cod review if this result only in filling a larger area with ground tiles or if the area that is completely rendered was increased. In the second case things will get slower if the rendered area is increased.

Framstag:
I pushed a variant of the above code to github. Please test.

PR #212, opened 2017-01-16
Label word wrap

Karry:
Thank you Tim that you worked on support of multi-line labels! It was something that was missing to me.

I want to setup label word wrap in Qt/client code to some reasonable default. But configure just maximum label length don't provide enough flexibility. In many cases, I want to wrap labels to fit mobile screen width (even if it is rotated to landscape). For many areas make sense to fit labels even to area bounding box...

For that reason, I added extra `MapParameter` properties:
```
size_t labelLineCharCount;        //!< Labels will be word wrapped if  they are longer then the given characters
size_t labelLineMinCharCount;     //!< Labels will be _never_ word wrapped if they are shorter then the given characters
size_t labelLineMaxCharCount;     //!< Labels will be word wrapped if they are longer then the given characters
bool   labelLineFitToArea;        //!< Labels will be word wrapped to fit object area
double labelLineFitToWidth;       //!< Labels will be word wrapped to fit given width in pixels
```

There is a new utility method `MapPainter::proposedLabelWidth` that compute proposed label width. It is used in `MapPainterQt` and `MapPainterCairo`...

### DIFF
--- a/libosmscout-client-qt/include/osmscout/PlaneDBThread.h
+++ b/libosmscout-client-qt/include/osmscout/PlaneDBThread.h
@@ -62,6 +62,7 @@ public slots:
   void onStylesheetFilenameChanged();
 
 private:
+  double                        canvasOverrun; // scale of rendered canvas, relative to screen dimensions
   osmscout::MercatorProjection  projection;
 
   mutable QMutex                lastRequestMutex;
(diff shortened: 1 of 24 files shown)


Framstag:
This one seems to break the iOSX backend because of signature changes. See the travis build (appveyor seem to have operation problems, I might look at that in the evening). Can you take a look at it. I fixed the iOSX backend in the past blindly if the change was trivial :-/

PR #213, opened 2017-01-18
fix DrawMapSVG demo

Karry:
I just wanted to compare our svg output with https://github.com/severak/lunarender and found that this demo is not functional...


Framstag:
I stopped working on the SVG backend due to lag of time and because I was not sure about the goals. There are different was to generate SVGs with different intentions. I think though that with medium effort one get reasonable result again. But things get tricky if you want to be able to later change individual properties.

PR #214, opened 2017-01-19
return from PlaneDBThread::DrawMap method early when there is no database

Karry:
It solves crash when you run application without database...


Framstag:
The Appveyor builds fails because of your previous change regarding label rendeirng. Could you take a look at it. Thus merging this pull request anyway...

Karry:
ok, I will look at it

PR #216, opened 2017-01-20
Text dimension fixes

Karry:
Patch from Vladimir and one update for JNI Painter

### DIFF
--- a/Android/OsmScoutLib/jni/include/jniMapPainterCanvas.h
+++ b/Android/OsmScoutLib/jni/include/jniMapPainterCanvas.h
@@ -83,7 +83,9 @@ namespace osmscout {
     bool HasPattern(const MapParameter& parameter,
                     const FillStyle& style);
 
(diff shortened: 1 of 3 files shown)


Framstag:
The Android JNI was handwritten. It likely currently does not compile because the libosmscout API has change since then and the JNI was not adapted. IMHO the better approach is to use SWIG to generate the JNI code. This is what liibsmscout-binding is for. SWIG on the other hand currently has the problem that is does not support or stdandard containers, especially the c++11 and higher versions. Thus not all API are currently useable. For some APIs it should be possible to offer alternatives are change the API to be compatible to SWIG, for some of them this will not work.

I assume that an SWIG API for MapPainter itself is possible and could writing an Android backend. I assume though that it will not be very fast because the is some performance loss on Java<=>C++ interaction and the MapPainter has many of them. For this one has to find an alternative anyway.

PR #228, opened 2017-01-29
Raise the FileWriter::MAX_NODES limit to 0x03FFFFFF (26 bits) to allo…

vyskocil:
…w more nodes in large multi-polygon like country borders

### DIFF
--- a/libosmscout/src/osmscout/util/FileScanner.cpp
+++ b/libosmscout/src/osmscout/util/FileScanner.cpp
@@ -2194,7 +2194,13 @@ namespace osmscout {
         if ((sizeByte & 0x80) != 0) {
           Read(sizeByte);
 
-          nodeCount|=sizeByte << 11;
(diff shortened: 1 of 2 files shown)


Framstag:
Hello Vladimir, doesn't the FileScanner needs a change, too? IT has to load the data back you have written.

Framstag:
See FileScanner::Read(). It reads only 3 bytes for the length. The code thus does not crash but it would not read all the coordinates because the nodeCount is smaller than it should?

Interesting enough it should crash if a data file is read sequentially because it would interpret the rest of the coordinates not read as begin of the next data entry...hmmm...strange.

Framstag:
OK :-) I wait for the central builds to finish and will then merge the request

Karry:
Looks good. But it seems to me that this change is not backward compatible - this code will not work with current databases. We should increase database version.

Framstag:
Right.

PR #231, opened 2017-01-31
Fix StyleEditor compile on Mac OSX, add libosmscout-client-qt.pro

vyskocil:
On Mac OSX using pkconfig with Qt is complicate, allow the build of StyleEditor without it


Framstag:
It looks like this broke some buids. Though I'm not sure how changing the StyleEditor project setup should influence it. Looks like possibly appveyor did some changes to the build environment. Looks like the cmake build now try to do an autoconf based build!?

PR #232, opened 2017-02-01
Data object cache

Karry:
Hi. When one way or area intersects more (cache) tiles, it is loaded twice into memory. Similar situation happens on zoom-out when cache tile can't be prefill. These duplicated loads cost some of cpu and memory. This MR adds cache to data files (map from `FileOffset` to object).

I run some performance tests on phone with armv7 cpu: https://gist.github.com/Karry/184ee8acefc5ff5cc0e1217faaeb5aba

When only a few cache tiles is loaded (zoom level <16), there is measurable overhead of data cache and no memory benefits (performance test don't print how many cache tiles was used). But when more cache tiles have to be loaded, there are benefits - significant memory and load time reduction.

### DIFF
--- a/libosmscout/include/osmscout/DataFile.h
+++ b/libosmscout/include/osmscout/DataFile.h
@@ -150,7 +150,7 @@ namespace osmscout {
   /**
    * Read one data value from the given file offset.
    *
-   * Method is thread-safe.
+   * Method is NOT thread-safe.
    */
   template <class N>
   bool DataFile<N>::ReadData(const TypeConfig& typeConfig,
@@ -158,8 +158,6 @@ namespace osmscout {
                              FileOffset offset,
                              N& data) const
   {
-    std::lock_guard<std::mutex> lock(accessMutex);
-
     try {
       scanner.SetPos(offset);
 
@@ -177,7 +175,7 @@ namespace osmscout {
   /**
    * Read one data value from the current position of the stream
    *
-   * Method is not thread-safe.
+   * Method is NOT thread-safe.
    */
   template <class N>
   bool DataFile<N>::ReadData(const TypeConfig& typeConfig,
--- a/libosmscout/include/osmscout/DataFile.h
+++ b/libosmscout/include/osmscout/DataFile.h
@@ -72,10 +72,16 @@ namespace osmscout {
   {
   public:
     typedef std::shared_ptr<N> ValueType;
+    typedef Cache<FileOffset,std::shared_ptr<N>> ValueCache;
+
+    typedef typename Cache<FileOffset,ValueType>::CacheEntry ValueCacheEntry;
+    typedef typename Cache<FileOffset,ValueType>::CacheRef ValueCacheRef;
 
   private:
     std::string         datafile;        //!< Basename part of the data file name
     std::string         datafilename;    //!< complete filename for data file
+
+    mutable ValueCache  cache;
 
     mutable FileScanner scanner;         //!< File stream to the data file
 
@@ -94,7 +100,7 @@ namespace osmscout {
                   N& data) const;
 
   public:
-    DataFile(const std::string& datafile);
+    DataFile(const std::string& datafile, size_t cacheSize=10000);
 
     virtual ~DataFile();
 
@@ -127,8 +133,8 @@ namespace osmscout {
   };
 
   template <class N>
-  DataFile<N>::DataFile(const std::string& datafile)
-  : datafile(datafile)
+  DataFile<N>::DataFile(const std::string& datafile, size_t cacheSize)
+  : datafile(datafile),cache(cacheSize)
   {
     // no code
   }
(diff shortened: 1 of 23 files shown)


Framstag:
Hmmm...From the first glance over the patch I must admit that older versions of libosmscout had such cached, but I removed it a while ago since I hoped that the tile cache was the better solution. I had the feeling that - at least from the performance view and on my desktop system - the old cache did not help improve performance. I did not check for memory consumption reduction at that time. It looks like the cache implementation and cleanup strategy was not optimal. I also assumed that handling multiple caches at different data flow location would make it difficult to control memory consumption.

I was though aware that the tile cache hold data redundant. I would have solved this though by some intermediate cache/map in the context of the map service.

I'm willing to rethink my old assumptions. Especially if they are based on findings on actual devices - which are a better fact base than my desktop assumptions.

Would it make sense to make some measurements and let some of the other users try the patch out before applying it?

Karry:
> I had the feeling that - at least from the performance view and on my desktop system - the old cache did not help improve performance. 

I don't know what is more narrower botleneck - if disk access or data parsing. It is possible that there was not any visible improvement on PC, but it is measurable on mobile. I can test it on Raspberry PI 1 (armv6), but I am not sure if I will have some time until weekend.

> I did not check for memory consumption reduction at that time. 

...and memory may be crucial on some devices like Raspberry ;-)

> I also assumed that handling multiple caches at different data flow location would make it difficult to control memory consumption.

It seems to me that duplicate objects (vector of nodes, feature buffers) are more memory consuming than reference caches.

> Would it make sense to make some measurements and let some of the other users try the patch out before applying it?

No problem. We don't have to hurry. I will try to do some measurements on my PC and Raspberry. Maybe another users will try it...

Karry:
Hi Tim, I have results from Raspberry PI (model 1 B; ARMv6; 512 MiB [371 for system]), it looks very promising and it confirms measurements from Jolla 1 phone (ARMv7; 1GiB RAM [800 for system]). When performance test render just one tile (OSM tile, it don't report count of used cache tiles), loading time is ~5% worse, but when bigger area is rendered (zoom level > 15), load time is almost half and use less than 50% of memory!

I am using this branch in my app on phone and improvement is really visible.

Here are logs:
 - Raspberry
    - master branch: [log 1](http://osmscout.karry.cz/perf/perf-armv6-master.1.log), [log 2](http://osmscout.karry.cz/perf/perf-armv6-master.2.log), [log 3](http://osmscout.karry.cz/perf/perf-armv6-master.3.log)
    - data cache branch: [log 1](http://osmscout.karry.cz/perf/perf-armv6-data-cache.1.log), [log 2](http://osmscout.karry.cz/perf/perf-armv6-data-cache.2.log), [log 3](http://osmscout.karry.cz/perf/perf-armv6-data-cache.3.log), 
 - Jolla 1
   - master branch: [log 1](http://osmscout.karry.cz/perf/perf-armv7-master-standard.1.log), [log 2](http://osmscout.karry.cz/perf/perf-armv7-master-standard.2.log)
   - data cache branch: [log 1](http://osmscout.karry.cz/perf/perf-armv7-data-cache-standard.2.log)
   - master branch, boundaries stylesheet: [log 1](http://osmscout.karry.cz/perf/perf-armv7-master-boundaries.1.log), [log 2](http://osmscout.karry.cz/perf/perf-armv7-master-boundaries.2.log), 
   - data cache branch, boundaries stylesheet: [log 1](http://osmscout.karry.cz/perf/perf-armv7-data-cache-boundaries.1.log), [log 2](http://osmscout.karry.cz/perf/perf-armv7-data-cache-boundaries.2.log)

libosmscout/include/osmscout/DataFile.h
```diff
@@ -144,16 +150,14 @@ namespace osmscout {
...
-   * Method is thread-safe.
+   * Method is NOT thread-safe.
    */
   template <class N>
   bool DataFile<N>::ReadData(const TypeConfig& typeConfig,
                              FileScanner& scanner,
                              FileOffset offset,
                              N& data) const
   {
-    std::lock_guard<std::mutex> lock(accessMutex);
```
rinigus: Lukas and Tim: it looks like this PR has made some methods non-thread safe. Can you tell what are the consequences of that?

Karry: Hi. I added lock to all public methods ( https://github.com/Framstag/libosmscout/pull/232/files#diff-76192701748b5e82dcf359e948e0e36cR435 ) and remove it from private ones. `DataFile` class is still thread-safe.

rinigus: Great idea, thank you!

aquiles2k:
Cant we add some tests to CI to get some basic memory measurements with every build?

Framstag:
In principle yes, though the resources of the build machines are rather limited. What concrete test do you have in mind?

aquiles2k:
Something to record memory usage at some predefined zoomlevel or after a certain number of map movements. 

The best would be if we could start the OSMScout2 app let it zoom to 3 default locations and make a measurement of the memory. This way we would have some compareable number at each build and would see if the numbers improve or get worse.

This is more of a integration/smoke test thing, not a unit test and i dont know if this is possible with the build machine though.

Just recording memory when running something like DrawMapQt might be interesting too in a first step but in my opinion wouldnt reflect all complexity of the client-qt-lib with all the caching and other stuff happening in the background.

Issue #202, opened 2017-01-10
Evaluate is_in tag in administrative_boundaries for building region hierachy


Framstag:
We currently use strict geometric "is in" evaluation for building the region tree. A child region belongs to a parent region simply because it is (mainly) contained by the parent. This is OK for simple regions with one area but does sometimes not work for more complex regions that consist of multiple areas which are also not close. See #191 for an concrete example.

To improve the region hierarchy we should also evaluate the ´is_in` attribute on administrative_boundaries and similar regions. For an example see https://www.openstreetmap.org/node/25293125. This example also shows that the content of this tag possibly has to get preprocessed and further validated for optimal use. In this case the complete region path is contained.

Framstag:
Information is now available in Region class. GenLocationIndex currently does check the is_in tag value against real name of parent region and throws a warning of they do not match.

Findings:
* is_in is not much used in my test area
* In some cases the parent name is not correctly spelled or has some pre/postfix that results in a mismatch. These are IMHO mapping errors. This is rather rare.
* is_in parent is not the direct parent as detected by evaluation geometric containment. I'm unsure if these are mapping errors or if this shows that building the tree based solely on containment is wrong. We have to further look at the actual data :-/ POssibly this is only solvable by also evaluating issue #203.

rinigus:
I will test it. So far, I have not seen any anomalies anymore after the commit related to https://github.com/Framstag/libosmscout/issues/191 . I had a question that I raised in that issue regarding rounding errors in the comparison. Should we treat all floating point comparisons as a source for errors and define some tolerance?

rinigus:
I do get quite a few warnings in Estonia. Actually, we are having a reform where several counties are joining and that would probably make is_in very error-prone in Estonia.

To check few errors further, it would be great to get OSM ID, in addition to the libosmscout ID. Maybe we should at least keep OSM ID for testing purposes somewhere? Or is it there already?

Framstag:
The code (besides the refactoring) doe snot change the generation of the index (yet). I added just additional checks against is_in spitting out warnings.

Regarding comparison of coordinates: I think there is no clear questions, it depends on what you want to archive. In cases where you want to compare coordinates for identity (have the same node as origin) you should compare serial and coordinate for exactness. In other cases a "softer" comparison might make more sense (containment,...). Note that libosmscout already strips some precision during import!

[...]

Issue #234, opened 2017-02-01
Performance degration after recent big area/way support


Framstag:
I see performance degrations after the recent change to support ways and areas with size >19 bit.

There are two likely reasons for this:
* The new read/write code is (much) slower then the older one
* We now load bigger boundaries and thus steps that make use of "area in area" or "point in area" comparisons are now slower since they have to compare against very big boundaries

The changes to the geofabrik exports - now making sure that relations are "complete" - may also result in more boundary relations required to compare against.

1.Verify what the real cause is (I assume option 2)
2. Find a way to compensate the performance loss

Karry:
It is interesting. I don't see any degradation with Czech. What country are you trying, Germany?

Karry:
Ah, I see. I am running France import right now. Step 21 is running 6 hours already. Here is few lines from `perf top`:
```
  42,23%  libosmscout_importd.so  [.] _ZNK8osmscout8GeoCoord6GetLatEv
  36,40%  libosmscout_importd.so  [.] _ZN8osmscout24GetRelationOfPointToAreaINS_8GeoCoordES1_EEiRKT_RKSt6vectorIT0_SaIS6_EE
  16,66%  libosmscout_importd.so  [.] _ZNKSt6vectorIN8osmscout8GeoCoordESaIS1_EEixEm
   4,59%  libosmscout_importd.so  [.] _ZNKSt6vectorIN8osmscout8GeoCoordESaIS1_EE4sizeEv
   0,03%  libosmscout_importd.so  [.] _ZN8osmscout21IsAreaSubOfAreaOrSameINS_8GeoCoordES1_EEbRKSt6vectorIT_SaIS3_EERKS2_IT0_SaIS8_EE
```

Karry:
I think that second reason is correct. When I connect gdb to this process, it spends majority of its time in `LocationIndexGenerator::AddRegion`. I will wait for @rinigus , it seems that he already has some idea how to optimise it :-) I just try fast area box checking: https://github.com/karry-space-with-my-second-forks/libosmscout/commit/17e5a979eaa66124b25eb69c7903c258f4367419 I will see how much it will helps.

rinigus:
Yes, please wait a little bit with optimizing the import. I have a feeling that the approach that I used gives excellent results (https://github.com/Framstag/libosmscout/issues/202) and, if needed, would require different ways to optimize the performance. 1-2 days and we can test it all on PR :)

rinigus:
[...]

PR #237, opened 2017-02-03
Area in area determined using raster

rinigus:
Hi!

here is a PR discussed at https://github.com/Framstag/libosmscout/issues/202 . At this stage, its mainly for reading and testing. I have few ideas on how to improve performance and plan to push it into this branch. Let me know how it works for you.

In Estonia, the only 2 regions that I have seen to be wrong are related to "islands" within other areas. Looking forward to feedback.

### DIFF
--- a/libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
+++ b/libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
@@ -103,18 +103,18 @@ namespace osmscout {
       double area_region = region.boundingBox.GetSize(); // reference area
 
       GeoCoord cornerMin( std::max( region.boundingBox.GetMinLat(), this->boundingBox.GetMinLat() ),
-			  std::max( region.boundingBox.GetMinLon(), this->boundingBox.GetMinLon() ) );
-      
+                          std::max( region.boundingBox.GetMinLon(), this->boundingBox.GetMinLon() ) );
+
       GeoCoord cornerMax( std::min( region.boundingBox.GetMaxLat(), this->boundingBox.GetMaxLat() ),
-			  std::min( region.boundingBox.GetMaxLon(), this->boundingBox.GetMaxLon() ) );
+                          std::min( region.boundingBox.GetMaxLon(), this->boundingBox.GetMaxLon() ) );
 
       GeoBox bx(cornerMin, cornerMax);
-      
+
       // 95% of the bounding box has to be covered
       if ( area_region * 0.95 > bx.GetSize() )
-	return false;
+        return false;
     }
-      
+
     for (const auto& bb : boundingBoxes) {
       for (const auto& rbb : region.boundingBoxes) {
         if (bb.Intersects(rbb)) {
@@ -131,51 +131,51 @@ namespace osmscout {
     probePoints.clear();
     if ( boundingBoxes.size() != areas.size() )
       CalculateMinMax();
-    
+
     for (size_t i=0; i < areas.size(); ++i)
       CalculateProbePointsForArea(i, 0);
   }
-  
+
   void LocationIndexGenerator::Region::CalculateProbePointsForArea(size_t areaIndex, size_t refinement)
   {
     const size_t targetprobes = 30; // target number of points
-				    // representing an area in region
+                                    // representing an area in region
 
     const size_t minprobes = 3;     // minimal number of required
-				    // points. If there are less
-				    // points than this minimum, the
-				    // probe points are supplemented
-				    // with the area border points
+                                    // points. If there are less
+                                    // points than this minimum, the
+                                    // probe points are supplemented
+                                    // with the area border points
 
     const std::vector<GeoCoord> &area = areas[areaIndex];
     const GeoBox &box = boundingBoxes[areaIndex];
 
     const double split_base = 10;
     double nsplit = split_base * (1 << refinement);
-    
+
     double delta_lat = (box.GetMaxCoord().GetLat() - box.GetMinCoord().GetLat()) / nsplit;
     double delta_lon = (box.GetMaxCoord().GetLon() - box.GetMinCoord().GetLon()) / nsplit;
 
     // checking if the selected step is already too small
     const GeoCoord bbox_max_lat(box.GetMaxCoord().GetLat(), box.GetMinCoord().GetLon());
     const GeoCoord bbox_max_lon(box.GetMinCoord().GetLat(), box.GetMaxCoord().GetLon());
-	
+
     const double distance_lat =
       bbox_max_lon.GetDistance( box.GetMaxCoord() ) /
       (box.GetMaxCoord().GetLat() - box.GetMinCoord().GetLat()); // distance along latitude per degree
     const double distance_lon =
       bbox_max_lat.GetDistance( box.GetMaxCoord() ) /
       (box.GetMaxCoord().GetLon() - box.GetMinCoord().GetLon()); // distance along longitude per degree
-	
+
     // 100 meters is taken as a smallest step
-    const double min_delta_lat = 0.1 / distance_lat; 
+    const double min_delta_lat = 0.1 / distance_lat;
     const double min_delta_lon = 0.1 / distance_lon;
-    
+
     if ( refinement > 0 ) {
       if (delta_lat < min_delta_lat && delta_lon < min_delta_lon )
-	return; // the refinement is considered to be too fine
-    
-      delta_lat = std::max(min_delta_lat, delta_lat); 
+        return; // the refinement is considered to be too fine
+
+      delta_lat = std::max(min_delta_lat, delta_lat);
       delta_lon = std::max(min_delta_lon, delta_lon);
     }
     else {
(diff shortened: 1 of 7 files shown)


Karry:
Thank you. I will take a look to your changes later, on pc. How do you verify correctness?

rinigus:
To check, I look through the IsIn warnings printed during import step 21. For example,

```
WW Area 26133329 - 'Tömbi küla' parent should be 'Paistu vald' but is 'Viljandi vald'
```

Then I go and see what OSM thinks about this place: http://www.openstreetmap.org/search?query=T%C3%B6mbi%20k%C3%BCla#map=14/58.2585/25.6769&layers=H . In this case, it agrees with our import: Village Boundary Tömbi küla, Viljandi vald, Viljandi maakond, Estonia. 

Almost all IsIn warnings in Estonia are caused by changes in counties. Now, exception was for two city parts:

```
   WW Area 10711140 - 'Kukruse linnaosa' parent should be 'Kohtla-Järve linn' but is 'Kohtla vald'
   WW Area 11225218 - 'Oru linnaosa' parent should be 'Kohtla-Järve linn' but is 'Toila vald'
```

This was due to the fact that these city parts are enclaves within Kohtla vald (see http://www.openstreetmap.org/relation/355233#map=13/59.3924/27.3501&layers=H and compare to http://www.openstreetmap.org/relation/351623#map=12/59.3656/27.4020&layers=H). Unfortunately, libosmscout does not support holes within areas. So, its up to an order in OSM whether the city part
in that city was given to the city 'Kohtla-Järve linn' or the county Kohtla vald'.

rinigus:
The latest commit (https://github.com/Framstag/libosmscout/pull/237/commits/9fff7b0f9b73e95ea98c5564b8c104df08acd93c) should address the performance issues (possible degradation of performance induced by this approach and maybe even address https://github.com/Framstag/libosmscout/issues/234). For Estonia, import step 21 went from 12 s to 6.5 s after the latest commit.

Please, let's test this import on few countries. I will continue testing tomorrow on some other countries to check whether this approach breaks something badly. Looking forward for feedback :)

rinigus:
Had a chance to test a bit further by checking all printed out is_in warnings. So, with the respect of is_in check:

denmark - imported perfectly;
finland - imported perfectly;
sweden - imported perfectly;

so far, looking good. I'll test on Germany as well - but it may take some time :)

rinigus:
And continuation of testing:

Czech Republic: perfect import (let's see what you think about it, @Karry)
Latvia: perfect import

rinigus:
With Germany, I did not look through all warnings. I did try to look through the ones that are probably 30-60% of all of them and the following were the only ones that were strange:

```
'Heißen' parent should be 'Mülheim an der Ruhr' but is 'Heißen'
```

So, according to my testing it all works quite nicely. 

At this stage, after the testing, I am asking to consider this PR for merging.

libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
```diff
@@ -107,6 +126,131 @@ namespace osmscout {
...
+  }
+
+  bool LocationIndexGenerator::Region::Contains(Region& child) const
+  {
+    // test whether admin levels allow this to contain child
+    if (this->level >= 0 /* check if we have admin level defined for this */ &&
+	( child.level >= 0 && this->level >= child.level ) )
+      return false;
+    
+    if (child.probePoints.size() == 0) // calculate probe points if they are missing
```
Karry: I suggest `empty()`

libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
```diff
@@ -94,8 +95,26 @@ namespace osmscout {
     return false;
   }
 
-  bool LocationIndexGenerator::Region::CouldContain(const Region& region) const
+  bool LocationIndexGenerator::Region::CouldContain(const Region& region, bool strict) const
   {
+    if (strict) {
```
Karry: hmm, should be this condition inverse? `if (!strict)`

libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
rinigus: Agreed! I'll change and submit all changes when the review is over :)

libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
rinigus: Maybe name is not the best, but its not tolerant as well :). When strict is true, we check first whether the overall bounding box can contain the proposed child region. If this check fails we don't even look for intersection and reject the proposed child. As such, its an additional check which makes it harder to satisfy the condition tested by the method.

Karry: Ah, I see it now - if it is true, it is more strict, so the name makes sence...

libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
```diff
@@ -466,35 +610,59 @@ namespace osmscout {
...
-                AddRegion(*childRegion,region);
-              }
-
-              return;
-            }
-          }
-        }
+      if ( childRegion->CouldContain(*region, true) ) {
+	added = AddRegion(*childRegion,region,false);
+	if (added) return true;
```
rinigus: `assume_contains` is just to allow insertion into the `<root>`. Unfortunately,  `parent.Contains(*region)` for root returns `false` and without this assume_contains I couldn't make it work. For all subregions of `<root>`, there is no need for it.

rinigus:
The speed increase is due to the immediate tests of child regions as deep as possible in the tree. So, we don't have to test large areas for containment.

libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
```diff
@@ -94,8 +95,26 @@ namespace osmscout {
...
+      // is taken into account
+      double area_region = region.boundingBox.GetSize(); // reference area
+
+      GeoCoord cornerMin( std::max( region.boundingBox.GetMinLat(), this->boundingBox.GetMinLat() ),
+			  std::max( region.boundingBox.GetMinLon(), this->boundingBox.GetMinLon() ) );
+      
+      GeoCoord cornerMax( std::min( region.boundingBox.GetMaxLat(), this->boundingBox.GetMaxLat() ),
+			  std::min( region.boundingBox.GetMaxLon(), this->boundingBox.GetMaxLon() ) );
+
+      GeoBox bx(cornerMin, cornerMax);
```
Karry: I suggest to create new method `Intersection` to `GeoBox` for this. I was surprised that it don't exists yet :-)

```
  /**
   * Create new GeoBox from intersection of this with other
   * If not Intersects, invalid GeoBox is returned
   * @param other
   */
  GeoBox GeoBox::Intersection(const GeoBox& other) const {
    assert(valid);
    assert(other.valid);

    if (!Intersects(other)){
      return GeoBox();
    }

    GeoCoord cornerMin( std::max( other.GetMinLat(), GetMinLat()),
                        std::max( other.GetMinLon(), GetMinLon()));

    GeoCoord cornerMax( std::min( other.GetMaxLat(), GetMaxLat()),
                        std::min( other.GetMaxLon(), GetMaxLon()));

    return GeoBox(cornerMin, cornerMax);
  }

```

libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
Karry: Ok, but shouldn't be the addition to `<root>` be solved in `SortInBoundaries` ?

rinigus: Yes, but for that `SortInBoundaries` calls `AddRegion` with the default arguments (i.e. `assume_contains=true` in this case)

libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
rinigus: Want to push it into the PR?

rinigus: Actually, since I am correcting formatting (tabs -> spaces), maybe its easier if I push it in. Otherwise we'll have lots of conflicts...

libosmscout/include/osmscout/util/GeoBox.h
```diff
@@ -110,6 +110,25 @@ namespace osmscout {
     }
 
     /**
+     * Create new GeoBox from intersection of this with other
+     * If not Intersects, invalid GeoBox is returned
+     * @param other
+     */
+    GeoBox Intersection(const GeoBox& other) const
```
Karry: It should be inline or in `GeoBox.cpp`. It seems that import for `std::max` missing here...

rinigus: Usually, all what's defined in the class headers is inline. But, since you already define all with "inline", I'll add that as well. And the max-include, sorry for missing it.

Framstag:
If it builds, it is OK for me. It seems, though. That at least one build still fails:

https://ci.appveyor.com/project/Framstag/libosmscout/build/576/job/lk4e1xjn65qytjem

It looks similar to https://stackoverflow.com/questions/5004858/stdmin-gives-error, but I would be surprised if this would be the actual cause. There should be no windows.h include in headers (and if so, we should try to get rid of it)! Strange, too. That only the mingw+cmake build has the problem!?

Regarding formatting (not that much of a problem here):
* no tabs, right
* I use sometimes more and sometimes less spacing
* I like block alignments for class attributes and variables
* The order of includes follow a pattern: if cpp file, initial include of corresponding header, then system headers, than library includes (ordered by library, grouped)

I tolerate small derivations (it is no standard formatting), but please at least try to be close to it

Framstag:
And of course: Thanks for working on this! Looks like this is definitely a nice improvement!

rinigus:
I'll look into this issue with compilation failure. Bit odd, but maybe your link to similar issue would help it. I'll probably move the function from inline version to cpp first and let's see if that would help (there the headers are a bit more under control)

PR #258, opened 2017-02-20
add public api for setting renderer font

aquiles2k:
So one can easily set the desired font and font size through some settings dialog


Framstag:
I have now two patches regarding font setting. Can you please check if they are additive or if I should only take one of them (same messages goes to the other patch ;-))?

PR #259, opened 2017-02-20
make map font configurable from QML

Karry:
Map with Comic Sans fonts, configured simply from QML:

```
    Settings {
        fontName: "Comic Sans MS"
    }
```

comic-sans-map

Why not :-)

### DIFF
--- a/OSMScout2/CMakeLists.txt
+++ b/OSMScout2/CMakeLists.txt
@@ -33,10 +33,12 @@ set(OSMSCOUT_USER_AGENT "OSMScout demo app %1")
 
 set(HEADER_FILES
     src/Theme.h
+    src/AppSettings.h
 )
 
 set(SOURCE_FILES
     src/Theme.cpp
+    src/AppSettings.cpp
     src/OSMScout.cpp
 
     # qml files in CMake sources make it visible in QtCreator
--- a/OSMScout2/OSMScout.pro
+++ b/OSMScout2/OSMScout.pro
@@ -29,9 +29,11 @@ RCC_DIR = $$DESTDIR/
 UI_DIR = $$DESTDIR/
 
 SOURCES = src/OSMScout.cpp \
-          src/Theme.cpp
+          src/Theme.cpp \
+          src/AppSettings.cpp
 
-HEADERS = src/Theme.h
+HEADERS = src/Theme.h \
+          src/AppSettings.h
 
 DISTFILES += \
     qml/custom/MapButton.qml \
--- a/OSMScout2/qml/main.qml
+++ b/OSMScout2/qml/main.qml
@@ -100,6 +100,9 @@ Window {
         id: settings
         //mapDPI: 50
     }
+    AppSettings{
+        id: appSettings
+    }
 
     GridLayout {
         id: content
@@ -120,15 +123,15 @@ Window {
 
             function setupInitialPosition(){
                 if (map.databaseLoaded){
-                  if (map.isInDatabaseBoundingBox(settings.mapView.lat, settings.mapView.lon)){
-                    map.view = settings.mapView;
-                    console.log("restore last position: " + settings.mapView.lat + " " + settings.mapView.lon);
+                  if (map.isInDatabaseBoundingBox(appSettings.mapView.lat, appSettings.mapView.lon)){
+                    map.view = appSettings.mapView;
+                    console.log("restore last position: " + appSettings.mapView.lat + " " + appSettings.mapView.lon);
                   }else{
-                    console.log("position " + settings.mapView.lat + " " + settings.mapView.lon + " is outside database, recenter");
+                    console.log("position " + appSettings.mapView.lat + " " + appSettings.mapView.lon + " is outside database, recenter");
                     map.recenter();
                   }
                 }else{
-                  map.view = settings.mapView;
+                  map.view = appSettings.mapView;
                 }
             }
 
@@ -142,7 +145,7 @@ Window {
             }
             onViewChanged: {
                 //console.log("map center "+ map.view.lat + " " + map.view.lon + "");
-                settings.mapView = map.view;
+                appSettings.mapView = map.view;
             }
             Component.onCompleted: {
                 setupInitialPosition();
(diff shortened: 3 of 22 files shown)


Framstag:
I have now two patches regarding font setting. Can you please check if they are additive or if I should only take one of them (same messages goes to the other patch ;-))?

Karry:
It is funny :) I thing that my solution is more complex: 

 - work for both rendering methods (I want to use it with tiled renderering)
 - persist setting
 - expose functionality to QML

@aquiles2k what do you think?

Framstag:
Complex has a negative message for me ;-) It is only allowed to be more complex if it is also more flexible and simply "better". Which solutions suites more the needs od the developer and end user?

aquiles2k:
Well, yeah looks like we had the same idea :)

Yours of course is more advanced regarding qml usage. The thing is i dont need all the qml settings stuff anyway, since we have our own qsettings structure in the main desktop application. So i just needed a simple way to tell the map component which font it should use.

I'm fine though with it being done your way since i need to adjust this anyway and work around it. 

But perhaps you could generally reconsider putting all settings stuff in the "qt client lib". In my opinion this belongs in a "qt client app", e.g. OSMScout2 where the developer decides how he wants or has to save user settings. In my case we for example need to save the settings in an .ini file beside the executable.

Karry:
I think that is useful when library remember its settings (font, overrided dpi, online tile provider...) without additional configuration, so integration with `QSettings` make sense to me... Question is what object should be called for configuration. I don't like huge DBThread singleton that contains almost everything. I would like to separate it to smaller classes, but I don't have idea how to do it (yet) and keep api simple... 

If you use `QSettings` to store your app specific configuration, library settings should be prepended by some path (`QSettings::beginGroup`) to avoid possible conflicts. if you want to store configuration to ini file to non-standard location, you can setup it globally by `QSettings::setPath`... 

I would like to find solution for both of us... What about to remove things from Settings that are not related with rendering (gpsFormat, mapView), transform Settings from singleton to shared-pointer inside DBThread and make it accessible from outer world to change rendering related configuration? 

Btw, another place where QSetting is used inside qt-client-library is `MapManager` that use it to store http cookies persistently...

aquiles2k:
Yes, i know :) I just wanted to point out that most likely normal desktop apps have there own usage of QSettings internally already and that suddenly, when integrating the libosm lib another Settingsentry is created they have no control about.

But anyway, yes your way with the shared pointer would be great. In fact thats the way i currently have my local modified client qt lib running with.

I just added an parameter to the singleton call to have the option to "inject" my own settings object.

```
Settings* Settings::GetInstance(QSettings *externalSettings)
{
    if (settingsInstance == NULL){
        settingsInstance = new Settings(externalSettings);
    }
    return settingsInstance;
}
```
```
Settings::Settings(QSettings *externalSettings):
    useExternalSetting(false),
    view(NULL)
{
    QScreen *srn=QGuiApplication::screens().at(0);
    physicalDpi = (double)srn->physicalDotsPerInch();
    
    // if you want to inject an external settings object - make sure to call Settings::GetInstance early
    if (externalSettings) {
        qDebug() << "Settings CTOR: using external settings object";
        settings = externalSettings;
        useExternalSetting = true;
    } else {
        qDebug() << "Settings CTOR: using internal settings object";
        settings = new QSettings();
    }
}
```
This way libosmscout can do whatever it wants with the setting and i can be sure that it uses my predefined QSettings object i create early on in my app with my necessary parameters.

Karry:
If I understand QSetting documentation correctly, it is not necessary. You should just define target and format of application settings storage by one static method ([`setPath`](http://doc.qt.io/qt-5/qsettings.html#setPath)), and then you can use multiple QSettings instances created with default constructor...

Anyway, I will prepare another MR tomorrow...

Karry:
@aquiles2k , @Framstag  please review my last commits. I moved `Settings` (`std::shared_ptr`) to `DBThread` instance, remove app specific properties (gpsFormat, mapView) and move all settings keys to `OSMScoutLib` group...

aquiles2k:
Wow thats some awesome work again :) Looks good to me. The only thing im wondering is your license header beeing all GPL. I was under the impression libosmscout is LGPL :)

Sidenote: I also found out why previously the settings between my app and the lib werent in sync. When using the empty QSettings ctor (like in libosm) it stores it under the structure "apporganization/appName.ini". But im using the QSettings(const QString &fileName, Format format) ctor with a custom filename and path (settings.ini besides the application executable). With QSetting you can only have control about defaultPath (i can set that to the application directory) and the default format (i can set that to ini format) but not how the ini file is named or the directory structure i mentioned above.

That all said it would be great to have a parameter in libosm Settings ctor to set this filename (could be default empty using the normal ctor). But i can understand if that is a too special case in which case i'll just  do it locally.

Karry:
To GPL/LGPL, what file do you mean? I just added `AppSettings` that is in OSMScout2 demo, it is under GPL... But we may "relicense" OSMScout2 and StyleEditor to LGPL too... It may cause problems in future - if we would want to move some classes from these examples to library, we should ask all authors for relicensing...

I will take a look at changing the file name, I just read documentation and `setPath` seemed adequate, but I never try it...

Framstag:
The LGPL ist for libraries. All changes to the libosmscout-client-qt library should thus be brought under LGPL. The demos and all other applications are under GPL (at least I did it this way when I started the project, for smaller demos we could even have a more relaxed licence). For OSMScout/StyleEditor IMHO we should keep GPL to enforce the "give back" mechanism.

Putting libraries under GPL and applications under LGPL does not make much sense IMHO license-wise (this is just not what the license were made for).

So, depending where you did the change and add new sources, choose the right license for it please.

aquiles2k:
Ah sorry your right, that AppSetting header is inside OSMScout et al. and thus under GPL.

Framstag:
Ah, sorry, too. I answered shortly form work and hat not the time to check the actual case.

So it this the pull request you both want me to get merged?

Karry:
Uff, QSettings really don't allow store all values to custom file when you are using default constructor :-( Only way how to do it is setup custom format as default and control writes and reads in your own functions... It is something that noone wants to do. So I added possibility to setup external QSettings storage to Settings, it is nicest way how to do it. @aquiles2k it is fine for you?

```
SettingsRef settings=std::make_shared<Settings>(new QSettings("test.ini", QSettings::IniFormat));
```

@Framstag I am fine with it, just wait for Adrian response ;-)

aquiles2k:
Beautiful :) Please merge!

PR #260, opened 2017-02-20
disable polygon otimisation for Qt, it is not needed

Karry:
without optimisation, map is nicer and it is not needed for Qt, it don't have any performance inpact


Framstag:
Yes, depending on the mobile device, its CPU speed and is graphics speed, it is possible that setting the option to none is faster. Which device are you using? If it is in the category low/mid end your statement holds more truth :-)

(Waiting for the central build to finish before merging)

Karry:
I have Jolla 1, it is slow arm system with relatively fast gpu... If someone will have problems, it is not problem to expose some configuration method for this or add some autodetection...

PR #112, opened 2016-09-09
new PostalCode feature

aquiles2k:
*added basic postal code to nodes in the location index through a new PostalCode feature for map.ost
*extended location lookup demo to display the postal code in results

PR #262, opened 2017-02-21
add area postal code import

aquiles2k:
this has been on my list for a while. with this areas like buildings also get there postal code added to the address for a better location description.

### DIFF
--- a/libosmscout-import/include/osmscout/import/GenLocationIndex.h
+++ b/libosmscout-import/include/osmscout/import/GenLocationIndex.h
@@ -286,6 +286,7 @@ namespace osmscout {
                                 const FileOffset& fileOffset,
                                 const std::string& location,
                                 const std::string& address,
+                                const std::string& postalCode,
                                 const std::vector<Point>& nodes,
                                 const GeoBox& boundingBox,
                                 bool& added);
--- a/libosmscout-import/src/osmscout/import/GenAreaAreaIndex.cpp
+++ b/libosmscout-import/src/osmscout/import/GenAreaAreaIndex.cpp
@@ -22,7 +22,7 @@
 #include <vector>
 
 #include <osmscout/TypeFeatures.h>
-
+ 
 #include <osmscout/AreaAreaIndex.h>
 #include <osmscout/AreaDataFile.h>
 
@@ -44,11 +44,12 @@ namespace osmscout {
   class AreaLocationProcessorFilter : public SortDataGenerator<Area>::ProcessingFilter
   {
   private:
-    FileWriter                 writer;
-    uint32_t                   overallDataCount;
-    NameFeatureValueReader     *nameReader;
-    LocationFeatureValueReader *locationReader;
-    AddressFeatureValueReader  *addressReader;
+    FileWriter                   writer;
+    uint32_t                     overallDataCount;
+    NameFeatureValueReader       *nameReader;
+    LocationFeatureValueReader   *locationReader;
+    AddressFeatureValueReader    *addressReader;
+    PostalCodeFeatureValueReader *postalCodeReader;
 
   public:
     bool BeforeProcessingStart(const ImportParameter& parameter,
@@ -71,7 +72,8 @@ namespace osmscout {
 
     nameReader=new NameFeatureValueReader(typeConfig);
     locationReader=new LocationFeatureValueReader(typeConfig);
-    addressReader=new AddressFeatureValueReader(typeConfig);
+    addressReader=new AddressFeatureValueReader(typeConfig);    
+    postalCodeReader=new PostalCodeFeatureValueReader(typeConfig);
 
     try {
       writer.Open(AppendFileToDir(parameter.GetDestinationDirectory(),
@@ -99,10 +101,12 @@ namespace osmscout {
         NameFeatureValue     *nameValue=nameReader->GetValue(ring->GetFeatureValueBuffer());
         LocationFeatureValue *locationValue=locationReader->GetValue(ring->GetFeatureValueBuffer());
         AddressFeatureValue  *addressValue=addressReader->GetValue(ring->GetFeatureValueBuffer());
+        PostalCodeFeatureValue *postalCodeValue=postalCodeReader->GetValue(ring->GetFeatureValueBuffer());
 
         std::string          name;
         std::string          location;
         std::string          address;
+        std::string          postalCode;
 
         if (nameValue!=NULL) {
           name=nameValue->GetName();
@@ -114,6 +118,10 @@ namespace osmscout {
 
         if (addressValue!=NULL) {
           address=addressValue->GetAddress();
+        }
+        
+        if (postalCodeValue!=NULL) {
+          postalCode=postalCodeValue->GetPostalCode();
         }
 
         bool isAddress=!ring->GetType()->GetIgnore() &&
@@ -143,6 +151,7 @@ namespace osmscout {
               writer.WriteFileOffset(offset);
               writer.WriteNumber(ring->GetType()->GetAreaId());
               writer.Write(name);
+              writer.Write(postalCode);
               writer.Write(location);
               writer.Write(address);
               writer.Write(r->nodes,false);
@@ -155,6 +164,7 @@ namespace osmscout {
           writer.WriteFileOffset(offset);
           writer.WriteNumber(ring->GetType()->GetAreaId());
           writer.Write(name);
+          writer.Write(postalCode);
           writer.Write(location);
           writer.Write(address);
           writer.Write(ring->nodes,false);
@@ -183,6 +193,9 @@ namespace osmscout {
 
     delete addressReader;
     addressReader=NULL;
+    
+    delete postalCodeReader;
+    postalCodeReader=NULL;
 
     try {
       writer.SetPos(0);
--- a/libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
+++ b/libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
@@ -1222,6 +1222,7 @@ namespace osmscout {
                                                       const FileOffset& fileOffset,
                                                       const std::string& location,
                                                       const std::string& address,
+                                                      const std::string &postalCode,
                                                       const std::vector<Point>& nodes,
                                                       const GeoBox& boundingBox,
                                                       bool& added)
@@ -1236,6 +1237,7 @@ namespace osmscout {
                                    fileOffset,
                                    location,
                                    address,
+                                   postalCode,
                                    nodes,
                                    boundingBox,
                                    added);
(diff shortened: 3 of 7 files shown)


aquiles2k:
Old Berlin import:
```
 - Index address areas...
   73471 areas analyzed, 68709 addresses founds, 4115 POIs founds
 - Index address ways...
   5 ways analyzed, 5 POIs founds
 - Index address nodes...
   334953 nodes analyzed, 316952 addresses founds, 19889 POIs founds, 321077 postal codes found

   + Summary...
   Mandatory files:
   File areaarea.idx: 321.1 KiB
   File areanode.idx: 1.3 MiB
   File areas.dat: 13.2 MiB
   File areaway.idx: 223.9 KiB
   File bounding.dat: 14 B
   File intersections.dat: 2.6 MiB
   File intersections.idx: 814.0 KiB
   File location.idx: 4.4 MiB
   File nodes.dat: 6.6 MiB
   File router.dat: 9.0 MiB
   File router.idx: 815.0 KiB
   File router2.dat: 555 B
   File types.dat: 27.6 KiB
   File water.idx: 10.5 KiB
   File ways.dat: 4.7 MiB
   => 44.0 MiB
   Optional files:
   File areasopt.dat: 71.6 KiB
   File waysopt.dat: 128.9 KiB
   => 200.5 KiB
   Import OK!
```
New Berlin import:
```
 - Index address areas...
   73471 areas analyzed, 68709 addresses found, 4115 POIs found, 70250 postal codes found
 - Index address ways...
   5 ways analyzed, 5 POIs found, 0 postal codes found
 - Index address nodes...
   334953 nodes analyzed, 316952 addresses founds, 19889 POIs founds, 321077 postal codes found

+ Summary...
   Mandatory files:
   File areaarea.idx: 321.3 KiB
   File areanode.idx: 1.3 MiB
   File areas.dat: 13.6 MiB
   File areaway.idx: 223.9 KiB
   File bounding.dat: 14 B
   File intersections.dat: 2.6 MiB
   File intersections.idx: 814.0 KiB
   File location.idx: 4.7 MiB
   File nodes.dat: 6.6 MiB
   File router.dat: 9.0 MiB
   File router.idx: 815.0 KiB
   File router2.dat: 555 B
   File types.dat: 27.6 KiB
   File water.idx: 10.5 KiB
   File ways.dat: 4.7 MiB
   => 44.7 MiB
   Optional files:
   File areasopt.dat: 71.6 KiB
   File waysopt.dat: 128.9 KiB
   => 200.5 KiB
   Import OK!
```

Framstag:
Hmmm, I'm unsure if this is the best solution. Using this approach we are adding a rather huge redundancy to the location index. The number of different postal codes is likely much smaller than the number of postal code found at all. This means you are storing the same string value again and again. For Berlin this seams to only increase the size of the database by ~0.7MB. But what about bigger exports? Isn't there a way to reduce such redundancy? I spend a lot of time to avoid such pitfalls.

aquiles2k:
old germany:
```
+ Summary...
   Mandatory files:
   File areaarea.idx: 40.3 MiB
   File areanode.idx: 25.3 MiB
   File areas.dat: 1.0 GiB
   File areaway.idx: 19.6 MiB
   File bounding.dat: 14 B
   File intersections.dat: 192.7 MiB
   File intersections.idx: 56.9 MiB
   File location.idx: 125.3 MiB
   File nodes.dat: 90.5 MiB
   File router.dat: 614.1 MiB
   File router.idx: 57.0 MiB
   File router2.dat: 1.9 KiB
   File types.dat: 27.6 KiB
   File water.idx: 13.5 MiB
   File ways.dat: 445.2 MiB
   => 2.6 GiB
   Optional files:
   File areasopt.dat: 25.6 MiB
   File waysopt.dat: 25.2 MiB
   => 50.8 MiB
   Import OK!
```
new germany:
```
+ Summary...
   Mandatory files:
   File areaarea.idx: 40.4 MiB
   File areanode.idx: 25.3 MiB
   File areas.dat: 1.0 GiB
   File areaway.idx: 19.6 MiB
   File bounding.dat: 14 B
   File intersections.dat: 192.7 MiB
   File intersections.idx: 57.0 MiB
   File location.idx: 162.8 MiB
   File nodes.dat: 90.5 MiB
   File router.dat: 614.4 MiB
   File router.idx: 57.0 MiB
   File router2.dat: 1.9 KiB
   File types.dat: 27.6 KiB
   File water.idx: 13.5 MiB
   File ways.dat: 445.2 MiB
   => 2.7 GiB
   Optional files:
   File areasopt.dat: 22.4 MiB
   File waysopt.dat: 25.2 MiB
   => 47.6 MiB
   Import OK!
```

aquiles2k:
Sure, one can possibly bypass saving the ZIP string everywhere with the postal code boundary relation and some logic like we were talking about some month ago. 

But to be honest i have so much on my plate right now and this was the easiest and quickest solution to get the problem solved and improve the geocoding resolution. 

I dont mind the 100MB extra on the map but i can understand if you dont want to integrate it.

Karry:
Hi. Do you have statistics how many percent of buildings has different `addr:postcode` tag than postal area that belongs to? I quess that this value can be stored only in case when these two values are different...

aquiles2k:
Hm no sorry i dont have such statistics. From what i read i know that there has been a lot of effort in the last years though to consolidate the postal code data in osm, at least for germany. There will of course always be some corner cases.

Framstag:
See mail to the mailing list to gather further opinions. If there is no veto I suggest to accept the patch and make a follow up issue for further improvement.

aquiles2k:
Fine with me!

Karry:
Adrian, I don't understand the motivation, please correct me if I am wrong. This patch reads postal code feature from all addressable objects if it is available, and write it to temporary `(way|area|node)address.dat` that is merged to  final `location.idx` then. Right? 

What is benefit of that? I probably don't see it. It helps to something that can't be solved now? I understand `location.idx` as place where is tree of administrative boundaries, streets/places and addressable objects. Other things like names, postal codes can be read from object features, it is not necessary to duplicate it to location index. Or not?

@Framstag it would be nice to have something like "DBMS data schema" for osmscout database, it would be nice addition to [file description](http://libosmscout.sourceforge.net/documentation/database/). I still don't have good knowledge about data structures and its "relations" :-(

Framstag:
You are right. The postal code information would be already available at the object itself. So one could gain the same information by loading the object by file offset as returned form the location index.

So the question is, do we create an index over data that is already there. Only reason in this case would be performance. Filtering is faster if I do not need to load the object but have all information in the index.

Question is: What is the use case and is the method of search and access to the index really performance relevant?

Regarding "DBMS data schema": In which format would you expect such information? Do you need a logical model or a physical model? Do you have an example for one of the index you know better?

aquiles2k:
The use case was simply to get the ZIP code from addressable objects when geocoding since this is something u always need to provide in germany when talking about adresses. 

This, if you remember our lengthy mailing list communications, wasnt possible back then. I would only get a streetname and a housenumber from search results. So i made this PR months ago to change it (https://github.com/Framstag/libosmscout/pull/112) But i only made it for Nodes just as a proof that it works. So now i got the time to also do it for areas.

If of course this could have already be done by some other means with existing code im fine closing this PR. But then we should also remove the node ZIP stuff.

aquiles2k:
Ah, do you mean something like what you did with the admin_region level in the search results previously?

call place.GetObjectFeatures() and collect the PostalCodeFeature i added in my old PR? If that works im fine with it :)

But i would highly recommend to add some convenience functions in Place. For new users its really confusing to have place.GetLocation() returning the street name and place.GetAddress() returning a housenumber but no ZIP. 

Would be good to have something like place.GetZIP() and potentially even have a place.GetCity() I think this would be a lot less confusing.

aquiles2k:
Oh i just wanted to mention that theoretically it would make sense to be able to search for ZIP Codes too beside region, streetname and housenumber. So you could search for "12345 Streetname" instead of "regionname streetname". For that i think you would need it indexed? But i might be wrong.

Karry:
Yes, I mean PostalCode feature on addressable object. When you use LocationDescription demo, it should be displayed already... 

If you want to **search** by expression `ZIP, street` you definitely will need another kind of index. Question is if this still fits to osmscout scope. Try to look at @rinigus project https://github.com/rinigus/geocoder-nlp , it uses trained models for parsing address expressions in natural language and lookup it in normalised database... I like this approach, but required dataset size looks scary for me.

rinigus:
Probably postal index would not be that bad as in my case (and I am planning to see if I can reduce the sizes of geocoder-nlp). @Karry's comment is spot on - if you want to search, you would need an index. You would have to think how to relate your postcode index to other indexes, such as location area. Since, if you want to search for `ZIP, Street`, you would have to have a shortcut that links them. Hopefully its not too difficult, but I don't know how to do it (and haven't thought about it much either).

Framstag:
You do not need to index data to be able to search for it. You can always scan the whole data sequentially looking for matches. This is though slow in most cases.

For certain seaches you can also apply some search criteria to the index, get a significant reduced list of possible matches and then apply further not indexed criteria to the rest. This is how (relational) database work. So for ZIP search you would need further criteria to get applied first. If you for example search for address, street, city additionally search for zip would not require an index. If you search for address, street, zip you currently would have to scan all cities in the index for a potential match. This is because the current patch places zip data along the address. Only if it would attach zip information to location or better region things would be simpler and faster.

So, for zip lookup the current solution does not really qualify since it does not index the zip data it simply places the zip information within the index (which is also a technic which is part of (some) relational databases).

But if the zip information is already (if configured in the ma.ost) added to the nodes, ways, areas - putting it into the index, too, would only add redundancy and would  not (significantly) improve lookup or lookup speed. The current patch also does not seem to delete zip information at nodes, area and ways (it could though). So Karry is right. 

Adding convenience methods: I hesitate to add convenience methods. The whole engine should not rel and too much features to be always set. This way the engine is more flexible and can get adapted in more ways. Currently zip information is a clear "optional" and even alternatives could get added without problem. Convenience method should not break this (if not required).

aquiles2k:
Hi, i dont need zip searching atm. I was just mentioning it as something that would be kind of "expected" in a geocoder, which is something libosm provides. So everything fine from my side here :)

I see that the things i need can be acheived through other means already so i'm closing the PR. I really only was adding to the same technique i used month ago in the other PR so i thought it would be ok.

But i can at least say that i sparked another good discussion :D

Framstag:
I hope that this is OK for you, @aquiles2k. I hate it, when people work on patches that finally do not get into the repository for one reason or the other. Because of that I pledge the people to first tell and discuss their plans on the list before they start working on bigger patches. Committers work is an expensive good I do not want to waste.

aquiles2k:
Nah, thats totally fine :) Dont worry it wont stop me from contributing!

PR #264, opened 2017-02-25
Qt font caching

Karry:
Fonts should not be cached just by its size, but using other parameters too (font name). It makes possible to change rendering font without creating new `MapPainterQt`.


aquiles2k:
But it looked cool having different fonts at different zoom levels 😃

PR #267, opened 2017-03-05
fix osmscout::GetLineIntersection for two different zero size vectors

Karry:
...such vectors can't intersects, and add simple test for this method

When I was playing with coastlines (https://github.com/Framstag/libosmscout/pull/190) I found two intersecting zero size vectors, one in Italy and second in Greek ;-) I added simple check for that corner case...


Framstag:
Seem like the broken build just timed out (because of travis internal problems?). I restarted the build.

Issue #276, opened 2017-03-14
Add explicit download timeout to Qt network api


Karry:
Qt don't support to configure download timeout. System default (about one hour) is used! This cause problems when you start map download (or online map tile) and go away from wifi signal... Downloading stucks then. Lost tcp connection is detected (by app) after hour and downloading is restarted. It is toooo late!

All tutorials for Qt networking api suggest to add custom download timeout and reset it when some data arrives (signal `QNetworkReply::downloadProgress`).

rinigus:
I will work on it by merging FileDownloader into the upstream. Timeouts are incorporated there already , together with resumption of downloads .

Framstag:
Since the pul request was merged...can we close this issue?

Karry:
I have sad news. File downloader cant recover from network transition - I started downloading on wifi, then go outside. Dowloading continue after few seconds and it seems that it was finished. But target directory contains just few (3) files. System log contains threse errors:

rinigus:
@Karry, I wonder if you propagated the error message by the downloader to GUI? Its possible that it didn't manage to download any data for a freshly started file and stopped with "network unreachable" error during transition. Note that the downloader is not bullet proof, but I found that it gives a reasonable result. 

Now, when I use the mapmanager in the OSM Scout Server, only files that are missing on the disk. That way, user can continue downloading dataset for the missing files only. That mechanism is probably missing in current implementation of the manager in libosmscout.

As for  this issue, I guess we are not talking about timeouts anymore?

Karry:
I handle error messages in QML here: https://github.com/Karry/osmscout-sailfish/blob/master/qml/pages/Map.qml#L21

It seems that it timeout correctly, but retry after timeout don't work... Should I create new issue for that?

Karry:
...downloading fails even with stable connection. If I understand it correctly, you reset `m_download_last_read_time` on download start and in `FileDownloader::onProcessRead` method. But this method is called only when you write data to the "unpack" pipe... When you use `Plain` mode, and download of one file will take longer than 60 seconds, downloading fails every time... You should reset time in `FileDownloader::onBytesWritten` too...

rinigus:
@Karry, I haven't tested it that much in Plain mode. Its reset in onnetworkread method, which is the same as onprocessread for Plain mode. So, it should be sufficient. Do you have an error message with which it fails?

Karry:
Just "timeout" :-)

rinigus:
[...]

PR #278, opened 2017-03-19
add FileDownloader to Qt client library

rinigus:
This PR replaces FileDownloadJob with FileDownloader from OSM Scout Server. It addresses issue #276 as well as offers resuming of downloads and keeps an eye on internal Qt buffers. All in all, it should provide with the working and tested implementation of downloads.

There are few issues regarding this PR:

* I tested the code in OSM Scout Server and it works quite well over there. On OSMScout2, the testing was limited

* Since FileDownloader already includes resumption of downloads, there is no need to have resumption at MapDownloadJob level. In my understanding, MapDownloadJob::onJobFailed should propagate the error message to the user. GUI hooks are out of the scope of this PR and, in my mind, should be dealt on application and not library level.

* Since there is only one download at a time, several functions in MapDownloadJob were simplified. In this way, we have only job[0] which is downloading and its deleted as soon as all is done.

* Coding style follows the style I am using for OSM Scout Server: member variables are prefixed with m_ and since my editor is not configured for table-like variable declarations, I have not done it over here either. FileDownloader though is a separate class and I hope its OK. When changing some other code inside the library, I do try to follow the coding style around it.

### DIFF
--- a/libosmscout-client-qt/CMakeLists.txt
+++ b/libosmscout-client-qt/CMakeLists.txt
@@ -18,6 +18,7 @@ set(HEADER_FILES
     #include/osmscout/private/Config.h
     #include/osmscout/ClientQtFeatures.h
     include/osmscout/DBThread.h
+    include/osmscout/FileDownloader.h
     include/osmscout/InputHandler.h
     include/osmscout/LocationEntry.h
     include/osmscout/LocationInfoModel.h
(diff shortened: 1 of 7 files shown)


rinigus:
Thank you!

Karry:
Thank you @rinigus . It looks good. I will test it when I will be near of  my computer :-)

rinigus:
Great, its always good to test properly. When I was testing as a part of the server, I was plugin off/on network and it was able to deal with large amount of errors. However, note that if you start download without network, it would give you an error alerting that the net is unavailable. Such scheme works well for mobile - we have sometimes bad connections or "airplane" mode and through such distinction we can track the both cases reasonably well.

The Map Manager code of OSM Scout Server has been updated in the last release (today). I hope there are no significant bugs left, let's see how ultimate test via users would go. After some testing time we would have to decide what to do with it. Probably would be good to read it before deciding whether to merge it here as well.

PR #280, opened 2017-03-21
File downloader filename

Karry:
Hi @rinigus please review it if this is ok for you. I am using `getFileName` as part of progress message visible to user, I would like to keep just name there (not path)...


rinigus:
@Karry, I don't have  a chance to  properly test it within few next days, but I think it's fine and should work as expected. Also, for your application users it would be better to get just filename, as you  propose. so, it looks good to me .

Framstag:
Change ok for rinigus, merging

PR #281, opened 2017-03-22
Fix 'M_E': undeclared identifier error.

hexonxons:


Karry:
Hi. What platform/compiler cause such problems? I suggest to define this constant in library `Math.h`:

```
#if !defined(M_E)
# define M_E		2.7182818284590452354	/* e */
#endif
```

Btw, I select natural `log` for jump animation because it looks more natural... Using `10` dont make sence here, it is necessary to `positionScale` be 1.0 when `progress` is 100% (1.0).

Framstag:
So, does anyone offer a new, better patch?

hexonxons:
@Karry Win10 cmake+msvc15 shared library build.  
> Using 10 dont make sence 

Agreed. I just didn't figured out what shall be there.  
I could move it to `Math.h` tomorrow, UTC+3

Framstag:
Thanks @hexonxons, that would be great!

hexonxons:
@Framstag @Karry  Done

PR #286, opened 2017-03-24
Small refactoring and fixes for FileDownloader

Karry:
Hi. 

This MR should fixes https://github.com/Framstag/libosmscout/issues/276 I will test it on phone today...

The main problem was that timeout timer was started on downloader constructor, but it have to be started when real download is started...

Beside this fix, I rewrite usage of Qt signals and initialise osmscout logger by system environment variable in OSMScout2...

@Framstag , @rinigus , please review it

### DIFF
--- a/OSMScout2/src/OSMScout.cpp
+++ b/OSMScout2/src/OSMScout.cpp
@@ -53,6 +53,25 @@ static QObject *ThemeProvider(QQmlEngine *engine, QJSEngine *scriptEngine)
     return theme;
 }
 
+#define DEBUG   4
+#define INFO    3
+#define WARNING 2
+#define ERROR   1
+
+static int LogEnv(QString env)
+{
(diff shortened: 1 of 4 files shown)


rinigus:
@Karry, yes, it looks that you are right regarding too early start of the timer. In OSM Scout Server, I don't construct FileDownloaders for a whole bunch of files, but construct when I want to start a download. As a result, download is started in constructor, together with the timer. In your Map Manager, instead of keeping jobs as a list of files and URLs, you immediately make FileDownloader objects. As a result, this bug was introduced.

As for signals/slots and other rewrites - all is fine if its preferred style in libosmscout. I prefer to have the signals/slots connected by compiler, if possible, and not at the run time to deal with issues while I compile.

One comment - in my mind, when FileDownloader fails that failure should be final.

Issue #282, opened 2017-03-23
TransPolygon may create complex polygon/way (not simple)


Karry:
Hi. When I debugging one issue with coastline (water on left side of coastline), I found that it is caused by `TransPolygon` used from `WaterIndexGenerator::GetCoastlineData`. It simplify one way and from simple coastline creates complex one (line intersects itself). It cause problems in cell intersection detection and following computation steps that expect that coastlines are simple...

It would be great to have option in `TransPolygon` that avoids creating such problems :-)

![bad-optimised-way](https://cloud.githubusercontent.com/assets/309458/24239561/b153946a-0fae-11e7-9344-52e28e6855ff.png)

Framstag:
There are two optimization variants for TransPolygon. Does the problem occur for both of them?

Karry:
Yes, this problem occurs with both variants, `OptimizeMethod::fast` and `OptimizeMethod::quality`. Be concrete, this problem occurs on [this Italy coast line](http://www.openstreetmap.org/way/47913421#map=17/43.90862/8.08878), on zoom level 12 and it cause this error:

![italy-intersecting-coastline](https://cloud.githubusercontent.com/assets/309458/24317377/bbea2594-10f6-11e7-9972-75a36cb185bd.png)

I can look at it after weekend if you want to focus on something else or you don't have time for it...

Framstag:
It would be nice if you can take a look. I likely would not have the time neither do I have an idea for a quick fix.

PR #290, opened 2017-03-29
TransPolygon should generate simple way/area

Karry:
work in progress... I want to fix https://github.com/Framstag/libosmscout/issues/282

* add test (currently broken) that TransPolygon generates simple way/area


Framstag:
Hello Karry: You need to update the branch before I can merge. If you have added a test that fails, why does the build not break?

Karry:
> why does the build not break?

Because this test was not in "checked tests" that are running as part of the build. Build should be failing now, after rebase...

Karry:
Hi, it should be working. I enforce simple coastlines in WaterIndexGenerator, and it fixes this issue: https://github.com/Framstag/libosmscout/issues/282

WaterIndexGenerator is little bit slower now, `FindIntersection` method is not optimized well...

Framstag:
I haven't check the code for correctness, but regarding performance there are some possible change you should check.  Also instead of a bool parameter an enum might be more suited.

Nevertheless: Thanks for taking a look!

Framstag:
Is the patch for the added Test still necessary? Looks like it is included here, too?

Karry:
Sorry, what Test do you mean?

Karry:
To performance of searching path intersection, it can be improved in the future. I wrote similar method here: https://github.com/Framstag/libosmscout/pull/190/files#diff-d0db01360850c3c8acce2050fc081f31R266 and it is much faster. It may be generalised and used on both places...

Framstag:
There was a pill request named "Test" yesterday, but it now has disappeared. Perhaps I missed a refresh in GitHub.

Karry:
You merged it already: https://github.com/Framstag/libosmscout/pull/289 I just rebase this one against master...

Framstag:
I did start a GitHub review but just have seen, that it was not finished :-/ I finished it now. Did you get my comments already earlier or just now?

Karry:
I see review comments when it is finished. Not before.

I added (const) references, it is better. But how should I replace `bool requireSimple`? Should I additional flags? Or just enum like this:
```
enum OutputRequirements
{
  none = 0,
  simple = 1
}
```

Framstag:
regarding visibility of comments: OK.

Regarding RequireSimple: Yes I suggest using an enum. For toggles that are not obviously of boolean character, using an enum create IMHo a better API. It also makes call to the method easier to grasp.

Karry:
Ok, updated. I add `OutputConstraint` enum...

Framstag:
Ok.  Can you make a new issue for your suggested optimization.

Issue #49, opened 2016-06-22
Allow to configure an initial offset for shapes on lines


Framstag:
Allow to configure an initial offset for shapes on lines

PR #291, opened 2017-04-01
Add support for “embankment” tag 

vyskocil:
(https://wiki.openstreetmap.org/wiki/Key:embankment)

Ways with embankment tag set to yes are often used like bridges, when these ways cross some part of seas or oceans the water detection code is confused and classify wrongly the corresponding tiles as ground instead of water.  Of course the EMBANKMENT keyword could also be used to style ways like TUNNEL and BRIDGE.
I modified the scanner and parser code by hands to make some tests because I don’t have the tools to compile the grammar file, could you generate this files properly ?

### DIFF
--- a/StyleEditor/src/Highlighter.cpp
+++ b/StyleEditor/src/Highlighter.cpp
@@ -90,7 +90,7 @@ void Highlighter::updateRules()
 
     kwTUNNELBRIDGEFormat.setFontWeight(QFont::Bold);
     kwTUNNELBRIDGEFormat.setForeground(QColor("#25a13b"));
-    rule.pattern = QRegExp("TUNNEL|BRIDGE");
+    rule.pattern = QRegExp("TUNNEL|BRIDGE|EMBANKMENT");
     rule.pattern.setMinimal(true);
     rule.format = kwTUNNELBRIDGEFormat;
     highlightingRules.append(rule);
--- a/libosmscout-client-qt/libosmscout-client-qt.pro
+++ b/libosmscout-client-qt/libosmscout-client-qt.pro
@@ -25,7 +25,10 @@ SOURCES += \
     src/osmscout/SearchLocationModel.cpp \
     src/osmscout/Settings.cpp \
     src/osmscout/TileCache.cpp \
-    src/osmscout/TiledDBThread.cpp
+    src/osmscout/TiledDBThread.cpp \
+    src/osmscout/MapStyleModel.cpp \
+    src/osmscout/StyleFlagsModel.cpp \
+    src/osmscout/MapObjectInfoModel.cpp
 
 
 HEADERS += \
@@ -52,7 +55,10 @@ HEADERS += \
     include/osmscout/TileCache.h \
     include/osmscout/TiledDBThread.h \
     include/osmscout/private/ClientQtImportExport.h \
-    include/osmscout/private/Config.h
+    include/osmscout/private/Config.h \
+    include/osmscout/MapStyleModel.h \
+    include/osmscout/StyleFlagsModel.h \
+    include/osmscout/MapObjectInfoModel.h
 
 macx {
     QMAKE_CXXFLAGS = -mmacosx-version-min=10.7 -std=gnu0x -stdlib=libc++
--- a/libosmscout-import/src/osmscout/import/GenWaterIndex.cpp
+++ b/libosmscout-import/src/osmscout/import/GenWaterIndex.cpp
@@ -633,6 +633,7 @@ namespace osmscout {
 
     BridgeFeatureReader bridgeFeatureRader(typeConfig);
     TunnelFeatureReader tunnelFeatureRader(typeConfig);
+    EmbankmentFeatureReader embankmentFeatureRader(typeConfig);
     FileScanner         scanner;
     uint32_t            wayCount=0;
 
@@ -658,6 +659,7 @@ namespace osmscout {
             !way.GetType()->GetIgnoreSeaLand() &&
             !tunnelFeatureRader.IsSet(way.GetFeatureValueBuffer()) &&
             !bridgeFeatureRader.IsSet(way.GetFeatureValueBuffer()) &&
+            !embankmentFeatureRader.IsSet(way.GetFeatureValueBuffer()) &&
             way.nodes.size()>=2) {
           std::set<Pixel> coords;
 
--- a/libosmscout-map/include/osmscout/StyleConfig.h
+++ b/libosmscout-map/include/osmscout/StyleConfig.h
@@ -169,6 +169,7 @@ namespace osmscout {
   private:
     BridgeFeatureReader      bridgeReader;
     TunnelFeatureReader      tunnelReader;
+    EmbankmentFeatureReader  embankmentReader;
     AccessFeatureValueReader accessReader;
 
   public:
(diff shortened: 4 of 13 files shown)


Framstag:
The code looks straight forward and ready to merge (the fix of the grammar should be easy).  I do not see any change the the standard.oss file? So where is the actual use? It would be nice to have some concrete enhancement of the style sheet to test the feature (and the name/link to an import file and a coordinate :-)).

vyskocil:
There is already on concrete use of embankment in GenWaterIndex.cpp :-)
I didn't changed the standard.oss file because my map.ost file has somewhat deviated from the standard one and my main interest in this was not the rendering but the coastline ;-)  but I'll try to also provide some sample rendering...

vyskocil:
I tried something but can't do really what I want : 
screen shot 2017-04-01 at 22 47 31

```
SYMBOL embankment
  RECTANGLE 0,2.5 0.1 x 5 {
    AREA { color: #000000; }
  }
  RECTANGLE 2,1.5 0.1 x 3 {
    AREA { color: #000000; }
  }

   [TYPE highway_unclassified, highway_residential SIZE 6m 0.1mm:3px<] {
     [EMBANKMENT] {
       WAY#lbridge {color: @bridgeColor; displayWidth: 0.2mm; offset: -3m; displayOffset: -0.25mm;zIndex: -2;}
       WAY#rbridge {color: @bridgeColor; displayWidth: 0.2mm; offset: 3m; displayOffset: 0.25mm;zIndex: -2;}
       WAY.SYMBOL { symbol: embankment; }
     }
   }
```
The missing features are (maybe I just overlooked them ?) : 
- a parameter  to set the space between two symbols in WAY.SYMBOL
- draw the symbol below the lines (zIndex ?) or have a offset

vyskocil:
After more tests it seems that there is some missing initialization for the embankment feature but I can't find where... Actually when I add a [EMBANKMENT] { ... }  style to a path class it is drawn for every instance if the corresponding path has a embankment=yes tag or not !

Framstag:
Regarding space between symbols: That would be a variant of #49?

The PathSymbolStyle has the following attributes:

```c++
  class OSMSCOUT_MAP_API PathSymbolStyle
  {
  public:

    enum Attribute {
      attrSymbol,
      attrSymbolSpace,
      attrDisplayOffset,
      attrOffset
    };
```

Looking at MapPainter::DrawWayDecorations() and MapPainterQt::DrawContourSymbol() it looks like the space between symbols is evaluated by

```c++
double symbolSpace=projection.ConvertWidthToPixel(pathSymbolStyle->GetSymbolSpace())
```

Perhaps the iOSX backend does not evaluate this yet?

Looking at MapPainter::Draw() Symbols are currently always drawn on top of the actual path. That can get extended by another style parameter (before/after). But in this case one should make sure that we do not query twice the StyleConfig fo the PathSymbolStyleRef() (it should possibly be moved to WayPathData?).

Regarding EMBARKMENT always being set: I would extend DumpData to drop this information to see if the data is wrong in the ways or if the style sheet evaluation is wrong. If you cannot find it, tell me, I the rest of the code it done, I can try to tale a look during merging (and fixing the OSS.atg file).

vyskocil:
Regarding DumpData, I had nothing to extend because EMBANKMENT is a boolean and DumpData already handle it (good work ! ;-)

```
$ DumpData . -w 46341489
File './router.dat' - Cannot open file for reading: No such file or directory
Cannot open routing database
Way {
  OSM id: 46341489
  fileOffset: 277153044
  type: highway_footway
  boundingBox: [43.61827 N 7.04796 E - 43.61853 N 7.04815 E]
  center: 43.61840 N 7.04806 E
  cell level: 19

  Name: Sentier pour tous
  Bridge: false
  Tunnel: false
  Embankment: true
  Roundabout: false
  Access {
    foot: both
  }

  node[0] { serial: 1 id: 3238132417897078017 lat: 43.61827 lon: 7.04815 }
  node[1] { serial: 1 id: 3238132422187964161 lat: 43.61853 lon: 7.04796 }
}

$ DumpData . -w 330676782
File './router.dat' - Cannot open file for reading: No such file or directory
Cannot open routing database
Way {
  OSM id: 330676782
  fileOffset: 277153080
  type: highway_footway
  boundingBox: [43.61853 N 7.04778 E - 43.61871 N 7.04796 E]
  center: 43.61862 N 7.04787 E
  cell level: 19

  Name: Sentier Pour Tous
  Grade: 1
  Bridge: false
  Tunnel: false
  Embankment: false
  Roundabout: false
  Access {
    foot: both
  }

  node[0] { serial: 1 id: 3238132422187964161 lat: 43.61853 lon: 7.04796 }
  node[1] { serial: 1 id: 3238132422196400385 lat: 43.61871 lon: 7.04778 }
}
```
We can see that the import process is ok in regards of embankment tag, it is true when it is also in the imported data (I verified that way with OSMId=46341489 has embankment=true in OSM) and false elsewhere.
Then the issue should be in the drawing code, perhaps even in my hand editing of the scanner and parser code ?

Framstag:
I have checkout the branch locally and try to see what is happening (or not ;-)).

Framstag:
Looks like you definitely cannot patch the parser this simple. The whole lookup tables have changed after regenerating the scanner/parser. So possibly the parser did strange things.

Framstag:
Have now downloaded "Provence Alpes-Cote-d'Azur" from geofabrik in the hope that it contains way 46341489.

Framstag:
Btw. you can download the compiler generator from http://www.ssw.uni-linz.ac.at/Coco/. You should use the C++ version. It only has to be somewhere int he search path.

Framstag:
```
src/DumpData ../maps/provence-alpes-cote-d-azur -w 46341489
Way {
  OSM id: 46341489
  fileOffset: 11223100
  type: highway_footway
  boundingBox: [43,61827 N 7,04796 E - 43,61853 N 7,04815 E]
  center: 43,61840 N 7,04806 E
  cell level: 19

  Name: Sentier pour tous
  Bridge: false
  Tunnel: false
  Embankment: true
  Roundabout: false
  Access {
    foot: both
  }

  node[0] { serial: 1 id: 3238132417897078017 lat: 43.61827 lon: 7.04815 }
  node[1] { serial: 1 id: 3238132422187964161 lat: 43.61853 lon: 7.04796 }
}
```

=> OK

Framstag:
Looks like in principle it works. https://www.openstreetmap.org/way/46341489 though does have have the right type to match against the style sheet entry given above.

If I add the EMBANKMENT section to the code below
```
   [TYPE highway_footway] {
     WAY#outline {color: #ffffff50; displayWidth: 1.2mm; cap: butt;}
     WAY {color: #f64762; dash: 1,1; cap: butt; displayWidth: 0.4mm; }

     [BRIDGE] {
       WAY#lbridge {color: @bridgeColor; displayWidth: 0.2mm; displayOffset: -0.6mm;}
       WAY#rbridge {color: @bridgeColor; displayWidth: 0.2mm; displayOffset: 0.6mm;}
     }
     
     [EMBANKMENT] {
       WAY { color: #ff0000; width: 8m;}
     }     
   }
```

bildschirmfoto von 2017-04-03 21-03-49

Framstag:
Should I send you the generated stubs and you retest again everything?

vyskocil:
it seems alright, AMBANKMENT attribute is only drawn where it should. You may merge it, I'll sync my local repository after.

Framstag:
Manually merged.

PR #296, opened 2017-04-10
add test for intesection of two moved vectors

Karry:
Hi. I found two examples where `LinesIntersect` is broken. I added test for it. Can someone take a look at it? It should be simple to solve...


Karry:
Hi Tim. I fixed special case when two lines are on one straight line. For example `(1,0) -> (2,0); (3,0) -> (4,0)`. I hope that I understand to used algorithm and don't broke something...

PR #307, opened 2017-04-29
ImportErrorReporter: initialise tagErrorCount variable to 0

Karry:
...importer detect [7451613997849147769 errors in Laos](http://osmscout.karry.cz/asia/laos-12-20170429-1532/tag.html) ;-)


Framstag:
:-)

Issue #284, opened 2017-03-23
Allow multiple different views/renders of offline map database at the same time [$50]


aquiles2k:
The most common usage of the current map view (qt) is one central map and the current implementation is based on this. 

For some other usage scenarios it would be nice to also have the ability to have multiple views e.g. side by side, overlayed (think overview map in a corner) or on different widgets in a desktop app.

For this i think the implementation has to be extendended so that its clear which view is requesting updates and which cache data has to be kept etc.

Some initial discussion https://sourceforge.net/p/libosmscout/mailman/message/35588323/

Framstag:
Can you give an example, where such feature would be helpful? Perhaps we can add it to the OSMScout demo and then have a new feature and a reproducer at hand. Else I would have do build a simple test application to reproduce the problem.

Karry:
It would be usefull in styleeditor for example. When it will have multiple map views, you can preview how your changes looks like on different zoom levels or even different places...

It is possible already with tiled db thread. But it is not possible with plane...

I would like to refactor db thread (split database management, search, rendering... to separate classes/threads) and allow such usage. But it is low on my todo list.

Framstag:
@aquiles2k: Are you using the existing libosmscout-client-qt and have thus the problem mentioned by @Karry, that the current DBThread implementation doe snot support this?

@Karry: I assume that the rendering code would have to move into the MapWidget while the data acquisition code could stay in in the DBThread?

aquiles2k:
The only test i did back then was putting two Map Items in qml side by side. This resulted in heavy flickering on the maps once i started dragging one of the maps and i came to the conclusion its because of all things said like heavy coupling of the rendering code with the database without differentiating which view gets which updates and which cache gets deleted etc.. 

Yes we use the client-qt-lib but avoided these problems by removing the singleton and simply have each widget have its own bundle of mapview and associated database. I think i mentioned this once on the list. 

Anyway i know this is kind of overkill and thats why i created this issue but for the time being it solved our problem :)

Karry:
*Plane* DBThread don't support it, but there is no such problem with *Tile* rendering...

Well, I am thinking about refactoring of Qt API long time. DBThread instance now contains *everything* (rendering, object lookup, searching...) and some API methods accessible from UI thread are blocking! I would like to add another layer (renderer) between MapWidget and database management... I am thinking about something like this:

[...]

PR #309, opened 2017-05-05
Qt refactoring: basic concept, LookupModule

Karry:
Hi all. 

This is first step of Qt API refactoring that I mention here: https://github.com/Framstag/libosmscout/issues/284#issuecomment-295604526

@Framstag @aquiles2k Please review API changes. Rendering is still in `DBThread`, I started with simple `LookupModule` that is used from `MapObjectInfoModel`. What was changed:

## Library initialisation

Instead of usage `DBThread` singleton as a central point where you have to choose between plane and tiled variant of rendering, I created `OSMScoutQt` singleton that is used to accessing `DBThread`, `Settings` and creating modules for models... Its initialisation uses builder pattern:

```
OSMScoutQt::NewInstance()
     .WithStyleSheetDirectory(stylesheetFile.dir().path())
     .WithStyleSheetFile(stylesheetFile.fileName())
     .WithIconDirectory(iconDirectory)
     .WithMapLookupDirectories(mapLookupDirectories)
     .WithOnlineTileProviders(":/resources/online-tile-providers.json")
     .WithMapProviders(":/resources/map-providers.json");
     .Init()
```

## Database access

I replaced database mutex with read-write lock, that will enable more readers at same moment. New functionality that is accessing database should not be added to `DBThread` (it should care just about databases, not for routing, searching...), but should be created `DBJob` that will acquire reader lock and do its business: 

```
  loadJob=new DBLoadJob(lookupProjection,maximumAreaLevel,/* lowZoomOptimization */ true);
  
  connect(loadJob, SIGNAL(databaseLoaded(QString,QList<osmscout::TileRef>)),
          this, SLOT(onDatabaseLoaded(QString,QList<osmscout::TileRef>)));
  connect(loadJob, SIGNAL(finished(QMap<QString,QMap<osmscout::TileId,osmscout::TileRef>>)),
          this, SLOT(onLoadJobFinished(QMap<QString,QMap<osmscout::TileId,osmscout::TileRef>>)));

  dbThread->RunJob(loadJob);
```

DBJob have to release its reader lock when its job is finished or when it is destroyed...


Framstag:
I'M currently looking through the code (looks good up to now). The Job API you mentioned look at little bit like working with Promise/Future. Does Qt provide something similar and would it be helpful to use such pattern? You will likely still need the signal/slot mechanism for getting notified though, too?

libosmscout-client-qt/src/osmscout/DBJob.cpp
```diff
@@ -0,0 +1,193 @@
...
+    for (auto &tile:tiles){
+      if (tile->IsComplete()){
+        emit tileStateChanged(path,tile);
+      }
+    }
+
+  }
+}
+
+void DBLoadJob::onTileStateChagned(QString dbPath,const osmscout::TileRef tile)
```
Framstag: Spelling mistake in function name

Framstag: const QString& for dbPath to optimize? I also use const& for XXXRef classes, though that usage could be discussed.

Karry:
Hi. Exactly. I expect that jobs may report computation progress (router, load...) or partial result (search, reverse lookup...) by signals. Futures dont fit to this.

With reference usages, I am not sure how it works with queued Qt signal connection. I expect that all arguments are copied anyway...? I will try to find it.

aquiles2k:
Sorry for not responding earlier, this looks good so far, though i cant really say truly until i have to use it myself. As far as i understand this has not yet decoupled different view rendering?

Karry:
@aquiles2k : Not yet, rendering is still in shared DBThread... I am working on it here: https://github.com/Karry/libosmscout/tree/qt-map-renderers plane rendering is rewritten, but there is some bug (probably concurrency) that cause random segfaults... help is welcome :-)

aquiles2k:
Ah thats pretty far along - i could have saved me the initial look at the pipeline in #284 then :)

So it looks like your giving each MapWidget its own MapRenderer which handles the rendering stuff the dbthread did before? I'll have a look.

aquiles2k:
Ok i tested it and it works great so far for me - no segfaults. Do you have some reproducible way i should test?

I tested it with just adding a second very simple Map below the other Map in OSMScout2 and started heavily paning and zooming in both maps.

aquiles2k:
Ok right after writing the last comment i got a segfault :)

Interestingly enough it would the crash on every new startup with the first render. Unfortunately as soon as i started to put in debug logging, it disappeared. I did also found out that putting qDebug() calls in OSMScoutQtBuilder::Init will segfault it too :)

aquiles2k:
Ok some pretty good way to let it segfault for me is to resize the whole Application window - probably because he then gets redraw calls for both views at nearly the same time.

aquiles2k:
@Karry So after finding out that the resize crashes was a problem on my side i retried and so far couldnt get a crash. Do you have some reproducible way on your side for the segfaults?

Karry:
It seems that segfaults was caused by using (read) `QList` instance from multiple threads. `std::list` is (should be) safe for using from multiple thread (for read). It seems to be stable now...

Karry:
Hi. It seems that there is (was) several causes of segfaults. One was caused by using `QList` (read) instance with databases from multiple threads. `std::list` is (should be) safe for using from multiple threads (for read)...

Another segfault is probably caused by `DataFile` that has some methods without mutex synchronisation...

Lets move discussion here: https://github.com/Framstag/libosmscout/pull/336 It seems that there is a lot of fun with multithreading bugs :-(

Issue #313, opened 2017-05-07
Compiler warnings in GenWaterIndex.cpp


Framstag:
See https://ci.appveyor.com/project/Framstag/libosmscout/build/job/644tpk6kyk8u0q3h

```
:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenWaterIndex.cpp: In member function 'bool osmscout::WaterIndexGenerator::WalkFromTripoint(osmscout::GroundTile&, const osmscout::WaterIndexGenerator::Level&, const osmscout::WaterIndexGenerator::CellBoundaries&, osmscout::WaterIndexGenerator::IntersectionRef&, osmscout::WaterIndexGenerator::IntersectionRef&, osmscout::WaterIndexGenerator::Data&, const std::__cxx11::list<std::shared_ptr<osmscout::WaterIndexGenerator::Intersection> >&, const std::vector<long long unsigned int>&)':
C:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenWaterIndex.cpp:2161:36: warning: 'outgoingAngle' may be used uninitialized in this function [-Wmaybe-uninitialized]
       if ((!outgoing) || (outgoing && angle < outgoingAngle)){
                          ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
C:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenWayWayDat.cpp:310:0: warning: ignoring #pragma omp critical [-Wunknown-pragmas]
 #pragma omp critical
 
C:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenWayWayDat.cpp:338:0: warning: ignoring #pragma omp critical [-Wunknown-pragmas]
 #pragma omp critical
 
C:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenWayWayDat.cpp:367:0: warning: ignoring #pragma omp critical [-Wunknown-pragmas]
 #pragma omp critical
 
C:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenWayWayDat.cpp:774:0: warning: ignoring #pragma omp parallel [-Wunknown-pragmas]
 #pragma omp parallel for
 
C:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenWayWayDat.cpp:783:0: warning: ignoring #pragma omp critical [-Wunknown-pragmas]
 #pragma omp critical
 
[...]

Issue #312, opened 2017-05-07
Compiler warnings in GenCoordDat.cpp


Framstag:
See https://ci.appveyor.com/project/Framstag/libosmscout/build/job/644tpk6kyk8u0q3h

```
C:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenCoordDat.cpp: In member function 'bool osmscout::CoordDataGenerator::FindDuplicateCoordinates(const osmscout::TypeConfig&, const osmscout::ImportParameter&, osmscout::Progress&, std::unordered_map<long long unsigned int, unsigned char>&) const':
C:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenCoordDat.cpp:108:16: warning: unused variable 'oldUpperLimit' [-Wunused-variable]
             Id oldUpperLimit=currentUpperLimit;
                ^~~~~~~~~~~~~
C:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenCoordDat.cpp: In member function 'bool osmscout::CoordDataGenerator::StoreCoordinates(const osmscout::TypeConfig&, const osmscout::ImportParameter&, osmscout::Progress&, std::unordered_map<long long unsigned int, unsigned char>&) const':
C:/projects/libosmscout/libosmscout-import/src/osmscout/import/GenCoordDat.cpp:283:19: warning: unused variable 'oldUpperLimit' [-Wunused-variable]
             OSMId oldUpperLimit=currentUpperLimit;
                   ^~~~~~~~~~~~~
```

Framstag:
This seems to be already fixed by the introduction of the unused macro.

PR #319, opened 2017-05-11
Compiler warnings

Karry:
Try to solve some compiler warnings mentioned here:

https://github.com/Framstag/libosmscout/issues/312
https://github.com/Framstag/libosmscout/issues/313

### DIFF
--- a/libosmscout-import/include/osmscout/import/GenCoordDat.h
+++ b/libosmscout-import/include/osmscout/import/GenCoordDat.h
@@ -29,6 +29,15 @@
 
 namespace osmscout {
 
+/**
+ * Some temporary variables used in asserts may be unused
+ * on production build, when asserts are wiped out.
+ *
+ * Usage of this macro for such variables makes production
+ * build happy, and avoids unused-variable warning
+ */
+#define _unused(x) ((void)(x))
+
   class CoordDataGenerator CLASS_FINAL : public ImportModule
   {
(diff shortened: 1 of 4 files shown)


libosmscout-import/include/osmscout/import/GenCoordDat.h
```diff
@@ -29,6 +29,15 @@
...
 namespace osmscout {
 
+/**
+ * Some temporary variables used in asserts may be unused
+ * on production build, when asserts are wiped out.
+ *
+ * Usage of this macro for such variables makes production
+ * build happy, and avoids unused-variable warning
+ */
+#define _unused(x) ((void)(x))
```
Framstag: That macro would be a good idea at a more central location. There is a small number of places, where it could be used, too. One question though: It underscore in macros allowed or is this reserved by the standard or similar (similar to two underscores...)

Karry: > ...more central location
I agree. We have more such warnings, where some arguments are unused just on windows... But where it should be? 
`osmscout/system/Assert.h`, `osmscout/system/Compiler.h` or somewhere else?

> Is underscore in macros allowed...
I have no idea :-) But underscore don't makes sense here, we can remove it...

Framstag:
I vote for Compiler.h

PR #329, opened 2017-05-21
prevent warning for autoclosing file

aquiles2k:
This avoids the "Automatically closing FileScanner for file " warning during location lookup


aquiles2k:
Hm, i have no clue about that WorkQueue test failing on cmake

PR #336, opened 2017-05-24
Qt refactoring 2: map renderers

Karry:
Hi. This MR moves rendering code from (Tiled/Plane) DBThread to new classes : Tiled / Plane Map Renderer. This change should allow multiple independent MapWidget instances ( https://github.com/Framstag/libosmscout/issues/284), every MapWidget has its own renderer with dedicated background thread...

Please test it carefully ( @aquiles2k ?), this code may contains some nasty, multithreading bugs.


Framstag:
@Karry, @aquiles2k: Since there were a number of additions/fixes over the day, can at least one of you give me an "ok for merging"?

Karry:
Please wait a bit. I am testing this branch on mobile and it seems that there is a bug that is caused by multithreading access to database. It leads to stack corruption, so gdb stacktrace is useless here :-/ At least with default compiler settings. When I compile it with `-fno-omit-frame-pointer -fstack-protector-all` app crashes faster - 
It seems that method `bool DataFile<N>::GetByOffset(IteratorIn, IteratorIn, size_t, std::vector<ValueType>& ) const` is not thread safe! Full stack trace: https://gist.github.com/Karry/4a6185f694e790615ae989ac83772ae4#file-crash-gdb-txt-L376

It is called from `LocationService::LoadNearNodes`...

Framstag:
I am already in The way to work so lookup of The actual code but it sounds strange that lookup of object by offset is not thread safe. I could only imagine that file handle my used globally and not thread local. 

One could change that by opening closing handles in each address, but would this not create considerable overhead. Also is my questionable if access to the same file from multiple threads is much faster than single threaded access (on mobile devices)?

Karry:
> Also is my questionable if access to the same file from multiple threads is much faster than single threaded access (on mobile devices)?

My subjective feeling is that application is faster. Rendering of map (and loading of map data) dont block location info and vice versa. Synchronisation in DataFile seems to add minor overhead...

aquiles2k:
Im testing it atm with OSMScout2 and a germany map using qt5.6.2 on mingw based win build. No crashes so far. The only think im getting from time to time at startup is:

QObject::moveToThread: Current thread (0x62e4790) is not the object's thread (0x5d601b0).
Cannot move to target thread (0x1406260)

QObject::moveToThread: Current thread (0x5d601b0) is not the object's thread (0x62e4790).
Cannot move to target thread (0x1406260)
For testing i've added a second Map below the original one in main.qml and a button to spawn a new Window like so:

MapButton {
  label: "n"
  onClicked: {
    var component = Qt.createComponent("main.qml")
    var window    = component.createObject(mainWindow)
    window.show()
  }
}

aquiles2k:
Im trying to integrate it into our code but right now it segfaults when doing the initial recentering.

This only happens in release, debug works \*sigh\*

Stack is always something like this (shortened):
```
Thread 13 received signal SIGSEGV, Segmentation fault.
[Switching to Thread 2016.0x1ab4]
0x00000000668e7e17 in QMetaObject::activate(QObject*, int, int, void**) ()
   from D:\Dev\Builds\tc_r64_5.6.2msys\bin\Qt5Core.dll
(gdb) bt
#0  0x00000000668e7e17 in QMetaObject::activate(QObject*, int, int, void**) ()
   from D:\Dev\Builds\tc_r64_5.6.2msys\bin\Qt5Core.dll
#1  0x000000006b2963dc in DBLoadJob::tileStateChanged (this=0x3, _t1=...,
    _t2=...)
    at D:/Dev/Compiler/msys64/home/adrian/dev/libosmbuild/r_newrenderer/libosmscout-client-qt/osmscout_client_qt_autogen.dir/moc_DBJob_664GOZI3ENSBD6.cpp:235
#2  0x000000006b288640 in DBLoadJob::<lambda(const TileRef&)>::operator() (
    tile=..., __closure=<optimized out>)
    at D:/Dev/Compiler/msys64/home/adrian/dev/libosm/libosmscout-client-qt/src/osmscout/DBJob.cpp:115
#3  std::_Function_handler<void(const std::shared_ptr<osmscout::Tile>&), DBLoadJob::Run(const std::__cxx11::list<std::shared_ptr<DBInstance> >&, QReadLocker*)::<lambda(const TileRef&)> >::_M_invoke(const std::_Any_data &, const std::shared_ptr<osmscout::Tile> &) (__functor=..., __args#0=...)
    at D:/Dev/Compiler/msys64/mingw64/include/c++/6.3.0/functional:1731
#4  0x000000000234ee74 in std::function<void (std::shared_ptr<osmscout::Tile> const&)>::operator()(std::shared_ptr<osmscout::Tile> const&) const (
    __args#0=..., this=0x191e2b08)
    at D:/Dev/Compiler/msys64/mingw64/include/c++/6.3.0/functional:2127
#5  osmscout::MapService::NotifyTileStateCallbacks (this=0x7828fb0, tile=...)
    at D:/Dev/Compiler/msys64/home/adrian/dev/libosm/libosmscout-map/src/osmscout/MapService.cpp:760
#6  0x0000000002352dd3 in osmscout::MapService::GetWays (this=0x7828fb0,
    parameter=..., wayTypes=..., boundingBox=..., prefill=false, tile=...)
    at D:/Dev/Compiler/msys64/home/adrian/dev/libosm/libosmscout-map/src/osmscout/MapService.cpp:599

```

Have you seen something like this before?

aquiles2k:
Ok error was on my side. I used AddMapLookupDirectories() from OSMScoutQtBuilder with my map directory directly besides my executable. Apparently the builders mapdirectory list gets initialized with current dir too, so i ended up haveing the same map two times. It seems osmscout didnt like that :)

Its running for now but i'll test some more tomorrow.

Karry:
You should call `WithMapLookupDirectories` if you don't want use default dir....

Karry:
`Cannot move to target thread` is probably Qt bug on windows. It is really strange :-/

Karry:
Hi. It seems to be stable after recent fixes: 

- delete thread when it is really finished
- delete objects with dedicated background thread by `deleteLater`
- avoid duplicates in open databases that may cause problems in `DBLoadJob` (https://github.com/Framstag/libosmscout/pull/336#issuecomment-304326801)

I think that we can merge it. Last problem is weird message on Windows: `Cannot move to target thread` found by @aquiles2k . I can't reproduce it on Linux and api should be used correctly...

aquiles2k:
Yea it only happens on startup when the initial view is on something object heavy like a city. Its not really a bug since the Job then just get handled in the mainwindow.

Issue #356, opened 2017-06-04
OSMScout2: Correctly handle drawing of map at mercator projection range limits


Framstag:
There are a number of situation where OSMScout2 stops rendering a map. Reason for this is:
* Mercator projection does have smaller limits than the planet (poles cannot be rendered)
* On low zoom map may not fit the screen, so there will be areas, where no map can be drawn, since the region has invalid coordinates on earth.
* Moving the map could result in moving around the poles (lat < -90 or > 90) or moving over lon -180/180.
* Resizing the windows may make invalid regions visible, without actually moving or zooming.

I tried to fix this by changing MercatorProjection so that is only allows valid values. However I'm not sure, that this enough. I assume that we have to change or enhance the Projection API to allow the client to act better.

@Karry: Please look at the current situation and your qt-client code. Let us then discuss how a proper solution could look like.

aquiles2k:
After some experimentation:
* I like the current restriction when going north but it still sometimes behaves a bit "choppy/laggy". 
* Also when going south i can go beyond antarctica until the restriction starts which looks strange. 
* I personally dont like that when zooming far out with base map active you often get black screens as if no map db is within view coords.

Framstag:
IMHO balck borders (or something similar) cannot be avoided. Under mercator projection (and any other projection of course, too) the world has some certain aspect ratio. If your application windows has another aspect ratio (on the desktop it can even change dynamically, while on mobile there will always be only two) than the world has, you have to find a solution to visualize this. Horizontally you could just wrap around (see issue #29) but vertically - at least for mercator projection - this does not make sense.

But that was exactly my intent. First clarify what we want to show and then take a look which information the applications needs from the projection, to get this solved.

Mercator projection (via the general Projection base class) could for example offer:
* Suggested aspect ratio
* Range limits for lat and lon
* Calculate "on range in bounds view point" in case where requested ranges are "invalid"

aquiles2k:
I dont meant black borders but the complete view goes black at certain positions, most likely because of the database in bounds check.

Framstag:
Ah, OK. It is also possible that the internal mercator projection or some other internal state regarding view point instance got into an invalid state. I tried to fix that but event processing and handling of the new view point in result were not obvious. The projection should not get invalid anymore but at least other parts of the state seems to assume that changing the view point of the projection always succeed. That is something @Karry should look into. I might break more than I heal.

aquiles2k:
Just for some visuals a gif :)

[...]

PR #364, opened 2017-06-08
Fix map rendering in Qt library when displayed area is undefined for Mercator projection

Karry:
Fixes for https://github.com/Framstag/libosmscout/issues/356

Tim, note that I changed `MercatorProjection::Set` behaviour. See updated documentation:

> Setup projection parameters.
>
> Return true on success,
> false if arguments are not valid for Mercator projection,
> projection parameters are unchnaged in such case.
>
> Note that coord (center) have to be valid coordinate
> in Mercator projection. But it is possible setup dimensions
> (width and height) that projection will cover area bigger
> than the one valid for Mercator projection. Bounding box
> is adjusted then to be valid for projection.
>
> In code:
>
>   projection.GetDimensions(bbox);
>   projection.GeoToPixel(bbox.GetMinCoord(),x,y)
>
> may be x >= 0


Framstag:
Changes look reasonable to me. Modified interface for Projection is OK for me. Thanks!

Issue #363, opened 2017-06-07
Investigate broken coastline on basemap


aquiles2k:
It seems there is some problem with the basemap coastline on russia/norway.
![brokencoast](https://user-images.githubusercontent.com/6095108/26894197-f1cfccc0-4bbd-11e7-9b22-d027b569fb0d.png)

Framstag:
I have already some initial checks.

* First assumption was that coordinates are broken during parsing. For this I checked the distance between directly following point in the coastline loaded from the shape file. There were no distances greater 1.0.
* Later on we do node optimization on the coastline (CalculateCoastlineData()). This results in bigger distance, bu no distance >5.0 was found. If we skip optimization, the problem is still visible.
* Next I dropped all coastlines whose boundary.size() was <1.0. No change, problem still there (but Greece lost most of its islands).

I have to think about that... @Karry, any idea?

Karry:
Hi. I reviewed shapefile in QGis tool (thanks @rinigus for tip) and it seems that coastlines crossing 180° longitude (antimeridian) are not closed! (Same way as usual OSM coastlines...)
![unclosed-coastlines](https://user-images.githubusercontent.com/309458/26963204-c1a874e6-4ceb-11e7-8013-abb4cddacc12.png)

In OSM data, when some coastline is crossing antimeridian, it is splitted to two closed areas, like [here on Wrangel Island](http://www.openstreetmap.org/way/23824393#map=7/70.745/-179.978)

It is most likely the root of the Russia and Antarctica problems... We should try:
 - edit shapefile manually to fit this restriction
 - find another source that will preprocess shapefile same way as OSM coastlines
 - add import step that will fix this

rinigus:
I wonder if Land Polygons could be used instead (see http://openstreetmapdata.com/data/land-polygons). Or http://openstreetmapdata.com/data/water-polygons . High-resolution variants are split into smaller boxes and, I hope, they deal better with antimeridian.

Framstag:
If I understand it correctly, these are then not costlines anymore but data is more structured like WaterIndex does it in the end. This might help, if we would throw away the current water index approach completely, but else we do not get coastlines - which are required by water index - easily from this.

Framstag:
Btw., they use https://github.com/osmcode/osmcoastline/blob/master/README.md for this. As you can see, there is no option regarding anti-meridian :-/

Editing the shapefile is no option. We do want to importer newer version of the coastline to get the newest fixes.

[...]

PR #366, opened 2017-06-12
Basemap import fixes

Karry:
Some fixes for basemap import https://github.com/Framstag/libosmscout/issues/363:
- it is necessary to merge coastlines, Euro-Asia coastline are splitted in shapefiles from `openstreetmapdata.com`
- then I closed coastline ways that are touching antimeridian + Antartica (it is little bit hack)
- and remove remaining unclosed coastlines

basemap

There are still some artefacts, but I think that it is rendering issue, not importer... 

Tim, I saw that coastlines from shapefile has different direction than is usual in OSM data. This is some rule? I see some "unfilled" islands, it is possible that coastlines in shapefile may be clock-wise and counter-clock-wise? Or it is some different bug?

### DIFF
--- a/BasemapImport/src/BasemapImport.cpp
+++ b/BasemapImport/src/BasemapImport.cpp
@@ -301,6 +301,7 @@ static bool ImportCoastlines(const std::string& destinationDirectory,
         if (std::abs(coastline->coast.front().GetLon())>179.999 &&
             std::abs(coastline->coast.front().GetLon()-coastline->coast.back().GetLon())<0.001){
           coastline->isArea=true;
+          coastline->coast.push_back(coastline->coast.front());
         }else if (coastline->coast.front().GetLon()<-179.999 && // reverse direction than OSM data!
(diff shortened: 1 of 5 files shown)


Framstag:
Thanks! I saw that the coastlines had different direction and I hasd to switch definition of land and sea but assume, that it is just the way the tool works. I though assumed that direction is always the same - simply because this is the way it works in OSM itself and I saw no reason why this should be different. I'm suprised, that one needs to merge coastlines, that was definitely not part of the documentation :-( The shape file format is rather simple, it is just a list of objects where each object has the same type (but different types are allowed) and consists of a list of coordinates. 

There was a problem though that coastline have been cut into individual polygons of as far as I remeber 5000 points. So there was no explicit marker which polygons make up a coastline. I used a heuristic: If the end coordinate was not == the start polygon and the polygon consists of exactly 5000 coordinates than the next part was part of the coastline, too. Thgis seems to be correct especislly for big coastlines. For smaller coastlines <5000 coordinates this should not be a problem.

Issue #365, opened 2017-06-09
OpenGL renderer: Enhance CMake file


fannymonori:
Check if the required packages have been installed, and skip the building of the OpenGL module if it does not find them.

Framstag:
The pull request from Luka did improve the situation, but the cmake build locally still errors:

```
[ 49%] Linking CXX shared library libosmscout_map_opengl.dylib
ld: library not found for -lGLEW
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [libosmscout-map-opengl/libosmscout_map_opengl.dylib] Error 1
make[1]: *** [libosmscout-map-opengl/CMakeFiles/osmscout_map_opengl.dir/all] Error 2
```
While cmake claims it can build the OpenGL backend, I get a linker error.

Karry:
Hi. What OS it is? `*.dylib` - it is Mac OS? Cmake GLEW integration (`FindGLEW.cmake`) is not included in libosmscout repository, so problem is probably in your GLEW instalation ;-)

Framstag:
Yes, this is Mac OS. FindGLEW.cmake is part of my cmake installation.

```
Tims-MBP:build tim$ ls -l /usr/local/share/cmake/Modules/FindGLEW.cmake 
-rw-r--r--  1 tim  admin  2158 31 Mai 16:00 /usr/local/share/cmake/Modules/FindGLEW.cmake
```

The library is also available:

```Tims-MBP:build tim$ pkg-config --libs glew
-L/usr/local/Cellar/glew/2.0.0/lib -lGLEW
```
But the linker call from the cmake generated makefile does not set the search path:

```
[...]

PR #367, opened 2017-06-14
cmake: detect glew, glm, glfw; skip OpenGL build when not found

Karry:
Hi @fannymonori , please review my changes. 

I am using Ubuntu 17.04, these libraries (dev packages) was missing on my system. CMake was fine with it, but build fails on missing header then... I hope that these changes don't break other platforms...


Framstag:
@fannymonori: The changes make sense and I already hinted at this, too. Since @Karry requested your review, I'm waiting for your OK :-)

fannymonori:
It works and looks OK. Also it closes issue #365  :)

PR #368, opened 2017-06-14
BasemapImport: make min/max index level configurable

Karry:
User may select basemap quality / size by specifying index levels. When you generate basemap just for level 4, it will have just 59 KiB !


Framstag:
Yes, good (and obvious) idea.

PR #375, opened 2017-06-19
Opengl rendering areas

fannymonori:
Implemented area-rendering pipeline. The rendering is working correctly, right now it is a bit distorted because of the not yet finished zooming, and also slow because the data loading, but those are also in progress.


Framstag:
I assume you have added sources to the opengl backend. This breaks the meson build, with like cmake requires list of sources. In such simple cases you do not need to worry about that, I can fix the meson files myself.

fannymonori:
Yes I added a few sources, and the meson builds did break.

PR #383, opened 2017-06-21
Fixing poly2tri files

fannymonori:


Framstag:
As before, I'll fix the meson build.

PR #395, opened 2017-07-04
Improved data loading on zoom event

fannymonori:


Framstag:
The first build broke because of some internal Travis problems. I restarted the failing build and it was OK. now however you broke the build :-)

fannymonori:
The error was: "call to 'abs' is ambiguous", and it seems like a problem with travis. Anyway I realized that using abs there is completely unnecessary.

Issue #57, opened 2016-06-25
PoC: Create a proof of concept for cross database routing


Framstag:
This proof of concept should take neighbouring OSM export from the same provider and realise the following tasks:
- Check that identical notes on the border are provides in both imports and have the same id
- Count the number of such route nodes that do exist in both imports (estimation for the size of a hash map that holds such "special case" ids).
- Find a cell size for an index that covers such border regions that hold "special interest" route nodes.
- Estimate which of the to approaches does require less memory,is more performant during runtime and does require less pre-routing processing.
- Add an abstracting/extention to the existing data structures that handles the lookup of nodes in multiple databases.
- Do some routing and performance tests.

Some backgound from one of my mails in response to above topic:

> There is currently no code that allows this. However the recent change 
> so that nodes (as part of ways or areas) and route nodes have an id 
> that is unique independent on the actual import allows in principle to 
> look for a specific route node in multiple databases.
> 
> There are of course some pre requirements:
> - The unique id is generated based on the coordinate. If different 
>   imports have different resolution of the coordinate the same node may 
>   get a different unique id.
> - The imports needs to have all nodes for the same coordinate since 
>   the unique id is based on the coordinate and a running number per 
>   coordinate.
> - The imports must overlap. That means that a way or area in the 
>   "border" must be available in both imports - or at least one node. So 
>   I route to the "end" of the database, then lookup for the same route 
>   node in the next database and continue the routing there.
> 
> You also need some kind of index that tells you which nodes you have 
> to look for in multiple databases, since there is no "look in other 
> databases, too"-maker. Such index could either be based on a 
> cell-approach (this cell is in the following databases...) or based on 
> the actual ids (this id is in multiple databases).

Framstag:
There is now MultiDBRouting.cpp, that realizes the first two points. Point three is now "only" some calculation/testing and running the tests on multiple database (belonging to more crowded and less crowded regions).

Karry:
Initial support (routing via two neighbouring databases) was implemented in #399. Can close this issue?

PR #399, opened 2017-07-06
Multi db routing

Karry:
Hi. 

I don't wanted to discuss multi-database routing https://github.com/Framstag/libosmscout/issues/57 before good knowledge of used algorithms. Now, it is right time - here is proof of concept. 

 - I added knowledge about source database to routing structures 
 - split `RoutingService` to `AbstractRoutingService`, that contains main routing algorithm, and `SimpleRoutingService` that support routing with one database
 - create `MultiDBRoutingService` that is based on Tim's `MultiDBRouting` test and use intersecting route nodes to jump from one database to another

It seems that it is working, it computes route from Prague to Vienna successfully! But it crash after computation in method `ResolveRNodeChainToList`. It seems that I broke chain of closed nodes. But it should be simple to solve.

Please review it :-)


Framstag:
The code in principle looks OK. A few question though, that I would else have to dig out of your code:
* You splitted the router into an interface with a simple and a multi db implementation. What would be the advantage to use the simple implementation? Is it faster? Does it thus make sense to keep both implementations?
* As I can see you switch from one database to the other. Under which conditions you take a look for a route node in both databases? I initial planed to reuse a new index, that tells me which tiles are supplied by which database?
* I assumed that a route node can be (exactly) on the cut out border, so it only holds partial information in each database. Thus I assumed I have (in some cases) to merge data for a route node that is available in both databases to make sure that it holds all ways. How do you handle this?
* You handle only overlap of two databases. What happens if more than two areas overlap (think of Bodensee?)? Will this be correctly handled.

Nevertheless, thanks for this fantastic work. Multi-DB routing is one of the few missing basic conceptional features! getting this to work is great :-)

Karry:
Hi. 

 - Multi DB routing service has different api and don't provide all possibilities yet - like routing via multiple routing nodes. Performance should be same, but I don't measure it yet. When complete functionality will be moved to abstract + multi-db routing service, we can remove the "simple" variant... Library users will have to use the new api, but change should be small.

 - I am using the same approach that you do in MultiDBRouting test - scan complete router.dat, store covered tiles from both databases, create intersection and then scan router.dat again and find common route nodes... It is slow, but it can be improved later by some special index or persisting common nodes after first computation - I plan to test it on phone and we will see if speed is usable... 

 - router iterate over open nodes and go through its ways and put targets to openList (it is same with current code), then load "node twins" from another databases, copy node costs and add it to openList too... So if node from one database don't contains some way because it was cropped, node from second database may contain it and it is used...

 - Current code supports routing via two databases only, when start and target databases overlaps. It can't find route between Prague and Paris (via Germany) now. But it may be improve later, abstract routing service is ready for it. It is "just" necessary to find fast method that will find common route nodes between multiple (N:N) databases.

You are welcome. I want this feature in my app :-) I will test restricted routes now and integrate new routing service to Qt router...

Framstag:
@Karry: You tell me, if I should merge?

Karry:
Not yet please. I will push next commits at Wednesday. I am out off civilisation now (writting from mobile).

Karry:
Hi Tim. You can merge it now. `RoutePostprocessor` was updated to support route via multiple databases, and Qt router use `MultiDBRoutingService`...

PR #417, opened 2017-07-15
move settings in cookiejar to osmscoutqt settings

aquiles2k:
PersistentCookieJar still had its own QSettings instance resulting in having some loose registry entry on win when using an external setting path.
Since the PersistentCookieJar is created in ctor of MapManager before the OSMScoutQt instance is created i had to hand the settings ref through all ctors into the cookiejar.


Karry:
lgtm

PR #430, opened 2017-07-29
Rendering texts

fannymonori:
Can render texts with any size and color.


Framstag:
You are missing freetype dependencies.

For meson you can tweak the build by adding ftDep.found() to the top level meson.build:
Existing code:
if openGLDep.found() and glutDep.found() and (glmDep.found() or glmAvailable) and glewDep.found()
  buildMapOpenGL=true
else
  buildMapOpenGL=false
endif

And also ftDep to the list of dependencies in libosmscout-opengl/meson.build

I can fix the autoconf build after accepting the autoconf build. If the meson build still breaks afte rabove changes I can cleanup myself, too.

PR #451, opened 2017-08-21
Fixed memory leaks in OpenGLRenderer

fannymonori:
I checked the program with valgrind, and fixed the memory leaks.
Now the leak summary seems ok:
==9832== LEAK SUMMARY:
==9832==    definitely lost: 0 bytes in 0 blocks
==9832==    indirectly lost: 0 bytes in 0 blocks
==9832==      possibly lost: 0 bytes in 0 blocks
==9832==    still reachable: 35,833,399 bytes in 1,259 blocks
==9832==         suppressed: 0 bytes in 0 blocks

### DIFF
--- a/libosmscout-map-opengl/include/osmscout/OpenGLMapData.h
+++ b/libosmscout-map-opengl/include/osmscout/OpenGLMapData.h
@@ -43,7 +43,13 @@ namespace osmscout {
 
     size_t fromOriginY;
 
+    ~OpenGLTexture(){
+      delete []data;
+    }
+
   };
+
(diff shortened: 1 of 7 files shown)


Framstag:
Very good. Note though, that "reachable" might also hint at memory leaks. It might be caused by objects that have not been cleaned up correctly and thus still referenced. Technically there should be no "reachable" at the end of the application, in practice this does not work though, because things like singletons are not deleted. This can be tolerated IMHO, especially if you do not have influence of third party libraries. One should check though, if the memory referenced should be still referenced at the application shutdown.

Nevertheless, doing valgrind and fixing memory leaks is a very good idea, I have not though of  :-)

PR #473, opened 2017-09-07
Add multi-line label rendering for iOS

vyskocil:
The same for OSX will come later


Framstag:
Restarted build because looks like on of the build got stuck (because of travis problems).

PR #474, opened 2017-09-07
Qt refactoring: SearchModule, LookupModule

Karry:
Hi. Here is next step in Qt refactoring when I want to break `DBThread` functionality to smaller specialized classes... Search functionality is moved to new `SearchModule` and evaluating of location description is moved to `LookupModule`.


Framstag:
Restarted build. Seems like Travis has problems with OS X build :-/ They start late or even hang and never start :-/

PR #475, opened 2017-09-09
Add CMake support and production of framework libraries for iOS and Mac OSX

vyskocil:
Here are the command lines I used to build all the libs and apps on a Mac : 

```
cd libosmscout
. ./setupAutoconf.sh
mkdir build
cd build
```

Then for MacOSX target : 

```
cmake .. -DCMAKE_PREFIX_PATH=~/Dev/Qt5.9.1/5.9.1/clang_64
make
```

And for iOS target : 

 ```
cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/iOS.cmake -DMARISA_INCLUDE_DIRS=~/Dev/marisa/lib -DPKG_CONFIG_EXECUTABLE=/usr/local/bin/pkg-config  ..
make
```

I don't know why `-DPKG_CONFIG_EXECUTABLE=...` is needed for iOS and not for Mac...
Don't forget to clean the build directory when switching between iOS and Mac, when you are in build directory do : 

```
rm -rf libosmscout*
rm CMakeCache.txt
```

### CODE REVIEW

**Comment by vyskocil** [comment]
The iOS & Mac OSX app is not yet upgraded to use the framework libs and is not compiled with CMake

**Comment by frogg** [comment]
Hey @vyskocil this looks very cool! I would love to use the .framework version of libosmscout for an iOS App. 

I have some problems with it because I some header files are missing in the OSMScout.framework:

For example, `'osmscout/MapPainterIOS.h' file not found`.
Also, should I drag the `OSMScout.h`, `OSMScout.m`, `OSMScoutMKTileOverlay.h`, `OSMScoutIOSView.h` manually in the project? In my opinion it would make sense to include these headers as well.

And in general, do you know why the compiled framework is ~150MB? That seems quite a lot for a mobile framework. Do you know if there is any way to shrink the size? E.g. by removing unused parts of the framework? Or is there some sample map data included that explains the file size?

**Comment by vyskocil** [comment]
Hi @frogg, I also realized that the headers are missing from the framework and that indeed the sample code compiled because it look for the headers in the sources... I've to work more on this, I already look on it but not finished.
The size might be explained because the frameworks contains also bitcode along the 64bits compiled code, some frameworks are not useful (maybe) on mobile like libosmscoutimport

**Comment by Framstag** [comment]
The import library is something that is definitely not designed to be run on mobile. IO and memory requirements are too high for this.

**Comment by frogg** [comment]
@vyskocil let me know if I can help! I have some iOS experience, if you point me into a direction how to fix the header situation I might be able to create a pull request.

**Comment by frogg** [comment]
I looked into the source code:

Is it these lines in `libosmscout/libosmscout-map-iOSX/CMakeLists.txt` that define which header files go into the .framework?
```
set(HEADER_FILES
    include/osmscout/private/Math.h
    include/osmscout/MapPainterIOS.h
)
```

aquiles2k:
Could you please also as a proof of concept change the OSMScout2 App to build as an app bundle which integrates the OSMScout frameworks? 

Right now i cant get our App (which deploys as app bundle) to use the frameworks and we had to change backk the cmake code to compile .dylibs. 

btw one of the missing things was a missing RPATH entry in the libraries inside the framework.

PR #477, opened 2017-09-12
normalise location name before lookup

Karry:
Hi. 

I added location name normalisation for lookup. Main problem was that no-breakable space is sometimes used in street name but it is missing in `addr:street` in address points. This change brings 1881 new address points* to osmscout in Czechia ;-)

`*` ...ok, it is not big issue, it is less than 1% of address points...

### DIFF
--- a/libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
+++ b/libosmscout-import/src/osmscout/import/GenLocationIndex.cpp
@@ -72,9 +72,9 @@ namespace osmscout {
   void LocationIndexGenerator::PostalArea::AddLocationObject(const std::string& name,
                                                              const ObjectFileRef& objectRef)
   {
-    std::string locationNameLower=UTF8StringToLower(name);
+    std::string locationNameNorm=UTF8NormForLookup(name);
 
-    RegionLocation& location=locations[locationNameLower];
+    RegionLocation& location=locations[locationNameNorm];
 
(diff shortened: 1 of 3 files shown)


Karry:
Question is if we should use this simple normalisation even for strings in location index and fulltext (Marisa) indexes. Because when I try `Demos/LocationLookup`, with pattern "v lesíčku, praha", it gives me no results, but "v lesíčku, praha" (with non-break space), it gives me:
```
Searching for pattern "v lesíčku, praha"
= Location (V Lesíčku) ~ PostalArea (15000) = Smíchov
   * Smíchov/Praha/okres Hlavní město Praha/Hlavní město Praha/Praha/Česko
   - Way 94858669 highway_residential
```
...I can even imagine how I will type non-break space on mobile :-D

Framstag:
Yes, preprocessing of input string and other parts of the index might make sense. Imaging use of multiple space instead of one...

I'm working on a better location lookup code (better implementation in LocationService for existing functionality, especially free form text lookup). The new code uses StringMatcher. The StringMatcher interface allows you to inject custom processing on index values and search strings. It is target for normalization even up to phonetic comparison. Initially it will be used on lookup time, but I could image that it could also be use during index generation time.
It may be (also) an appropriate place to handling "soft" matches.

Karry:
Nice, I can't wait for it :-) When you are working on improvements in location index, I move my energy to sorting search results in Qt model...

PR #483, opened 2017-09-21
SimpleRoutingService.cpp: remove unused variables

waldyrious:
Fixes the unused-but-set-variable warning when compiling, related to minDistanceLat and minDistanceLon


Framstag:
Sorry for the late response. Need 5 minutes free time and concentration to see, what the original author wanted and what exactly the patch fixed :-)

Thanks for offering the fix!

waldyrious:
Perfectly understandable -- in fact that's how it should be :) Cheers!

Issue #378, opened 2017-06-20
OSMScoutOpenGL window creation fails with OpenGL 3.3


Karry:
Hi. 

OSMScoutOpenGL fails on window creation, `glfwCreateWindow` returns NULL.  `glxinfo` on my notebook reports `Max core profile version: 3.3` ([glxinfo full output gist](https://gist.github.com/Karry/6673945119951e43cabb6aee96f67969)) When I used lower hint `GLFW_CONTEXT_VERSION_MAJOR / MINOR` as @fannymonori suggested window is created properly...

If I understand http://www.glfw.org/docs/latest/window_guide.html, these hints should contains lowest required OpenGL (ES) version:

> ...but creation will fail if the OpenGL version of the created context is less than the one requested

Framstag:
That has been fixed, too.

Karry:
Hi. I am still getting startup error with master branch: 
```GLFW Error: GLX: Failed to create context: GLXBadFBConfig```

Framstag:
@Karry: Is this issue still valid?

Karry:
Yes, it is still valid. I create MR that is solving problem for me, but I am not sure if it will work for everyone...

PR #485, opened 2017-09-23
use lowest possible OpenGL version hint

Karry:
Hi. I want to fix this issue: https://github.com/Framstag/libosmscout/issues/378 
But I am not sure if setting lower OGL version hint may have some cons. @fannymonori please review it :-)


Framstag:
Waiting some time for feedback from @fannymonori ...

PR #488, opened 2017-09-26
Qt search improvements

Karry:
Hi. These changes allows to search street without region. search model lookup current town (admin region 8) and use it as default region for search. In OSMScout2 demo is used map center for region lookup...


aquiles2k:
Just want to note that admin level 8 is not always a town. For example Berlin is admin level 4.

PR #504, opened 2017-10-18
Fixed iOS-build specific refactoring oversights

frogg:
This PR fixes two fatal build errors for the iOS build.

My suggestion to prevent this in the future: add the iOS build to travis as well and don't merge PRs that fail to build on iOS.


Framstag:
Regarding iOS builds on Travis:
* Does travis support this?
* If yes: I'm no iOS developer, so I depend on getting pull requests that configure such builds for travis. If I would get such pull request, I would of course pass it through.

frogg:
Yes, TravisCI definitely supports iOS. I'll try to figure something out tomorrow after my first lecture. It should be quite similar to macOS.

PR #505, opened 2017-10-19
Trying to add iOS as build target for TravisCI

frogg:
I am not sure if that already works. I am creating the PR to trigger TravicCI to see if it now builds for iOS as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,13 @@ matrix:
       env:
         - BUILDTOOL=cmake
         - TARGET=build
+    - os: osx
+      osx_image: xcode9.1
+      compiler: clang
+      env:
+        - BUILDTOOL=cmake
+        - TARGET=build
+        - PLATFORM=ios
     - os: linux
       dist: trusty
       sudo: required
--- a/ci/travis/build.sh
+++ b/ci/travis/build.sh
@@ -14,7 +14,7 @@ locale
 echo "Build start time: `date`"
 
 if [ "$TARGET" = "build" ]; then
-  if  [ "$TRAVIS_OS_NAME" = "osx" ]; then
+  if  [ "$TRAVIS_OS_NAME" = "osx" ] && [ "$PLATFORM" != "ios" ]; then
     export PATH="/usr/local/opt/qt/bin:$PATH"
     export PATH="/usr/local/opt/gettext/bin:$PATH"
     export PATH="/usr/local/opt/libxml2/bin:$PATH"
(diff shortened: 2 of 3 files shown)


frogg:
Any idea what this could be?

```
CMake Error at /usr/local/opt/qt/lib/cmake/Qt5Gui/Qt5GuiConfigExtras.cmake:9 (message):
 Failed to find "gl.h" in
"/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk/System/Library/Frameworks/OpenGL.framework/Headers;/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk/System/Library/Frameworks/AGL.framework/Headers".
Call Stack (most recent call first):
 /usr/local/opt/qt/lib/cmake/Qt5Gui/Qt5GuiConfig.cmake:167 (include)
 cmake/features.cmake:110 (find_package)
 CMakeLists.txt:160 (include)
```

Framstag:
I'm no iOS specialist, but nevertheless:
* Does iOS support OpenGL or does it only support OpenGL ES? Is this possibly the reason, that the header is not found?
* I would suggest to create a minimal build for iOS. As long as you do not plan to use Qt on iOS (is this supported at all), the cmake build should only build libraries that are required/specific for iOS development. So I would suggest to drop Qt and build of Qt based code for iOS. This way we can still be sure nothing breaks on iOS, build gets faster and we do not have to find a solution for above problem ;-)

Framstag:
Though I have no local iOS build, I would nevertheless work on this. Since I do not know of a way to add to an external pull request, I plan to copy and modify the code.

frogg:
Thanks, that sounds good. I couldn't figure out how to implement this for iOS myself.

Framstag:
From the output of the cmake iOS build it looks like it is already generating code for iOS arm64. Plan is to simply remove some dependencies (espeically) Qt from the installation/cmake call to just workaround the problem. Since it is unlikely that you will use Qt under iOS if a native renderer is available, dropping the Qt map renderer should be no real problem.

Karry:
Regarding to binaries for arm, it is also necessary to skip tests on Travis ;-) 
https://github.com/Framstag/libosmscout/pull/520/commits/9f5699c53ed1e84a740e8cf3bae0ef48f9e8f6e3

Framstag:
Yes, I know  :-) Time just run out yesterday.

Framstag:
iOS build now seems to work (thanks to @Karry for fixing test execution). Please check, if it is OK for you. I would now close the pull request!?

Framstag:
Closed because parts of the patch were applied manually.

Issue #266, opened 2017-03-03
POI search: merge near objects with same label and type


Karry:
With Qt client code, when POI search is enabled (with Marisa), search may result many duplicates. For example for metro station that is created by many separate points.

![metro-search-duplicates-small](https://cloud.githubusercontent.com/assets/309458/23543306/e10370b2-fff1-11e6-8606-635f16c7380b.png)

I am not sure where merging should be done, there are two possible places: `LocationListModel`, `DBThread`. Model slots are called in UI thread context, some expensive computations may hit usability, but DBThread works with databases independently here...

Framstag:
There is another place? What about filling the TextIndex instead based on the node, area and way data files based on the location index itself? At least I did not see the same effect in the normal location search (though I ask myself, why it would be different)?

So: What does the location index return for the same query? And if the location index has the same problem, how could/would be solve it there? And would filling the text index based on the location thus help?

Karry:
Well, when I test "Florenc, Praha", I get results from location index (I guess). There are duplicates too. There are few `railway_subway_entrance` entries that are just a few meters around station... Such entries can be merged in result view...

Framstag:
Can you give a concrete command line call for location and text index search via the demo tools so I can take a closer look?

Karry:
```
$ wget https://osmscout.karry.cz/europe/czech-republic-11-20170301.tar
$ tar -xf czech-republic-11-20170301.tar 
$ ./Demos/LocationLookup europe-czech-republic "Florenc, Praha" | grep -B 2 highway_bus_stop
= POI (Florenc) = Praha
   * Praha/Česko
   - Node 43231264 highway_bus_stop
--
= POI (Florenc) = Praha
   * Praha/Česko
   - Node 43229708 highway_bus_stop
--
= POI (Florenc) ~ Hlavní město Praha
   * Hlavní město Praha/Praha/Česko
   - Node 43231264 highway_bus_stop
--
= POI (Florenc) ~ Hlavní město Praha
   * Hlavní město Praha/Praha/Česko
   - Node 43229708 highway_bus_stop
--
= POI (Florenc) ~ okres Hlavní město Praha
   * okres Hlavní město Praha/Hlavní město Praha/Praha/Česko
   - Node 43231264 highway_bus_stop
--
= POI (Florenc) ~ okres Hlavní město Praha
   * okres Hlavní město Praha/Hlavní město Praha/Praha/Česko
   - Node 43229708 highway_bus_stop
```

[...]

PR #506, opened 2017-10-22
Qt search result merging and sorting

Karry:
Hi. This MR improving search results with Qt API. Developer may control what locations should be merged and how results should be sorted just by setting two JavaScript functions. 
```
LocationListModel {
        compare: function(a, b){ return -1; }
        equals: function(a, b){ return false;  }
}
```

...I was trying to wrote some "smart" logic in C++ first, but it will not fit to all usecases. This solutions is more flexible from my point of view.

### DIFF
--- a/OSMScout2/qml/custom/LocationSearch.qml
+++ b/OSMScout2/qml/custom/LocationSearch.qml
@@ -238,11 +238,15 @@ LineEdit {
             } else if (loc.type=="object"){
                 var rank=1;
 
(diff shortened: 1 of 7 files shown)


Karry:
It will also fix my earlier issue with multiple near POIs with same type: https://github.com/Framstag/libosmscout/issues/266

Framstag:
@Karry: Regarding your last comment: Should I already merge or wait?

Karry:
You may merge it. I just waned to add link to issue.

PR #519, opened 2017-11-23
Various DirectX backends enhancements

laarmen:
We're currently using libosmscout as part of one of our product, and we've had to patch a bit the DirectX map backend. Here are those patches, rebased onto master.

There are a few bug fixes, and an implementation of the contour string rendering in order to display the street names.

Note that since those patches predate the ContourLabelHelper class, they don't use it just yet.

The version we use in production is available on our master at https://github.com/ScriptAndGo/libosmscout . Be warned : the history there is a terrible mess :)


laarmen:
Rebased, there was a typo that made the build fail. Should be fixed.

Karry:
Hi Simon. I have no experience with DirectX api and I don't have configured development stack on my Windows virtual machine. Can you post screenshots before and after for me please? I am just curious ;-)

laarmen:
Fair enough, although to be honest that's not a trivial thing you're asking me : all the DBs I have on my computer ATM are for our forked version, which is stuck at the previous DB format revision.

I'll give it a try tomorrow, while looking into this latest MSYS build failure.

Framstag:
Problems with min/max have most of the time one or more of the following as cause:
* Not all required headers imported (C and C++ offer implementations)
* Datatypes are not equal for left and right hand side
* Missing std:: prfix for version that are only implemented in C++

Should be easy to fix. Will you take a look?

Framstag:
See http://en.cppreference.com/w/cpp/algorithm/max

PR #520, opened 2017-11-26
Gpx lib

Karry:
Hi.

This MR adds new `libosmscout-gpx` library with basic support for gpx file import / export and manipulating. There is just one demo that use it - `GpxPipe`. I will prepare integration to `client-qt` library later.

There is support just for `cmake` and `meson` build systems.


Framstag:
Looks like you have problems with availability of date/time stuff under non-Linux OS (Windows, OS X). Chrono (http://www.cplusplus.com/reference/chrono/) should be available on all OSs, is the required functionality available there?

Karry:
...there are few non-standard [`std::tm`](http://en.cppreference.com/w/cpp/chrono/c/tm) fields that seems to be missing on other systems. Try to remove its initialisation, `std::tm time{};` should setup all fields to zero anyway...

PR #521, opened 2017-11-30
fix canvas tranformations in Qt PlaneMapRenderer

Karry:
Here are fixes for problems discussed on mailing list. 

 - `final-image` center is computed correctly (don't use bounding box center)
 - transformation `target rectangle` keeps image ratio

Adrian, please review and check these changes with your code.


aquiles2k:
Yes that fixes both issues for me. Thanks!

PR #524, opened 2017-12-04
notify tile callbacks everytime when tile data is marked as completed

Karry:
When low-zoom data file don't contains data for current zoom level,
tile data is marked as completed without notify tile callbacks.
In rare situations (related with high load probably), when low-zoom
tasks are executed as the last one for current tile (it may be reproduced
by thread sleep), tile is setup as completed without notify callback.
As a result, Qt/DBLoadJob is newer completed and rendering pipeline stops.

This change should solve it.

### DIFF
--- a/libosmscout-map/include/osmscout/DataTileCache.h
+++ b/libosmscout-map/include/osmscout/DataTileCache.h
@@ -78,7 +78,7 @@ namespace osmscout {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Framstag:
Wow, fine analysis and bad bug by me :-(

Karry:
...don't worry, it happends :-)

Issue #405, opened 2017-07-10
Is there any simple/built-in way to draw custom points/paths on map? [$50]


hexonxons:
I want to draw paths and points with bearing over qt map, but I couldn't find any api to do that.
Right now I'm drawing it using `QPainter` in custom `MapWidget` overloaded class, but there are a lot of different calculations to handle map scrolling/zooming. It works, but I wonder if there are any public apis to draw points and paths by its lat/lon coordinates with already implemented zoom/scroll calculations.
![2017-07-10-121340_3840x1080_scrot](https://user-images.githubusercontent.com/663637/28011273-72e187a2-656a-11e7-871f-ede6f93e153c.png)

Karry:
Hi. Such api don't exists now. We don't support even route rendering, I removed it when I was adding support for rendering multiple databases... But it would be great to have it! Map rendering library without custom objects on the map is not so useful :-/

Question is how do it
 - extend (tiled, plane) `MapRenderer` and render custom objects with `MapPainterQt`. Style of custom objects may be customised by map stylesheet, object labels may be taken into account during label placing...
 - create some abstract overlay renderer that will be "injected" into `OSMScoutQt` and called from `MapWidget` after `MapRenderer`? It may be useful for custom heatmap, hill shading...

I plan to work on it, but I have no schedule. It may takes me months or years... Merge requests are always welcome ;-)

Framstag:
The problem Lukas mentiones is a problem of the libopsmscout-qt-client library. This library implements many feature above the basic libosmscout libraries for QWt based backends. It also allows drawing local maps and can downloads tiles, too.

The various render backends do allow to inject custom objects using the poiXXX attributes in the MapData structure. In this case the poiXXX attributes will not be touched by the MapService and they will be rendered by the backend as normal objects.

See http://libosmscout.sourceforge.net/tutorials/renderonmap/ for more details.

This above trick is not support my libomscout-qt-client yet, because you cannot simply use the normal Qt backend if you want to render on top of downloaded tiles.

It should be still possible though but would require some additional handling. You can configure the Qt backend to do not draw any backgrounds. So mentioned overlay can possible simulated by an additional Qt MapPainter with matching configuration.

Framstag:
@Karry: Can this now get closed?

Framstag:
[...]

PR #525, opened 2017-12-05
Qml map overlay (tiled)

Karry:
Hi.

This MR creates new possibility how to render on on top of QML map widget. It was requested 5 months ago in https://github.com/Framstag/libosmscout/issues/405 (uff, time is fast). It is not necessary to modify `MapWidget`, just create new component that is child of  `MapOverlay`, overlap Map component with it and connect `view` property.

First implementation of such overlay is `TiledMapOverlay` that allows to add some online tile source as overlay. One nice example is hill shading:

```qml
  Map{
    id: map

            TiledMapOverlay {
                anchors.fill: parent
                view: map.view
                enabled: true
                opacity: 0.5
                // If you intend to use tiles from OpenMapSurfer services in your own applications please contact us.
                // https://korona.geog.uni-heidelberg.de/contact.html
                provider: {
                      "id": "ASTER_GDEM",
                      "name": "Hillshade",
                      "servers": [
                        "https://korona.geog.uni-heidelberg.de/tiles/asterh/x=%2&y=%3&z=%1"
                      ],
                      "maximumZoomLevel": 18,
                      "copyright": "© IAT, METI, NASA, NOAA",
                    }
            }
  }
```
result:
hillshading overlay

### DIFF
--- a/OSMScout2/qml/main.qml
+++ b/OSMScout2/qml/main.qml
@@ -136,6 +136,23 @@ Window {
                 }
             }
 
+            TiledMapOverlay {
+                anchors.fill: parent
+                view: map.view
+                opacity: 0.5
+                // If you intend to use tiles from OpenMapSurfer services in your own applications please contact us.
+                // https://korona.geog.uni-heidelberg.de/contact.html
+                provider: {
+                      "id": "ASTER_GDEM",
+                      "name": "Hillshade",
+                      "servers": [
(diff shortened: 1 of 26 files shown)


Framstag:
That screenshot looks impressive :-)

Karry:
Agree. Hill shades looks great on the map. It would be interesting to render it offline, just with usage of contour lines... How dense contours should be to get usable shadings? How fast rendering will be? 

I am still postponing this idea, because I have more practical things on my virtual todo list.

PR #529, opened 2017-12-10
fix StyleEditor signal connection

Karry:
fix for Qt warning:
```
QObject::connect: No such signal DBThread::InitialisationFinished(DatabaseLoadedResponse) in libosmscout/StyleEditor/src/MainWindow.cpp:32
```

### DIFF
--- a/StyleEditor/src/MainWindow.cpp
+++ b/StyleEditor/src/MainWindow.cpp
@@ -29,9 +29,9 @@ MainWindow::MainWindow(DBThreadRef dbThread)
    dbThread(dbThread)
 {
   connect(dbThread.get(),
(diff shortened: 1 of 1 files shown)


Framstag:
```
ninja: error: '../OSMScout2/OSMScout2/pics/Download.svg', needed by 'OSMScout2/qt5-res_qrc.cpp', missing and no known rule to make it
```

Looks like this error might be introduced by the update to meson 0.44. The build script always uses the most current version :-/

meson 0.44 did not yet arrive on my system (Arch), so I cannot fix it quickly...

Will merge the pull request if this is the only build error (which is likely with this small change).

Framstag:
See https://github.com/mesonbuild/meson/issues/2763

Karry:
Thank you for creating the issue :-)

PR #531, opened 2017-12-13
Add search breaker (SearchForLocationByString)

Karry:
Hi, when I am searching some place on mobile phone, location lookup took long time sometimes. Combined with search suggestions, it may happen that background search thread processing lookup of "Pr" when user already typed complete world "Prague". This simple change adds breaker to search parameter and setup breaker instance that already exists in Qt `SearchModule`. It boost search experience ;-)


Framstag:
This is a good idea. I'm however not sure, if this is enough.  I have to look st your cod ein detail, but I assume that you do not break within visitor calls. So individual traversals of the index will still not get interrupted.

PR #537, opened 2017-12-19
Plane Qt renderer: round target rectangle possition and size to avoid extra antialisign

Karry:
This patch was initialised by @aquiles2k , it round position of rendered `finishedImage` to avoid additional antialiasign. Difference is small, but visible.


Framstag:
Just for my understanding (while waiting for the build to complete): The problem is/was, that the maps as rendered by the MapPainter is copied on screen at non-aligned pixel coordinates and then aliasing effects of the used backend (Qt) did come into action?

Karry:
Exactly. `targetTopLeftX` (and `targetTopLeftY`) are not aligned to pixel, it has value -100.0001 for example. And such small un-aligment cause additional anti-aliasign on Qt component side.

PR #543, opened 2017-12-28
Prepare for navigation

Karry:
Hi. 

This is initial step for navigation support in Qt and its improvements in core library. This is not finished yet, but I don't want to merge it in one big merge request (this is even so huge).

The main change is new "navigation simulator" inside OSMScout2 app. It reads GPX file and simulate position source from its routes... Just try it:
```
./OSMScout2/OSMScout2 --simulate-track test-track.gpx ~/Maps/ ../stylesheets/standard.oss
```
You can download [my test track](http://osmscout.karry.cz/tests/test-track.gpx).

Integrating this simulator to existing OSMScout2 application was simplest way, but we may separate it to new application. What do you think?

obrazek


Framstag:
CMake Windows build breaks.

Framstag:
Looks like either the cmake file is missing files or import/export macros are wrong?

Framstag:
Since adding a new application has some setup cost, I would keep the simulator in the OSMScout2 demo. This way we can also share some possible improvements in the GUI etc.. It make sense to keep the code separate so at leats we can move it later on if there is demand. General OSMScout2 strategy: Move as much features that belong in a real navigation application into it, as long as the GUI can handle it. If we have too many buttons so we do not see the map anymore we should either think of a new GUI concept or start moving things out.

Framstag:
Btw., cool feature :-)

Karry:
Agree, it would be fine to make OSMScout2 fully usable on devices with QtQuick components... But I am using it just like a test app now, so I dont give so much attention to usability.

I will take a look on windows build next year :-)

PR #547, opened 2018-01-11
Route step icons

Karry:
Another small step for improving routing and navigation - expose route step type to qml and display icon for given type. Icons are copied from [poor maps](https://github.com/otsaloma/poor-maps) by @otsaloma that should be licensed as GPL...

obrazek


otsaloma:
The icons are originally from Mapbox, CC0-licensed.

https://github.com/otsaloma/poor-maps/blob/master/qml/icons/navigation/README.md

PR #549, opened 2018-01-20
Qt: Translatable route step description 

Karry:
Hi all. 

Here is possibility to translate route steps with Qt API (finally). I was thinking about two options - transfer all route step properties to qml and make translation there, or translate phrases in C++ code and provide already translated descriptions to QML. I selected second and simpler way (but less customisable for library users).

There are two variants of description - short that is simple plaintext and full that contains simple formatting...


Framstag:
Looks like there is a problem with detecting the translation module you are now using? Does it need another package install?

Karry:
You are correct, there is need for `qttools5-dev-tools` package on Debian/Ubuntu. Beside it, Meson usage of `qt5.compile_translations` was incorrect.

There is remaining TODO in build - what directories should be used for translation files on Windows and Mac OS...?

Framstag:
Good question, I do not know anything about translations in Qt, so it is difficult to give a recommendation :-/ What do the Qt tools itself do under these platforms?

Karry:
I don't find any suggested solution how to install translation. Best solution from my point of view is to install generated *.qm files to subdirectory inside platform specific data location `QStandardPaths::AppDataLocation` (default path on Linux is /usr/local/share/).

I added specific paths for Linux/Mac/Windows to CMake and Meson in recent commits. I tested it only on Linux. Do you have possibility to check install directory with the rest of platforms? OSMScout2 should find translations after install... 

Anyway, we should review all installed files - we are installing all demos to default binary path, stylesheets to `<prefix>/share/`... Or we may say that libosmscout builds should not be used for install and it is responsibility of "distributors" to deliver all files to correct locations.

Framstag:
I have never done any install for libosmscout :-/ The initial cmake files written by me (and also the initial meson files) did not contain any installation code. This was added by the community. The problem with installing is, that you have to do manually cleanup if things go wrong...so I hesitate to test it :-/

And you are right, we need an installation concept. IMHO the libraries and headers should be installed anyway. Likely also OSMScout2 and StyleEditor. I also suggest to install the Import and DumpData tools, though here already name conflicts are likely. Name clashes are even more likely with the demo applications. Since I assume that different distribution have different concepts for this, we possible should not install them at ll?

For Linux I also think, building a flatpack package, a docker image or similar might be a wiser effort than spending time in the install (besides the basic tools)?

Btw,, the performance of Travis seems to get worse. Mac OS builds are slow (they throttled open source projects recently). This is not helpful. I'm open to suggestions (like moving parts of the builds somewhere else for Mac OS?).

Framstag:
I updated master since the last push from you. Can you update to pull request (builds seems to be green now) so I can (automatically) merge?

Karry:
Ok, I just do rebase. You can merge it, beside iOS build that should be fixed now, there are failing two windows builds - one with Meson bug and second (cmake/msvc) caused by translation list probably...

Issue #294, opened 2017-04-07
Better rendering of multiple databases


Karry:
With current Qt client api map databases are painted to canvas one by one. Implementation is here: https://github.com/Framstag/libosmscout/blob/master/libosmscout-client-qt/src/osmscout/PlaneDBThread.cpp#L441

There can appears some errors on database boundaries. For example ways or labels covered by areas:

![database-boundary](https://cloud.githubusercontent.com/assets/309458/24820554/f8ab4208-1be9-11e7-9499-e7c853148afc.png)

How to improve this situation? I have two ideas:
 - rewrite `StyleConfig`, `MapPainter` to be independent on database (`TypeConfig`), merge data from all databases before paint and then use one `MapPainter`
 - expose painting steps from `MapPainter` (prepare, ground tiles, areas, ways, labels...) as public api and paint map by "layers" with multiple MapPainters...

Framstag:
StyleConfig, MapPainter are dependend on the database, because they need to interpret the FeatureValueBuffer based on concrete type information offered by TypeConfig (which in turn is offered by the database). Joining data this way would enforce all databases being imported using the identical map.ost. Currently you can still mix versions.

Merging data from all databases is not easy, because the is no simple definition of identity. Objects are (likely) equal because they have the same length of nodes and (all) nodes have the same id (coordinate and serial id).

The first question to ask: Do databases overlap? If they overlap are there always regions, that contain equal data by definition? If yes, we do not need to merge data, we just need to render data in a way that we render (sub tiles) with data from either one database. This would possibly the simpler solution.

And if this is not the case: Can we possibly change the import to fix this? If not, we are back to your suggestions...

rinigus:
Just few comments from my side. 

First, regarding the overlapping of the databases. Yes, databases of the neighbouring countries overlap. This is due to the new common practice of splitting the planet that way that it keeps the relationships intact. So, for example, if one node is in Germany on the border, the whole line/polygon relationship will be incorporated into Germany PBF as well as into the other country. So, some data will overlap, but not all.

When we discuss rendering of the multiple databases, I would like to consider https://github.com/Framstag/libosmscout/issues/57 as well as https://github.com/Framstag/libosmscout/issues/287 . 

[...]

PR #552, opened 2018-01-27
create abstract util MapPainterBatch and its Qt implementation

Karry:
It allows to render map consisting from multiple databases
step by step using recently added ability to draw just
range of RenderSteps. It draw map "layer" by all database painters
at once, so it helps to solve this issue: https://github.com/Framstag/libosmscout/issues/294

### DIFF
--- a/libosmscout-client-qt/src/osmscout/MapRenderer.cpp
+++ b/libosmscout-client-qt/src/osmscout/MapRenderer.cpp
@@ -271,7 +271,8 @@ void DBRenderJob::Run(const osmscout::BasemapDatabaseRef& basemapDatabase,
     }
   }
 
-  // draw databases
+  // prepare data for batch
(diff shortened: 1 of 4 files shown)


Framstag:
msys/cmake build now fails, too. Looks like a forward definition is missing? See https://ci.appveyor.com/project/Framstag/libosmscout/build/1459/job/5oa3lu20tp094af5

Framstag:
Still fails, the misisng symbol is:

```
__imp_vtable for osmscout::MapPainterBatchQt
```

I can not see what the cause it, but can reproduce it locally...I testing afew things. If you have an idea...

Framstag:
I got it to work:
* Move method implementation (constructor, destructor) of MapPainterBatchQt form header to source
* Remove the virtual of the MappainetrBatchQt destructor (should already be covered by base class).

Issue #555, opened 2018-02-04
areanode.idx database opening don't take care of GetAreasDataMMap()


vyskocil:
The Open code pass a hard-coded true value when opening the scanner : 

> bool AreaNodeIndex::Open(const std::string& path)
>  {
>     datafilename=AppendFileToDir(path,AREA_NODE_IDX);
>     try {
>       scanner.Open(datafilename,FileScanner::FastRandom,true);
> ...
> 

In my App I'm getting crashes from time to time with these logs : 
`
Cannot mmap file '/var/mobile/Containers/Data/Application/AEC9EBF6-B01B-412B-A68C-570561FB8B46/Documents/France.osmscout/areanode.idx' of size 12152060 (Cannot allocate memory)
MyFreemap(7867,0x1b6231b80) malloc: *** mach_vm_map(size=16007168) failed (error code=3)
*** error: can't allocate region
`

Karry:
Well, `DatabaseParameter` control just access (mmap/seek) to data files, not index files. Indexes are always mmaped. What parameters you are using? You can try to use similar mechanism as in Qt code: [DBThread.cpp:214](https://github.com/Framstag/libosmscout/blob/master/libosmscout-client-qt/src/osmscout/DBThread.cpp#L214). For 32 bit environment, where mmap may cause out-of-memory states, is computed summary size of all databases, and if this size is bigger than 1.5 GiB (experimental safe limit for mmap), there is setup slow seek mode for some data files...

vyskocil:
This issue is happening on a iPhone with 64 bit iOS code, it seems that even being 64 bit it has  severe limitations on virtual memory that can be mmap'ed...
I'm already disabling mmap'ing for all the database files where it is possible but as you point it out, actually, indexes couldn't. Before I was manually disabling the whole mmap support in config files and didn't had this issue but I try to not have to hack in the osmscout code and use "standard" compiled frameworks :-)

Framstag:
IMHO the right solution is to make this configurable like it is done for other files. As always, bugs are appreciated else I might do it myself, as soon as I have some time.

Framstag:
Closing the issue since code has been supplied :-) Thanks @vyskocil for giving the hint and @Karry for the patch!

PR #559, opened 2018-02-10
optionally disable memory mapped index files

Karry:
It should helps to solve problems with memory mapped files on iOS... https://github.com/Framstag/libosmscout/issues/555


Framstag:
OK, I have done part of this locally, too, but you changed even more so I through away my local changes :-) I admit I should have told that I started with the fix :-(

Framstag:
Waiting for the build to finish...

Karry:
Sorry for duplicate work. I hope that you dont invest too mutch time to it.

PR #562, opened 2018-02-15
Style editor improvements

Karry:
Hi. Here is another batch of improvements for style editor - mainly line numbers and block indentation by tab/shift+tab keys. 

I am still surprised how terrible support for text editing is with Qt Quick! I just wanted to create small map style and ended up with two days of StyleEditor hacking ;-)


vyskocil:
You are very welcome for these days of StyleEditor hacking ! Thanks you @Karry :-)

PR #566, opened 2018-02-21
allow to route by foot on cycleway

Karry:
When I was experimenting with router in terrain, it plans the foot route by detour. But there was a shorter alternative by cycleway. It is allowed to go on cycleway in Czechia, but when I am looking to [OSM wiki](https://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restrictions) it is forbidden in many countries. 

Should we allow it in router, or make it more complicated and create separated type (when permit is explicit)? How to handle country specific defaults? ...it is just minor problem, I don't want to create any country specific table ;-)

### DIFF
--- a/stylesheets/map.ost
+++ b/stylesheets/map.ost
@@ -345,7 +345,7 @@ TYPES
   TYPE highway_cycleway
     = WAY ("highway"=="cycleway")
       {Name, NameAlt}
-      PATH[BICYCLE] PIN_WAY
+      PATH[FOOT BICYCLE] PIN_WAY
       DESC
         en: "cycleway"
         de: "Fahradweg"


Framstag:
As long as we have imports that only cover one country, it possibly easiest way would be to pass corrections via command line.

Ideally that imported data already holds data that is complete and does not have any implicit location specific rules (this means that all access would be either explicit or there is a common default for access for the whole world for e specific type). I though fear that that is not the view of the common mapper  :-)

A lookup table per country is not a complete solution since the import itself does not know to which country the data does belong. For this we would need to evaluate borders. Most imports though will not contain complete countries. Things will be even more difficult, if the import contains multiple (but only partial) countries.

So passing "local" overwrites via command line would be a solution that is rather easy to implement and will result in a good enough solution for most cases. It is then up to the person that creates the import to choose the right "default" for the specific import.

Framstag:
I'll pass the merge, but we have to see if somebody urges to revert it.

PR #574, opened 2018-03-07
Initial version of tiled routing

Framstag:
RoutingDB is now using a new RoutingDataFile class, which uses a tiled data segments approach.


libosmscout/src/osmscout/routing/RouteNodeDataFile.cpp
```diff
@@ -0,0 +1,181 @@
...
+      FileOffset indexFileOffset;
+      uint32_t   dataCount;
+      uint32_t   indexEntryCount;
+      uint32_t   tileMag;
+
+      scanner.Open(datafilename,
+                   FileScanner::LowMemRandom,
+                   memoryMappedData);
+
+      scanner.Read(indexFileOffset);
```
Karry: current router.dat (database format 16) don't contains any index right? This is causing crash with segfault with already existing databases!

libosmscout/src/osmscout/routing/RouteNodeDataFile.cpp
```diff
@@ -0,0 +1,181 @@
...
+      tileCalculator=TileCalculator(std::pow(2,tileMag));
+
+      scanner.SetPos(indexFileOffset);
+      scanner.Read(indexEntryCount);
+
+      for (size_t i=1; i<=indexEntryCount; i++) {
+        Pixel      cell;
+        IndexEntry entry;
+
+        scanner.Read(cell.x);
```
Karry: isn't this duplication of coverage.idx ?

libosmscout/src/osmscout/routing/RouteNodeDataFile.cpp
Framstag: Right, I have to increase the data version.

libosmscout/src/osmscout/routing/RouteNodeDataFile.cpp
Framstag: Yes and no. Both indexes can offer the same information: "Is the given tile covered by the database". But IMHO they still differ: The coverage.idx should be much smaller and faster. Since it maps its information onto a single bit (in memory). The index that is part of the router data file, holds information where tile data is places and how much entries it has. This means: The index is bigger in memory and lookup is slower (not bitmap lookup (O(1)) but a map (O(logn)).

If I need coverage for map, I wold possibly use coverage.idx, and the route data file only if I actively do routing.

The coverage.idx is pure optional, so it is up to the actual client, which index it wants to use.

Karry:
Thanks to clarify it. I will do some measurements on my arm devices at monday. I am away from my crosscompilation environment right now ;-)

Framstag:
OK, I'm eager to hear the results. If you measure, please try to find out, what exactly is slower (or consumes more memory), so I can take a look at possible optimizations.

Karry:
It will takes more time that I expected, I made few mistakes in build for Raspberry :-/ I got similar numbers for x86 like you, but I am more curious for multi-database routing performance. But `MultiDBRouting` test seems to be broken :-/ Can you take a look on it please? You will be faster with it ;-) Problem seems to be in `MultiDBRoutingService::ReadCellsForRoutingTree` method, that should not be needed now, right?

Framstag:
I'll see, what I can find...give me some days though.

Framstag:
In my test case I get the following output:

```
./maps/koeln-regbez ../maps/arnsberg-regbez 51.5717798 7.4587852  50.6890143 7.1360549
Opening database 1...
TypeConfig::TypeConfig()
Opening TypeConfig: 0.002
Done.
Opening database 2...
TypeConfig::TypeConfig()
Opening TypeConfig: 0.002
Done.
Opening router...
Done.
Opening AreaAreaIndex: 0.000
Opening AreaWayIndex: 0.000
Opening AreaDataFile: 0.000
Opening WayDataFile: 0.000
Opening AreaAreaIndex: 0.000
Opening AreaWayIndex: 0.000
Opening AreaDataFile: 0.000
Opening WayDataFile: 0.000
Opening routing file '../maps/arnsberg-regbez/router.dat'
Error while reading '../maps/arnsberg-regbez/router.dat': File '../maps/arnsberg-regbez/router.dat' - Cannot read int16_t: Numerical result out of range
Automatically closing FileScanner for file '../maps/arnsberg-regbez/router.dat'!
Can't find common routing nodes for databases ../maps/arnsberg-regbez, ../maps/koeln-regbez
Route failed
Database::~Database()
TypeConfig::~TypeConfig()
Database::~Database()
TypeConfig::~TypeConfig()
Deinitializing Logging
```

Is the error similar to your error?

Framstag:
```
  bool MultiDBRoutingService::ReadCellsForRoutingTree(osmscout::Database& database,
                                                      std::unordered_set<uint64_t>& cells)
```
...reads the router.dat directly. Since the file format has changed, the code is (of course) broken. I missed to fix and test that one :-/

Framstag:
Same for 

```
  bool MultiDBRoutingService::ReadRouteNodesForCells(osmscout::Database& database,
                                                     std::unordered_set<uint64_t>& cells,
                                                     std::unordered_set<Id>& routeNodes)
```

I have fixed these methods, but of course using the new RouteNodeDataFile it should be (much?) simpler and faster to gain the requested information.

Karry:
Here are initial results from my measurements: https://docs.google.com/spreadsheets/d/1j9__FoikvTDeXgwrQUTKNgUS_9ItrCdxOVy_4gl1ziU/edit?usp=sharing
It seems that routing performance is more limited by storage speed than cpu performance - Odroid C2 is the most powerful arm device what I have, but routing performance is terrible.
Anyway, I think that you may merge this branch, performance penalty is tolerable. We should adapt multi-database routing to use new index and measure multi-database routing then...

Framstag:
That makes sense, since we are always loading all data for a tile. If we are in luck, we need the data anyway, but most of the time not all RouteNodes of a tile are visited. So in result the overall data loaded is more. I hoped that bulk loading all data for a tile makes things faster than permanent random access, but it does not look like (in fact bulk instead of random access should only be fatser on hard disk, random access has no penalty on SSD or similar).
Obe possible optimization though: Since the number of data in a tile is much smaller than the whole data file, we can load the raw data in memory and parse it from there. Similar to memory mapped file handling. In fact you can use memory mapped file access on parts of the file AFAIK. I would hope that this will make disk access faster (for this case). This is something worth to invest in.

Framstag:
One question regarding performance: Did you use memory mapped files on the Odroid C2? If yes, optimizations to use selective memory mapped data would not change anything, because memory mapped files would already be active in this case.

Framstag:
Hmm, what about reading the individual tiles only up to the node we are currently interested in (we need some more state in the tile index). This might reduce the amount of data actually read?

Framstag:
The latest push measurably improved routing time (but it is still slower than the old approach). I see the following further alternatives:
* Asynchronous preloading (is likely in conflict with the last push)
* Improved data format, because of loading data in block  (I'm not sure how big the potential is, is in conflict to the last push)

Karry:
...sorry, I was in skiing recent days, I hope that I will have more time for coding this week.

> Did you use memory mapped files on the Odroid C2?

Yes, I was using Demos/Routing executable that uses default database parameter, that enable memory mapping... I will provide updated results (that includes your recent changes) later...

PR #576, opened 2018-03-11
Add agg and opengl to PerformanceTest and a new DrawMapOpenGL 

Hyspirit:
The opengl backend has the OSMScoutOpenGL, but I wanted to compare the resulting images with the same parameters quickly, so here is a new DrawMapOpenGL.

Also, the PerformanceTest was mising Agg and OpenGL, so here they are.
Both CMake and meson build work, but I couldn't manage to run without error the executables made with meson that use OpenGL (DrawMapOpenGL and PerformanceTest called with opengl).


Framstag:
This is an very interesting pull request. Thank you!

While the performance test will not tell the whole truth (regarding overall (subjective) application performance) it would still be a very helpful thing to be able to directly compare most rendering backends.

If you are lost regarding the remaining build errors please tell so explicitly. I'll then try to take a look and see if I can fix the remaining errors. Before that I'll wait :-)

Hyspirit:
I never worked with CMake and meson before, so it took me some time but I think I fixed the build (appveyor is still incomplete at this point).

I didn't merge back your new commit into my branch, so I don't have the wercker file, but I believe it should work when you merge everything.

PR #579, opened 2018-03-22
use new tiled index in multi database router

Karry:
Hi. These changes in `MultiDBRoutingService` should make possible (not tested yet) to route via multiple (not just two) databases! It even improves performance, routing time from Prague to Berlin took 1:40 with master branch, but just 12 seconds with this update (on my x86 notebook).

I will provide more measurements on later...

### DIFF
--- a/Tests/src/MultiDBRouting.cpp
+++ b/Tests/src/MultiDBRouting.cpp
@@ -32,11 +32,10 @@
 
 struct Arguments
 {
-  bool                   help=false;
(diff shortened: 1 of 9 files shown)


Framstag:
This requires my tiled index code to be merged before?

Karry:
I create this merge request to your `tiled_routing_db` branch, not `master`. 

Anyway, I think that you may merge both to master ;-)

PR #584, opened 2018-03-26
DirectX: Add a guard against null styles for areas

laarmen:
Apparently sometimes some areas don't have fill styles. I know at least
in one instance it's a bug since I can see a fill color in the default
style (which I used), but until I figure out why exactly this happens,
and because the other renderers all seem to do the same, I'm patching the
symptom before fixing the underlying cause.

For the record, the problem is when drawing an area in Rennes,France which has
the type "tourism_attraction_building". I'll open an issue if I can't find out what's going on.


laarmen:
OK, turns out it is normal for fill styles to be null for areas when drawing the inner rings. My fix is even more valid then :)

PR #602, opened 2018-04-28
Several Targets Not Installed on Ubuntu16

adasta:
There seems to be errors in several cmake lists.  They refer to a variable THE_TARGET_NAME which has not been set.  As a result, their targets are not installed. This variable has been replaced  with the actual target name.  Now these targets are properly installed on Ubuntu16


Karry:
LGTM. Nice catch :-)

Framstag:
One build failed. Looks like an appveyor problem. Restarting build. If OK, I will merge! Thanks for the patch!

PR #612, opened 2018-06-03
add brand as a tag for name and namealt

rinigus:
Some nodes don't have names, but use brand as a corresponding tag [https://wiki.openstreetmap.org/wiki/Key%3Abrand]. This commit brings "brand" as the third choice for Name and NameAlt


rinigus:
Tim, please hold on with it - there is some strange interaction when I specify Alt language during import. I'll look into it and then will update the request. Will let you know when its ready

rinigus:
I made a mistake while copy-and-pasting the registration calls. Now it seems to be fine.

PR #613, opened 2018-06-04
fix symbol rendering, update rectangle primitive documentation to correspond with code

Karry:
Hi. 

Rectangle primitive for symbols uses `topLeft` as coordinate, but its `GetBoundingBox` method don't respect it. So I updated this method and documentation to be in sync with code. Symbols in stylesheets had to be updated too.

Another fix in `MapPainterQt::DrawSymbol` move symbol horizontally to put its center to given point.

I tested these changes with Qt backend, it is possible that symbol rendering with other backends is broken now!

### DIFF
--- a/Demos/src/DrawMapAgg.cpp
+++ b/Demos/src/DrawMapAgg.cpp
@@ -109,6 +109,7 @@ int main(int argc, char* argv[])
 
   if (!styleConfig->Load(style)) {
     std::cerr << "Cannot open style" << std::endl;
+    return 1;
   }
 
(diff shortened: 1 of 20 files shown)


Karry:
I tested Cairo and Agg renderers and both are broken... Please wait with merging, I will try to update other renderers (at least Agg and Cairo).

Karry:
Now all backend should render symbols the same way (except svg that don't support it).

I found another bug related with symbols, Qt map painter double rectangle radius. I updated all circles in stylesheets. Other backend don't copy this bug...

PR #626, opened 2018-07-07
drawmaposx demo application

vyskocil:


Framstag:
The iOS build is broken, will you take a look?

vyskocil:
I excluded the build of DrawMapOSX on iOS

Framstag:
That makes sense!

PR #628, opened 2018-07-25
Ubuntu 18.04 docker image

Karry:
Regarding the plan, [support for Ubuntu 17.10 ends](https://wiki.ubuntu.com/Releases). So, I replaced docker image of Ubuntu 17.10, by latest Ubuntu 18.04...


Framstag:
Merging. I do not know why the meson msys build fails. Regarding wercker: Do you mind to take a look at it? You Qt knowledge is better. Problem occurred after I switch to a more recent Arch image with a newer meson version.

Karry:
It seems that rcc tool is executed in bad directory and relative paths are not resolved properly. We may workaround it with symlink of qml directory to build directory... I will try it.

PR #629, opened 2018-07-26
WIP: add wercker CI workaround for Qt resources

Karry:


Karry:
It is strange. I am able to reproduce wercker build error with archlinux_gcc_meson docker image, but when I start docker with `--privileged` option (I just wanted to run rcc tool with strace), it works magically...?!

Framstag:
I do not think so. The docker image is build by docker hub itself. We could build our own bu would have to push it ourselfs. The execution of the docker image (the docker daemon) is is done by wercker. I do not see anywhere in the documentation, that you can influence the docker daemon version and I have to admit, I would have been surprised. So we can only wait for wecker or rcc to update?

Karry:
Or you may try to use debian/meson image... It works for me without privileged container execution. In other hands, it fail with older kernels (3.10.0, on my server). So it have to be tried...

PR #635, opened 2018-08-11
fix DirectX demo on x64 Windows

Karry:
long has 32 bites on x64 Windows (msvc), but pointers 64 bites.
So converting pointer to long discard upper bites.
Use LONG_PTR to avoid this issue.

### DIFF
--- a/Demos/src/DrawMapDirectX.cpp
+++ b/Demos/src/DrawMapDirectX.cpp
@@ -396,7 +396,7 @@ private:
 			::SetWindowLongPtrW(
 				hWnd,
 				GWLP_USERDATA,
-				PtrToUlong(pDemoApp)
+				(LONG_PTR)pDemoApp
 				);
 
 			result = 1;


Framstag:
Hmm, I do have Windows 64 and VS 2017, too, and did not have these problems with the cmake based build.

Note that I currently try to add additional dependencies using vcpkg, avoid manual installatin and builds of dependencies. See https://ci.appveyor.com/project/Framstag/libosmscout/build/job/tjdjb15k0q6cdd6f

Currently it look like I'm close, I just would have to make positioning module optional or have to check if the module is supplied by the Qt5 vcpkg packages, too..

Karry:
And are you using x64 build? This demo crash every time on my virtual machine...

Framstag:
I compile for x64., I only did build the software, not run it. I assume you had *build* problems.

Anyway, the vcpkg build now works. I have to check, if the resulting work, too.

PR #637, opened 2018-08-14
Karry label rendering

Framstag:
Initial work and integration newer Layouter code in the directx backend. Normal labels work, positioning seems still to be wrong (likely some (cenetr?) offsetting relative to label origin). Contour labels and shields still missing.


Framstag:
OK, this as good as I get it without any in deep knowledge of DirectX. I would have to spend much more time to get nicer.

At least labels are now visible again (while partly broken) and the integration into the new layouter is at least done that improvements should now be easier.

@Karry Please merge
@laarmen Please find some time to fix my bugs ;-)

Framstag:
If any of you have further questions, just ask :-)

Karry:
Thank you Tim for initial DirectX support. I am merging it, I did not found anything wrong. Contour labels may be finished later...

PR #638, opened 2018-08-16
make possible to setup name feature for overlay objects

Karry:
...it is useful when you want to display named waypoint on the map


Framstag:
What does it exactly do? Can you attach a screenshot (sounds like this would explain it)?

Waiting for the builds to finish...

Karry:
you can have this simple QML / JavaScript code for adding overlay point:
```js
var wpt = wptPreviewMap.createOverlayNode("_waypoint");
wpt.addPoint(model.latitude, model.longitude);
wpt.name = "test";
wptPreviewMap.addOverlayObject(0, wpt);
```
such overlay object has various benefits:
 * is styled usual way
 * its symbol and label are handled by label layouter
 * it is rendered by background renderer thread by design, so you can add hundreds without adding UI lack

test_waypoint

PR #641, opened 2018-08-23
add support for label and icon padding

Karry:
Add support for label and icon padding (again). I just removed `sameLabelSpace` option, because it would be hard to implement with current layouter.

### DIFF
--- a/libosmscout-map-agg/src/osmscout/MapPainterAgg.cpp
+++ b/libosmscout-map-agg/src/osmscout/MapPainterAgg.cpp
@@ -422,7 +422,7 @@ namespace osmscout {
                                  const MapParameter& parameter,
                                  const MapData& /*data*/)
   {
-    labelLayouter.Layout();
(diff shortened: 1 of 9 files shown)


Framstag:
Hmm, this seam to create two different build errors? Though I did not see a change that would have triggered that!?

Karry:
It seems that Qt translation files was removed by accident. There is some cmake / lupdate bug that delete them sometimes...

I will restore it...

Framstag:
Thanks!

Karry:
Missing translations files seems not to be the issue. But broken Qt installation in AppVeyor / msys2 instance...

```
C:/msys64/mingw64/bin/lupdate.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory
```

Framstag:
Hmmm, I rebuild everything. Lets see if it persists. The Qt5 package itself is the same as always.

Framstag:
Push it anyway...

PR #642, opened 2018-08-23
Qt: reset painter's pen before drawing way cap

Karry:
Pen may be setup from previous steps and it affect way cap then. This simple patch fix it.

obrazek


Framstag:
Simple fix, merge without waiting for build result

PR #643, opened 2018-08-24
webpage: add table of backend features

Karry:
Hi. I added missing backends (DirectX, MacOS / iOS) to "rendering" page and replace simple table with feature matrix. I hope that values are correct, I wrote it from my memory...

obrazek


Framstag:
Thanks! That should be really helpful. I found the feature-side of the documentation not "good" as it does not really precisely describes what is possible,  but this change is definitely a step in the right direction :-)

PR #646, opened 2018-08-26
add icon size to map parameter, tell requested icon size to painter

Karry:
* add support icons rescaling to Qt map painter
* set default icon size to 3.7 mm (~14px with 96 DPI)

Hi @Framstag , @vyskocil and others. Before merge this MR, we should decide how to handle icon size. Label layouter now assumes that icon size is 14 mm (not px), but Qt, Cairo and Mac render icon with dimension of source image (14px with default icons), iOS renderer use upscaling by `contentScale`.

14 pixels is too small for hi-DPI displays and there is no possibility to change it... So I added icon size to render arguments (with default value 3.7 mm = 14px @ 96 DPI) and use this value in layouter and DrawIcon method then...

### DIFF
--- a/Android/OsmScoutLib/jni/include/jniMapPainterCanvas.h
+++ b/Android/OsmScoutLib/jni/include/jniMapPainterCanvas.h
@@ -77,6 +77,7 @@ namespace osmscout {
   protected:
 
     bool HasIcon(const StyleConfig& styleConfig,
+                 const Projection& projection,
                  const MapParameter& parameter,
                  IconStyle& style);
 
--- a/Android/OsmScoutLib/jni/src/jniMapPainterCanvas.cpp
+++ b/Android/OsmScoutLib/jni/src/jniMapPainterCanvas.cpp
@@ -51,6 +51,7 @@ namespace osmscout {
   }
   
   bool MapPainterCanvas::HasIcon(const StyleConfig& styleConfig,
+                                 const Projection& /*projection*/,
                                  const MapParameter& parameter,
                                  IconStyle& style)
   {
--- a/Demos/src/DrawMapCairo.cpp
+++ b/Demos/src/DrawMapCairo.cpp
@@ -114,6 +114,13 @@ int main(int argc, char* argv[])
       drawParameter.SetLabelLineFitToArea(true);
       drawParameter.SetLabelLineFitToWidth(std::min(projection.GetWidth(), projection.GetHeight()));
 
+      /*
+      std::list<std::string> paths;
+      paths.push_back("./libosmscout/data/icons/14x14/standard/");
+      drawParameter.SetIconMode(osmscout::MapParameter::FixedSizePixmap);
+      drawParameter.SetIconPaths(paths);
+      */
+
       projection.Set(osmscout::GeoCoord(lat,lon),
                      osmscout::Magnification(zoom),
                      DPI,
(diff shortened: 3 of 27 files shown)


Framstag:
The original idea was as follows:
* Have SVG icons
* Render png icons based on the SVG icons in multiple sizes
* Choose during map render initialisation which pixel-based icon size you want

The actual code implementation though was fixed and a hard coded size of icons was expected though I have to admit.

The reason for that initial plan:
* Loading and rendering of png is simple in most backends.
* Rendering SVG is not
* SVG though allows good scaling of images for various sizes. There is even a (not tested) way to show or less icon detailed based in the target pixel size.
* Upscaling of bitmap images has it costs and it not visibly perfect.

I would still allows this approach. So perhaps the layouter should handled icons in pixel sizes. I of course do not have a problem with individual backend allowing to up(or down-)scale icons to the requested pixel size.

Btw., plate labels also have hardcoded pixel sizes for border AFAIK a least as seen in cairo and directx backend. That should be fixed, too.

Karry:
I agree that upscaled png images looks ugly. But fixed size 14px is too small for current displays. So what to do next? Qt and Cairo should be able to use svg icons directly, I believe that iOS/Mac backend will have api for it too. There is still option to use png as fallback... Should I try to use SVG icons?

What do you mean with plate labels?

laarmen:
https://github.com/Microsoft/Windows-universal-samples/tree/master/Samples/D2DSvgImage
this tells me it should be possible in the directx backend as well.

Le mer. 29 août 2018 à 11:32, Lukáš Karas <notifications@github.com> a
écrit :

Framstag:
OK,

I would *suggest* the following:
* Layouter and MapPainter should be able to correctly render icons of any size. So the support for just 14x14pixel icons should be lifted.
* So we need a configuration value that states the expected size for the images on disk in the given directory
* Actual metrics though should be based on pixel to be able to support backends  that do not support SVGs and to support potential users, that nevertheless would like to use bitmap icons of a given size.
* We though check, that it is easy for the user to calculate pixel size based on dpi and screen size in mm. AFAIK the helper functions for this are there. So if the user decides to automatically scale icon size based on screen metrics, this should be easy.
* I could image to a enum for multiple modes and configuration values for all of them.
* I still assume the independent on SVG or bitmap icons, I make sense to only icons with a full pixel count. Nobody want to scale and render an image to 14,3x14,3 pixels - resulting in strange aliasing effects - or worse.
* Backends that support SVG are free to to additionally support this (why not!). I assume that in this case an implicit conversion from SVG to Bitmap for faster rendering would be done. Such conversion would be likely done on first use of an icon (if you have 30 icons you would not convert them on start). Possibly the backend would cache them even on disk  - which in turn could mean, that the conversion is a application manually triggered function and the backend itself still only knows the bitmap icons on disk?
* Backends could trigger SVG or PNG mode by file postfixes, if conversion is not outside the render.
* I like modular code, so movng conversion outside the render makes things for flexible.

Does this help?

Framstag:
@Karry: Plate labels: The shield label for ref names of the way (normally the alphanumerical motorway number).

Karry:
Hi Tim. I agree with you. Lets split this hi-dpi task to multiple steps (merge requests). In this MR I just defined icon mode in `MapParameter`:
```
    enum IconMode
    {
      FixedSizePixmap,  // !< raster icons should be used, iconPixelSize will be used for rendering
      ScaledPixmap,     // !< raster icons should be used, icons will be scaled to iconSize
      Scalable          // !< vector icons should be used, icons will be scaled to iconSize
    };
```
Default values should not change current behavior - `FixedSizePixmap` mode with pixel size 14 is used. I added SVG icon support to Qt map painter and pixmap scalling to DirectX and Cairo...

Please review it and if you like it, you may merge it ;-)

PR #649, opened 2018-09-04
add support for gpx file metadata

Karry:
Support for importing and exporting of gpx file metadata - name, description and time...


Framstag:
Rebuild triggered for appveyor.

PR #652, opened 2018-09-14
Svg painter icons

Karry:
...just for fun ;-)

in output svg it looks like that:
```
    <defs>
        <image id="icon_1" with="21.8504" height="21.8504" xlink:href="data:image/svg+xml;base64,..." />
    </defs>

   <use x="100.993" y="252.021" width="21" height="21" xlink:href="#icon_1" />
```


Framstag:
I assumed that the SVG backend is mostly broken? Is it in better state than I assumed? Neverthless I'm happy for every patch :-) Thanks!

Karry:
No, it is not broken. Output from SVG painter looks fine. It is just necessary to open in some good viewer - Inkscape or Firefox. In comparison with Qt or Cairo backend, just area petterns are not supported now.

PR #655, opened 2018-09-16
Implement contour label drawing

vyskocil:


vyskocil:
Thanks @Karry, but you did a great work on this :-)
Yes all the features should be ok now, the are still some improvement that could be made like multi-line labels (it was there with the old layouter), optimizations like checking intersection with the viewport when drawing labels,...
And there are allocated data (line runs) in Layout that should be freed be I don't know when...

vyskocil:
The sizes of the fonts should also be tweaked to take into account the screen DPI of the device

vyskocil:
@Karry CTLineRef are reference to CoreFoundation objects that need to be manually ref counted. I found and used a smart pointer for CF objects lib that solve the releasing issue.

PR #659, opened 2018-09-23
add Ubuntu1804 image to appveyor CI

Karry:
...it should be working now :-) 

Building with debug code is broken now (`-DLABEL_LAYOUTER_DEBUG=ON -DDEBUG_GROUNDTILES=ON -DDEBUG_COASTLINE=ON -DDEBUG_TILING=ON -DDEBUG_ROUTING=ON`), because it is not tested. I will fix it later...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,11 +4,13 @@ branches:
   only:
     - master
 
-os: Visual Studio 2015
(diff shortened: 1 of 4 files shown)


Karry:
Here is fix for debug code: https://github.com/Framstag/libosmscout/pull/660

PR #663, opened 2018-10-02
run unittests on Appveyor with GCC sanitizers

Karry:
three sanitizers are used now - address, thread, undefined behavior

In C++ is easy to overlook some undefined behavior or invalid memory access. It may leads to invalid results that may appears just with some compiler or just on some architectures. These errors are hard to find. For that reason, it is useful to execute unittests with sanitizers that may discover this kind of bugs...


Karry:
It seems that thread sanitiser has many false positives :-(

PR #667, opened 2018-10-18
Fix Segmentation fault

alex-dranoel:
Segmentation fault  on line 211 when width * height < 8. It occurs to me today while playing with tiny osm files.


Karry:
LGTM

PR #668, opened 2018-10-18
Fix segmentation fault

alex-dranoel:
Segmentation fault at line 251:
nodeTypeData[type->GetNodeId()]
when nodeTypeData is empty.

It happened to me while playing with tiny osm files.

### DIFF
--- a/libosmscout/src/osmscout/AreaNodeIndex.cpp
+++ b/libosmscout/src/osmscout/AreaNodeIndex.cpp
@@ -232,6 +232,10 @@ namespace osmscout {
                                  std::vector<FileOffset>& offsets,
                                  TypeInfoSet& loadedTypes) const
   {
+    if (nodeTypeData.empty()) {
+      return true;
+    }
+    
     StopClock time;
 
     loadedTypes.Clear();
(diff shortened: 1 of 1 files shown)


Framstag:
It look like the error is bigger. The code also fails (and still fails after your patch) if the requested type is bigger than the nodeTypeData area. It looks like it would be better to guard the call to the area and check before if type->GetNodeId() < nodeTypeData.size()

Agree? Can you modify your patch?

alex-dranoel:
I have a small zip file containing tiny data, that I imported using the Import tool, that leads to the crash. If you extract it and run one of the Demo script with it, like DrawMapCairo, you should be able to reproduce it.
[rail_data.zip](https://github.com/Framstag/libosmscout/files/2496787/rail_data.zip)

alex-dranoel:
And here is a stack trace obtained using valgrind. Search for GetOffsets to get quickly to the segfault.
[valgrind_stack_trace.txt](https://github.com/Framstag/libosmscout/files/2496830/valgrind_stack_trace.txt)

Framstag:
The call stack looks OK to me. Seems like there are node types that never occur in the actual data and thus not in the index.

Karry:
After fast look to `AreaNodeIndexGenerator`, it seems to be valid case that `nodeTypeData` contains just a subset of available node types, when input data don't contains these types. So your current fix makes perfectly sense...

Issue #55, opened 2016-06-25
PoC: Add initial support for node specific render data caching in MapPainterQt.


Framstag:
To evaluate the effect of caching of render specific data in regard to render performance and memory consumption an initial proof of concept should be done:

* Add another Draw method that either gets concrete tiles passed or the MapService (so it can request tiles on demand itself). 
* Implement a cache, that allows assignment of one or one per object type structs to a tile id in the cache.
* Implement a simple API to allow a specific renderer to extend the data stored for an object type by its own custom information.
* Create one or multiple data objects for caching as much as possible information for a node. Such information could be: One or multiple resolved label and symbol styles including label data, dimensions and positions.

In the first step Information should be stored projection on on-screen position independent. Before drawing only the following steps should be required to be done:
* Transform geo location of the node to an on screen coordinate
* Evaluate the label in relation to other labels resulting in "visible" or "not visible"

In a second step one could already evaluate visibility based on the other labels within the same tile.

Framstag:
@Karry What is your opinion, is this really necessary? It would cost additional memory and IMHO the renderer is currently fast enough?

Karry:
All primitive transformations, label layouting, sorting of areas (...) is done again and again for every rendering. There is place for improvements definitely, rendering may take more than second in dense areas. But the bad UX may be compensated by canvas transformations (or tile transformations) as implemented in Qt MapWidget. So I don't see this as priority right now.

I guess that OpenGL painter may help here and question is if we should invest energy to improve it, instead of optimizing software renderers...?

It is up to you, if you want to keep this issue open as an idea for improvement...

PR #669, opened 2018-11-06
Bounding box cache for Area::Ring and Way

Karry:
Hi Tim. 

I want to return `boundary_country` to stylesheets. Before doing this, I just wanted to know how mutch and why rendering of this type is expensive. So I create this simple stylesheet:

```
OSS
  CONST
    COLOR waterColor                 = #9acffd;
    COLOR landColor                  = #f1eee9;
    COLOR countryBorderColor         = #ff0000;

  STYLE
    [TYPE _tile_sea] AREA {color: @waterColor;}
    [TYPE _tile_land] AREA {color: @landColor;}

    [TYPE boundary_country] WAY {color: @countryBorderColor; displayWidth: 0.4mm; dash: 7,3;}
    [TYPE boundary_country] AREA.BORDER {color: @countryBorderColor; width: 0.4mm; dash: 7,3;}
END
```

...and started experimenting with `PeformanceTest` demo:
```
./Demos/PerformanceTest \
  ~/Maps/europe-czech-republic-20180823-024442/ \
  boundaries-test.oss \
  50.0918 14.402 50.0918 14.402 \
  8 18 \
  1200 1200 \
  Qt
```

`boundary_country` type for Czech Republic has just one area with 87 127 nodes. I updated performance test to repeat map rendering 100x. Without any other modifications, rendering of zoom level 16 took **19 ms** in average (my x86 notebook).

```
Level: 16
Tiles: 100
 Used memory: max: 24.9778 MiB avg: 255.772 KiB
 Tot. data  : nodes: 0 way: 0 areas: 1
 Avg. data  : nodes: 0 way: 0 areas: 0
 DB         : total: 88.0596 min: 88.0596 avg: 0.880596 max: 88.0596 
 Map        : total: 1913.49 min: 18.031 avg: 19.1349 max: 34.7981
```

From perf-top is clear that computation of bounding box is expensive for such huge area. So I pre-compute it for `Area::Ring`. With this change, rendering took **15 ms**.
```
Level: 16
Tiles: 100
 Used memory: max: 24.9794 MiB avg: 255.789 KiB
 Tot. data  : nodes: 0 way: 0 areas: 1
 Avg. data  : nodes: 0 way: 0 areas: 0
 DB         : total: 6.50177 min: 6.50177 avg: 0.0650177 max: 6.50177 
 Map        : total: 1541.2 min: 13.841 avg: 15.412 max: 28.865
```

For such huge areas like `boundary_country`, it is clear that it is not complete visible for higher zoom levels. When just a small segment is intersecting with the projection viewport. It is not necessary to convert all points to screen coordinates and render it. Segments outside viewport can be simplified to boost performance. So my second modification was to pre-compute segment (1000 nodes) bounding boxes and replace complete segment with line when it is outside viewport... With this modification, rendering took **4.9 ms**

```
Level: 16
Tiles: 100
 Used memory: max: 21.0395 MiB avg: 215.445 KiB
 Tot. data  : nodes: 0 way: 0 areas: 1
 Avg. data  : nodes: 0 way: 0 areas: 0
 DB         : total: 99.8611 min: 99.8611 avg: 0.998611 max: 99.8611 
 Map        : total: 497.905 min: 4.55919 avg: 4.97905 max: 7.57603
```

When both optimisations are used, rendering took **0.9 ms**!
```
Level: 16
Tiles: 100
 Used memory: max: 21.0395 MiB avg: 215.445 KiB
 Tot. data  : nodes: 0 way: 0 areas: 1
 Avg. data  : nodes: 0 way: 0 areas: 0
 DB         : total: 9.7306 min: 9.7306 avg: 0.097306 max: 9.7306 
 Map        : total: 92.1714 min: 0.605772 avg: 0.921714 max: 3.41345
```

---

I think that it is clear that it makes sense to pre-compute such values for huge areas (and ways?) and reduce rendered nodes outside projection viewport. For these areas, it is huge change that it will be rendered multiple times during its live in cache. Question is how to do it properly? Store pre-computed values in `Ring` directly as in this POC? Use new separate cache just for rendering https://github.com/Framstag/libosmscout/issues/55 ? How such cache should looks like?

### DIFF
--- a/Demos/src/DrawMapDirectX.cpp
+++ b/Demos/src/DrawMapDirectX.cpp
@@ -28,6 +28,7 @@ src/DrawMapDirectX ../maps/nordrhein-westfalen ../stylesheets/standard.oss 7.465
 
 #include <SDKDDKVer.h>
 #define WIN32_LEAN_AND_MEAN
+#define NOMINMAX // msvc issue with std::max/min
 #include <windows.h>
 #include <tchar.h>
 #include <cmath>
--- a/Tests/src/FileScannerWriter.cpp
+++ b/Tests/src/FileScannerWriter.cpp
@@ -187,273 +187,276 @@ int main()
 
     writer.Close();
 
-    scanner.Open("test.dat",osmscout::FileScanner::Normal,false);
-
-    // Read/Write
-
-    scanner.Read(inBool);
-    if (inBool!=outBool1) {
-      std::cerr << "Read/Write(bool): Expected " << outBool1 << ", got " << inBool << std::endl;
-      errors++;
-    }
-
-    scanner.Read(inBool);
-    if (inBool!=outBool2) {
-      std::cerr << "Read/Write(bool): Expected " << outBool2 << ", got " << inBool << std::endl;
-      errors++;
-    }
-
-    scanner.Read(in16u);
-    if (in16u!=out16u1) {
-      std::cerr << "Read/Write(uint16_t): Expected " << out16u1 << ", got " << in16u << std::endl;
-      errors++;
-    }
-
-    scanner.Read(in16u);
-    if (in16u!=out16u2) {
-      std::cerr << "Read/Write(uint16_t): Expected " << out16u2 << ", got " << in16u << std::endl;
-      errors++;
-    }
-
-    scanner.Read(in16u);
-    if (in16u!=out16u3) {
-      std::cerr << "Read/Write(uint16_t): Expected " << out16u3 << ", got " << in16u << std::endl;
-      errors++;
-    }
-
-    scanner.Read(in32u);
-    if (in32u!=out32u1) {
-      std::cerr << "Read/Write(uint32_t): Expected " << out32u1 << ", got " << in32u << std::endl;
-      errors++;
-    }
-
-    scanner.Read(in32u);
-    if (in32u!=out32u2) {
-      std::cerr << "Read/Write(uint32_t): Expected " << out32u2 << ", got " << in32u << std::endl;
-      errors++;
-    }
-
-    scanner.Read(in32u);
-    if (in32u!=out32u3) {
-      std::cerr << "Read/Write(uint32_t): Expected " << out32u3 << ", got " << in32u << std::endl;
-      errors++;
-    }
-
-    scanner.Read(in64u);
-    if (in64u!=out64u1) {
-      std::cerr << "Read/Write(uint64_t): Expected " << out64u1 << ", got " << in64u << std::endl;
-      errors++;
-    }
-
-    scanner.Read(in64u);
-    if (in64u!=out64u2) {
-      std::cerr << "Read/Write(uint64_t): Expected " << out64u2 << ", got " << in64u << std::endl;
-      errors++;
-    }
-
-    scanner.Read(in64u);
-    if (in64u!=out64u3) {
-      std::cerr << "Read/Write(uint64_t): Expected " << out64u3 << ", got " << in64u << std::endl;
-      errors++;
-    }
-
-    // Read/WriteNumber
-
-    scanner.ReadNumber(in16u);
-    if (in16u!=out16u1) {
-      std::cerr << "Read/WriteNumber(uint16_t): Expected " << out16u1 << ", got " << in16u << std::endl;
-      errors++;
-    }
-
-    scanner.ReadNumber(in16u);
-    if (in16u!=out16u2) {
-      std::cerr << "Read/WriteNumber(uint16_t): Expected " << out16u2 << ", got " << in16u << std::endl;
-      errors++;
-    }
-
-    scanner.ReadNumber(in16u);
-    if (in16u!=out16u3) {
-      std::cerr << "Read/WriteNumber(uint16_t): Expected " << out16u3 << ", got " << in16u << std::endl;
-      errors++;
-    }
-
-    scanner.ReadNumber(in32u);
-    if (in32u!=out32u1) {
-      std::cerr << "Read/WriteNumber(uint32_t): Expected " << out32u1 << ", got " << in32u << std::endl;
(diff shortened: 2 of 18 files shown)


Framstag:
(Not yet looked at actual patch): Yes previous performance findings by me showed: Large objects with many nodes are epensive for precalculation, rendering and (not mentioned by you) also loading.

Up to now I had two optimization strategies:
* For low zoom levels calculate a special version of objects (of types that tend to create large objects and are likely to be rendered in low zoom) to reduce disk size and number of nodes. This worked very well.
* For evne larger object (water, land) I splitted the objects in smaller objects (see water index). This also reduced disk IO (and possible reduces the problem that internal CPU/memory bus might not be fast enough to transfer the data) and nodes size and enable to make index very selective.

Your solution seems to tackle the problem by optimizing calculations and clippings, but keep the original object. I'm not sure if you handle IO problems this way? I would suggest to test this by doing a performance test on the device to check you it behaves with flash memory and CPU/Memory architecture of a mobile. 

If IO gets a problem one would need to make smaller equivalent objects from the original object like water index did (but hpefuly simpler).

If IO is not the problem:
* I would try to avoid caches, since memory is a constraint and  getting cache handling right can be tricky.
* The file format alrady uses magical signal bits to signal variants in the file format. We could add another bit to signal that a precalculated bounding box is available.
* Alternative one could put the bounding box into a internal featureValue, this way you reuse the existing feature bit mask mechanismn to tunnel additional data.
* The question though is, how to handle this in the actual C++ object?  You would need to variants of the area data type, one with bounding box caculation, one with bounding box as attribute. You would ned to find a solution that does not add additional memory (address pointer => 8 byte, std::vector => 16 byte, making Area virtual => ???)  requirements for all instances.

Framstag:
Ah, I assume that you precompute (in the final solution) the bounding box during import?

Karry:
Hi. 

> Ah, I assume that you precompute (in the final solution) the bounding box during import?

No, I don't think that adding more information to disk format is necessary. I think that bounding box (+segment boxes) should be computed during loading of nodes from disk. When IO is bottleneck, extra computation during load should not affect performance much. 

But yes, I need to measure it :-) I am not focusing to IO performance right now, because these large areas like country boundaries, lakes and river banks should be loaded just once and live in memory caches for long time.

> see water index

I don't want to add too much complexity. It seems that these rendering optimisations will be enough.

> For low zoom levels calculate a special version of objects ...

Yes, data files for low zoom helps a lot. It should be enabled for `boundary_*` types when we will want to render it. Problem that I am focusing is the fact that `country_boundary` is prepared and rendered completely even when you are viewing high zoom levels (>10) in central area of the country where no boundary line is not visible. And for country border areas, are usually visible just fraction of points (up to thousand).

> I would try to avoid caches

Agree. I thing that these bounding boxes should live inside `Ring` and `Way` objects and be prepared during data loading. Do you agree?

---

my next steps:
 - Verify my findings on mobile
 - Test loading times with map on the SD card
 - Prepare final variant of the patch

Framstag:
> > Ah, I assume that you precompute (in the final solution) the bounding box during import?

> No, I don't think that adding more information to disk format is necessary. I think that bounding box (+segment boxes) should be computed during loading of nodes from disk. When IO is bottleneck, extra computation during load should not affect performance much.

OK. might be that the difference can be neglected. And yes, calculation during loading would be simpler then.

> But yes, I need to measure it :-) I am not focusing to IO performance right now, because these large areas like country boundaries, lakes and river banks should be loaded just once and live in memory caches for long time.

> > see water index
> 
> I don't want to add too much complexity. It seems that these rendering optimisations will be enough.

OK. That can be measured. We will see. Important is speed on the actual device. Desktop measurements cannot be trusted but for differential speed tests. Water index would definitely a much complex thing and it would be nice if it could be avoided. I would be surprised if it is good enough since large objects had an measurable impact in the past. But we have done a number of optimizations since then. It there any reasonable worst case scenario? Some areas on the world with many borders or where the border node count is dramatically higher? 

> Agree. I thing that these bounding boxes should live inside `Ring` and `Way` objects and be prepared during data loading. Do you agree?

Right.

> my next steps:
> 
>     * Verify my findings on mobile
> 
>     * Test loading times with map on the SD card
> 
>     * Prepare final variant of the patch

Sounds reasonable.

vyskocil:
This development would also be very useful for iso-elevation contour lines, they might be very long in  almost flat locations

Karry:
Hi.

I am satisfied with patches now. Performance tests on my phone confirms that pre-computing bounding boxes is really useful for huge objects and combined with optimised rendering using "segments" speedup its rendering 20x. This speedup was measured with stylesheet that contains just `boundary_country` type. When `standard.oss` stylesheet was used, rendering was around 3 ms worse (4%) for zoom levels <= 10 (optimised data for low zoom is used) but **59 ms** better (10%) for zoom levels >= 11.

There is some extra computation during data load of course. Its cost is in average 25 ms for zoom levels >= 9. But data should be loaded just once and hold in the caches in usual use case. There are problematic zoom levels 15 and 16 where perfomance impact was 155,25 and 135,48 ms. As discovered later, it was caused by small (way) cache, where single load request don't fit to it and these ways had to be loaded multiple times (for each tile). So performance impact of extra computation was multiplied. When cache size was increased, results was mutch better...

Another cost of these patches is increased code complexity. Please consider it Tim.

Here table with the numbers:
https://docs.google.com/spreadsheets/d/19Ygj_MxAD7Jv6J1qL8JpowZiXts-CDIdXqlfpHeXSks/edit?usp=sharing

Framstag:
Excellent work and well prepared. Only potential problem I see, we do have a memory penalty for each way and area ring (vector<SegementBoundingBox> + BoundingBox>). Did you see any significant memory increase?

I'll pass it anyway, since if it works on Sailfish memory footprint should not be a problem.

PR #677, opened 2018-12-02
Crash router windows

DerKleinePunk:


laarmen:
Please, rebase this PR before @impaktor has a hear attack!

Framstag:
Yes, the patch looks a little bit suspicious. I have applied the fix manually and thus will close the patch without applying it.

PR #682, opened 2018-12-12
fix warning from cppcheck - condition with always false 

Karry:
cppcheck message:

Logical conjunction always evaluates to false: `factor < 0.0 && factor > 1.0`


laarmen:
To be fair, you're not just fixing a warning, but also a real bug :-)

Issue #683, opened 2018-12-18
Search for Cairo Lib


DerKleinePunk:
## In windows it is need that all libs build with debug or no debug.

```
if(CAIRO_FOUND)
  # set it back as false
  set(CAIRO_FOUND FALSE)

  find_library(CAIRO_LIBRARY_RELEASE 
			   NAMES cairo
               HINTS ${_pc_cairo_LIBRARY_DIRS}
  )
find_library(CAIRO_LIBRARY_DEBUG 
			   NAMES cairod
               HINTS ${_pc_cairo_LIBRARY_DIRS}
  )

  if(NOT CAIRO_LIBRARY_DEBUG AND CAIRO_LIBRARY_RELEASE)
    set(CAIRO_LIBRARY_DEBUG ${CAIRO_LIBRARY_RELEASE})
  endif()

  set(CAIRO_LIBRARIES 
		debug ${CAIRO_LIBRARY_DEBUG}
		optimized ${CAIRO_LIBRARY_RELEASE})

  find_path(CAIRO_INCLUDE_DIR cairo.h
            HINTS ${_pc_cairo_INCLUDE_DIRS}
            PATH_SUFFIXES cairo
  )
  set(CAIRO_INCLUDE_DIRS "${CAIRO_INCLUDE_DIR}")

  include(FindPackageHandleStandardArgs)
  find_package_handle_standard_args(Cairo DEFAULT_MSG CAIRO_LIBRARIES CAIRO_INCLUDE_DIRS)
endif(CAIRO_FOUND)
```

Framstag:
Hi @DerKleinePunk: Which change do you exactly request? AFAIK you try to differenciate between release and debug builds!? Is this a common cmake pattern? Can you offer an pull request (I know you are busy)? This way we could see, if you change request would break anything?

DerKleinePunk:
in the vcpkg projekt its the Standard generate an lib with an optional d as filename to signal it is an debug version. CMake is also add an d on filename wen i build it with vcpkg and vs2017. The problem is vs add some addtional code when you compile in Debug mode. New and STL Container (many more) different in Debug an Release. VS use this to detect error and memoryleaks. 
The problem i have is when i use the cairo map painter and cairo for some addtional paintings there to versions of Cairo loaded. And than the Application crashed with heap corruption.

I hope you can understand this

Framstag:
Supplied patch applied, closing issue.

PR #684, opened 2018-12-21
Search debug cairo

DerKleinePunk:
#683 
I hope it hepls


Framstag:
The history again look rather strange, but I'll apply the patch content by hand as soon as I have some time. Looks like the patch doe snot break anything.

Framstag:
@DerKleinePunk: No, sorry. I would suggest to start with a fresh checkout of the master branch (of course before with saving/backup your local changes...). It looks like your actual delta is just a modified .gitignore file.

Framstag:
Applied patch locally and pushed it.

PR #689, opened 2019-01-03
Optimise TileId

Karry:
Following mailing list discussion, it seems that most expensive part of node index loading is building of std::map in TypeData struct.  I created micro-benchmark that open node index multiple times (100x, France DB), build it with -O2 for Arm7 and tried multiple modifications:

```
current master:                                1203.87 ms
inline compare and equal operators of TileId:  1073.74 ms
replace std::map with std::unordered_map :     1441.61 ms
```

It is clear that inlining helps a lot...


Framstag:
Interesting, that HashMap is more expensive in this case than the Map implementation.

PR #700, opened 2019-02-11
disable Client/Qt library on Wercker CI

Karry:
Current Qt tools is not possible to run with old Docker.
See the issue: https://github.com/docker/for-linux/issues/208
We need to temporary disable building of Client/Qt library
until Oracle update its Docker in Wercker service.

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -5,7 +5,7 @@ build:
   steps:
     - script:
         name: Configure project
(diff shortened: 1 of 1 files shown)


Karry:
ping @Framstag

Framstag:
Sorry, I waited for the build to finished and than missed this :-/

PR #701, opened 2019-02-18
Navigation2

Karry:
Next batch of improvements to navigation. 

 - properly remove passed route steps (navigation instructions) in `RouteInstructionAgent`
 - expose following route steps in navigation model (Qt)
 - skip informative route steps from "next navigation instruction"
 - small improvements and cleanup of navigation model

### DIFF
--- a/OSMScout2/qml/NavigationSimulation.qml
+++ b/OSMScout2/qml/NavigationSimulation.qml
@@ -151,8 +151,8 @@ Window {
         id: topContainer
         anchors.left: parent.left
         anchors.top: parent.top
-        width: Math.min(400, parent.width - rightContainer.width)
-        height: 120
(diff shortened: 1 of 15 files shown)


Framstag:
OK. How do things work out? Looks like you at least should have a prototype. Does the new "agent" design help? How good is it intergratable in Qt? Is the code as good to handle as I hoped?

Karry:
Hi. New api works fine :-) It is easily extendable and is easy to keep agents small, just with limited responsibility. Yes, Qt navigation model/module now provides basic functionality that behaves better than previous solution. There is need to provide integration with map view and voice instructions - it will be almost perfect then ;-)

But it is still the beginning with the navigation. It will be necessary to test it in real situations, tune agent parameters, route instructions...

PR #702, opened 2019-02-19
upgrade Travis images from Ubuntu 14.04 (Trusty) to 16.04 (Xenial)

Karry:


Karry:
We are not able to build Import tool statically on Ubuntu Xenial, I am keeping Trusty image meanwhile...

PR #705, opened 2019-02-25
replace NULL macro with nullptr keyword

Karry:
From C++11 `nullptr` keyword is preferred for null pointers instead of `NULL` macro. We are mixing both in library. It is not problem until it is not argument of template, but I think that we should use just one... 

I skip iOS/Mac code and OST parser...


Karry:
msys2 build crashed with internal error, but it seems to be not relevant with these changes...

PR #706, opened 2019-03-05
Navigation: arrival estimate

Karry:
add navigation agent that computes remaining distance and estimated arrival


Framstag:
I have canceled some intermediate builds...

PR #709, opened 2019-03-08
try to fix Travis CI build of Cairo backend on Mac

Karry:


Karry:
`cairo.h` include seems to be fixed, but now it is complaining about `<pango/pangocairo.h>` :-(

Framstag:
I'll take a look...

Framstag:
Looks like you need a
```
export PKG_CONFIG_PATH="/usr/local/opt/libffi/lib/pkgconfig
```

before a call to 
```
meson debug
```

internally pkg-config is called to get the -I options. Seems like it returns nothing without above export. Looks like something in the brew packages did not change for good.

Framstag:
Do you want to improve the patch or should I try to fix it and commit it?

Karry:
Great catch. It helps!

Framstag:
For this I have a Mac Book Pro. With the same installation on a local device is was simple to found the cause :-)

Framstag:
Thanks for fixing the first part!

Karry:
Little bit expensive testing device :-) You are welcome!

PR #729, opened 2019-04-23
Fix area node index

Karry:
Hi Tim. 

I found out that many towns in Czechia disappear from the map. When I was investigating the issue, found a bug in area node index - It expects that tile contains node coordinates in some conditions. But these coordinates was never written to file. As a result, file scanner read garbage. Fix was simple, just write coordinates (in `GenAreaNodeIndex`) when `storeGeoCoord` flag is set. 

To found these bugs early, I added additional validity checks for GeoCoord and boolean to file scanner to make sure that it don't read total garbage from position where is stored different data type. I hope that these checks don't affect performance too much. I made it similar to standard asserts, just disable it when NDEBUB macro is defined to get maximum performance. Or do you have better idea howto check sanity of read data? It would be great to have unittests for all data files that would check compatibility of importer and reader, but it would be huge amount of work :-/

### DIFF
--- a/libosmscout-import/src/osmscout/import/GenAreaNodeIndex.cpp
+++ b/libosmscout-import/src/osmscout/import/GenAreaNodeIndex.cpp
@@ -422,7 +422,9 @@ namespace osmscout {
     FileOffset previousOffset=0;
 
     for (const auto& tileEntry : tileData) {
-      //writer.WriteCoord(tileEntry.first);
+      if (storeGeoCoord) {
+        writer.WriteCoord(tileEntry.first);
+      }
       writer.WriteNumber(tileEntry.second-previousOffset);
 
       previousOffset=tileEntry.second;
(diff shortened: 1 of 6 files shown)


Framstag:
IMHO such errors should fail fast, that means use assert() to test and signal an error. I always have the fear that such error, if they do not get asserted, will possible take a lot of time to get detected and analysed.

Assert forces you to fix such bugs.

If you use assert() you can also be sure that there is no performance issue, since assert calls will be dropped if compile in release mode.

As I can see you do not use the assert macro but the surrounding defines to create a similar solution.

Problem with tests, is that the bug is (in some situations) in the combination of importer writing and test reading the data. To test this, data writing logic in the importer possibly would have to get wrapped in data objects that are (de)serialized. This way we could make simpler tests - but it would still be a major effort.
Thanks for the fix, waiting for the builds to finish.

PR #730, opened 2019-04-24
Import doc

Karry:
Hi. 

One user reports me that one Finland lake is not visible on the map: https://www.openstreetmap.org/relation/402543

Problem was with the limit of relation members during import. We have option for it already (thank you!), so I just increased this limit in my scripts. I updated the documentation at least :-)


Karry:
Damn. It seems that multiple islands in this lake is still missing :-/

PR #741, opened 2019-05-23
Vehicle position

Karry:
Next step to make navigation with libosmscout / QML more usable - ability to display estimated vehicle position (and its bearing) on the map. As vehicle icon is used simple arrow that is placed in directory with other map icons....

As a next step I want to use vehicle position to automatically rotate and move map.

obrazek


GNUuser:
looks very nice @karry; you the best on it. : )

Issue #754, opened 2019-06-22
Wrong WaterIndex generated by BasemapImport


vyskocil:
When I render the water.idx computed by BasemapImport there are large holes and sometimes small rectangle missing and the patterns evolve when zooming.
It seems that large holes surround small islands.

Here are some sample screenshots : 

![IMG_0027](https://user-images.githubusercontent.com/452704/59969677-eba82400-9553-11e9-83d2-cd274c45b7a8.jpg)
![IMG_0028](https://user-images.githubusercontent.com/452704/59969678-f19e0500-9553-11e9-8b2b-550507548cf4.jpg)
![IMG_0029](https://user-images.githubusercontent.com/452704/59969680-fe225d80-9553-11e9-99df-391bf8827f34.jpg)
![IMG_0030](https://user-images.githubusercontent.com/452704/59969681-024e7b00-9554-11e9-92b3-7a0e671a070d.jpg)

Karry:
Hi. 

The huge "unknown holes" in oceans and seas are created by algorithm naturaly and it should be possible to fill it just by tuning imporr settings. 

Small rectangles around some coastlines is bug. I saw it already, but dont had time (motivation) to investigate it yet. I will try to fix it...

Karry:
I just found why the problem is not visible with standard stylesheet - unknown color is same as water... `COLOR unknownColor = @waterColor;`

vyskocil:
I compiled a new water index with this code using the following command line :

`./BasemapImport --destinationDirectory world --coastlines coastlines-split-4326/lines.shp  --maxIndexLevel 12`

At lower zoom levels the basemap seems ok but when I zoom in I got the following map : 

![IMG_0031](https://user-images.githubusercontent.com/452704/60140216-3f637900-97b1-11e9-979b-20459168309a.PNG)

Karry:
@vyskocil this issue is well known for me. problem is that coastlines sometimes intersect with each other after simplification and algorithm then badly evaluate water/land tiles, because it is expecting that coastlines don't intersects. This change with of parameter makes this problem "global". 

For that reason, I updated pre-processing method that is removing smaller objects when intersection is detected. https://github.com/Framstag/libosmscout/pull/758 In the past, these intersections was checked just for ways vs. areas, but in basemap, even continents are areas... So I sorted all objects by its size and then check first 100...

[...]

PR #758, opened 2019-06-26
WaterIndexProcessor fixes

Karry:
solution for most visible problems with waterindex - https://github.com/Framstag/libosmscout/issues/754#issuecomment-505658824

explanation in code:

 > In some countries are islands too close to land or other islands
 > that its coastlines intersect after polygon optimisation.
 >
 > It may cause problems in following computation, that strongly rely
 > on the fact that coastlines don't intersect.
 >
 > For that reason, we will try to detect such intersections between land
 > (line coastlines) and islands (area coastlines) to remove the most visible
 > errors. Detecting intersections between all islands is too expensive.
 > errors. Just biggest coastlines (top 100) will be considered, detecting
 > intersections between all islands is too expensive.


Framstag:
Seems, like there are problems under Windows regarding std:min, min usage...is this because of your changes?

Karry:
@Framstag I will check it

@vyskocil recent commit should fix unknown tiles. problem was that in code was used bounding box of original coastline, not transformed. and for some tiles, it provides invalid results...

PR #759, opened 2019-06-26
Support for basemap rendering in DrawMap* demos

Karry:


Karry:
@vyskocil I make base map rendering as separate rendering step and removed duplicate code from Qt and iOS MapPainters. Can you please verify that iOS painter is working correctly? Thank you!

Framstag:
Build your patch under OS X. Looks ok to me and identical to the original rendering. Is this enough or dso need explicit base map rendering?

test

Karry:
It would be great to test standard database with basemap. I am pretty sure that it will be fine, but I cannot test it...

Framstag:
Hmmm, the current DrawMapOSX does not support basemap rendering - at least it does not have an option for it.

Framstag:
I did port the demo to the new approach...

test

More helpful?

Karry:
Definitely :-) You are free to merge it, together with your changes in OS X demo...

vyskocil:
@Karry and @Framstag  just had time to also check it, thank you ! 
map

PR #761, opened 2019-06-30
enable asserts even for release, optimised builds

Karry:
Hi, what do you think about this change?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,14 @@ if((CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANGXX OR CMAKE_COMPILER_IS_G
   endif()
 endif()
 
+# Do *NOT* turn assertions off in release and rel-with-deb-info modes.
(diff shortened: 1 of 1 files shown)


Framstag:
IMHO people do not want to have assertions in production code, because this may make the application crash before customer eyes and often give no change to get a bug report. I at least read about this opinion sometimes. That sounds reasonable for me.

IMHO having assertions on in debug mode is OK, since we can assume that the actual assertion will happen on the developer system - and he should be made aware of his or our bugs.

I could image to give the user the option by supplying simple switches for this, though IMHO still the default should be defensive.

You did not give your argumentation, please do so.

Karry:
Well, I usually want to have optimised code `-O3` with asserts. When you get to invalid state, application crash soon or later anyway (in most cases). Without asserts, you will just gets confusing error message. And it is possible that Sailfish userbase is different, people are reporting crashes back to developers :-)

But it is true that same combination of flags can be made by: `-DCMAKE_BUILD_TYPE=DEBUG -DCMAKE_CXX_FLAGS=-O3`

PR #768, opened 2019-08-08
Follow vehicle during navigation

Karry:
Initial implementation for input handler (Qt MapWidget) that is following vehicle. It moves and rotate map based on car position and changing magnification level based on distance to next routing step...

Solution is not final. There is missing interaction with usual user gestures. My idea is to use delay (4 seconds?) when `VehicleFollowHandler` will be inactive after user input (gesture). But I need to test on the phone first. Do this complex UX properly is hard :-)

### DIFF
--- a/OSMScout2/qml/NavigationSimulation.qml
+++ b/OSMScout2/qml/NavigationSimulation.qml
@@ -42,9 +42,9 @@ Window {
         }
 
         onPositionChanged: {
-            if (!map.lockToPosition){ // don't set again when it is true already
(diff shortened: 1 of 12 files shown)


Framstag:
Hello @Karry: There seem to be more than usual build breaks. Can you take a look?

Karry:
It seems that I was using `QQuickItem::size()` method that was not available in older versions. Another fix was for `showCoordinates` method. Lets see how it will be going...

MacOS on Travis seems to have broken Pango installation...

PR #775, opened 2019-09-17
Qt LookupModule: add info about all outer area rings …

Karry:
…even when ring type is ignored.

### DIFF
--- a/libosmscout-client-qt/src/osmscout/LookupModule.cpp
+++ b/libosmscout-client-qt/src/osmscout/LookupModule.cpp
@@ -116,13 +116,25 @@ void LookupModule::addObjectInfo(QList<ObjectInfo> &objectList, // output
                    std::map<osmscout::FileOffset,osmscout::AdminRegionRef> &regionMap)
 {
   for (const auto &ring:a->rings) {
-    if (!ring.GetType()->GetIgnore()) {
+    if (ring.GetRing() == Area::outerRingId || ring.GetRing() == Area::masterRingId) {
+      // for outer ring use type and features from  master ring (area type)
       addObjectInfo(objectList,
                     "area",
                     a->GetObjectFileRef(),
                     ring.nodes,
                     // Master ring don't contains nodes! Use intersection of all outer rings instead
                     (ring.nodes.empty() ? a->GetBoundingBox() : ring.GetBoundingBox()).GetCenter(),
+                    a,
+                    reverseLookupMap,
+                    locationService,
+                    regionMap);
+    } else if (!ring.GetType()->GetIgnore() && !ring.nodes.empty()){
+      // inner rings with defined types and some nodes
+      addObjectInfo(objectList,
+                    "area",
+                    a->GetObjectFileRef(),
+                    ring.nodes,
+                    ring.GetBoundingBox().GetCenter(),
                     &ring,
                     reverseLookupMap,
                     locationService,
--- a/libosmscout-client-qt/src/osmscout/MapObjectInfoModel.cpp
+++ b/libosmscout-client-qt/src/osmscout/MapObjectInfoModel.cpp
@@ -120,6 +120,8 @@ QObject* MapObjectInfoModel::createOverlayObject(int row) const
         o->addPoint(p.GetLat(), p.GetLon());
       }
     }
+  } else {
+    qWarning() << "Object " << obj.name << " (" << obj.type << " / " << obj.objectType << ") has no points!";
   }
 
   return o;


Karry:
Hi Tim. 

Idea behind `LookupModule::addObjectInfo` is to provide information about element objects that `LocationEntry` consist from (LocationEntry is search result for example). For nodes and ways, it is simple. They consists just from one object. But for complex areas, there is master ring that represents whole structure (OSM relation), outer rings and inner rings - clippings. Every ring may have its own type and features...

Do you remember why, in the (https://github.com/Framstag/libosmscout/blob/master/libosmscout-map/src/osmscout/MapPainter.cpp#L1322)[MapPainter], for outer rings is used type of Area, but features of the ring itself? Inner rings (island on the river bank for example) seems to be simple, ring type and features are used... It would be great to document Area/Ring more deeply with few examples.

Karry:
As I am looking to https://wiki.openstreetmap.org/wiki/Relation:multipolygon , outer ring should not have type right? Or there is any valid example when outer ring should have type? Or better question, type that is different than master ring?

And second question - as I am looking to `RelAreaDataGenerator::ConsumeSubs` function, ring id / hierarchy basically determine level of encapsulation, right? So all top-level outer rings should have id 1 (`outerRingId`), all inner rings should have ringId 2, when there is another outer ring (island in OMS wiki terminology), it should have ringId 3 ?

Framstag:
I would have to look into more details to make more concrete explanations.

But base of the code is https://wiki.openstreetmap.org/wiki/Relation:multipolygon and https://wiki.openstreetmap.org/wiki/Multipolygon_Examples

Because of wrong tagging or historical tagging (which is now deprecates, not support by most tools and under active migration) some transformation have to be done to result in correct drawing. IMHO we would have to look at the code to see if some parts are now obsolete and can be removed (see above).

Yes ring number defines the tree-like hierarchy. See the actual drawing code that "walks the tree". Rings with hider ring id should be fully contained by the prior ring with lower id. Note that "fully contained" is not enforced or checked and thus can result in rendering errors if the constraint is broken.

Without looking at the code, uneven rings should always have a type (so the outer ring should always have a type), even rings can have a type but can also only be a clipping region. I do remember though that the outer ring is special and there were situations where it does not have an area itself.

As far as I remember there could also be multiple outer rings with different types?

Does this help or do you need more details (and I have to look at the code again)?

Karry:
Thank you. You explains it. I am little bit worry that we don't render outer rings with `ringId >= 3` properly. 
At least when this case is correct? https://www.openstreetmap.org/relation/24011#map=15/50.3942/14.8547
On the `Relation:multipolygon` is this case mentioned as a valid one, but `Multipolygon_Examples` suggest that the "island" should be as separate relation - it sounds like overkill to me. 

I will to make some refactoring and fixes around it...

Framstag:
Regarding above relation: From the OSM relation view it looks like the relation has the following structure:

relation (landuse=forest)
* 26019383 (no type)
  * 26019386 (no type)
    * 314945870 (no type)

I assume that libosmscout would give them ringId 1, 2 and 3. It should draw the ring 1 as forest (inheriting the type of the relation itself) using ring 2 as clipping. It then should draw ring 3 using its type. Since it does not have a type -what should libosmscout do? draw again a forest?

Also: While technical OK, ways 9389365 and 9863644 define additional areas within the relation. IMHO they should be part of the relation to make sure things are rendered in correct order. Libosmscout tries to handle this by making each ring a separate area and ordering areas before drawing, but nevertheless...

Also ring 3 looks rather artificial, is it really "on ground"?

And finally: Libosmscout ignores inner and outer hints and decides on itself base don containment. If rings overlap or touch this can fail.

Framstag:
Looking at island with a hole: The example does not give information which ways have a type. Libosmscout will only draw rings with a type. Rings without a type will be just used for clipping. So the island with a while would need to have the following structure in libosmscout:

Relation (island)
* outer way (no type, inherits type from relation)
  * hole (no type, just clipping for island)
    * inner island (type=island)

if the inner island does not have a type, it will not be drawn, because IMHO its type is not obvious. That would be in line with "Forest with two lakes and an island (Nested multipolygons)" from the examples where the inner island has an explicit type.

Karry:
Well, I think that it should be forest... Take a look to https://wiki.openstreetmap.org/wiki/Relation:multipolygon , paragraph "Island within a hole"

> A construct like this would  previously have required different multipolygon relations, one with way 1  being outer and way 2 being inner, as well as one with way 2 being  outer and way 3 being inner. Such cascading is still needed when the  "island" in the middle is something else than the area on the outside,  but where the "island" is the same stuff it can just be made a hole in  the hole. | Figure 7: Island within a hole

It seems to be enough to use different definition of "outer ring" in Painter (and few other changes):
```
      // ring level is odd, it is some outer ring
      inline bool IsSomeOuterRing() const
      {
        return (ring & outerRingId) == outerRingId;
      }
```
It provides the same output as Mapnik...

Framstag:
Sorry I missed to answer. I'm still unsure, because IMHO we could not distinguish clipping areas from "no type but you the type of the outer ring" rings? Does it make sense to ask in the mailing list?

Karry:
It is true that we cannot distinguish it in runtime, we don't have original relation data anymore. But we can check that "island has outer role" during import time and remove inner rings that are nested in another inner ring - that is geometry bug in data. But then code that is handling rings will be even more complicated :-( I am playing with the idea of visitor pattern inside Area object that may simplify `MapPainter::PrepareArea` method... 

Anyway, I will ask for that special case on Czech OSM mailing list, to be sure.

I will prepare better solution and closing this merge request meanwhile...

PR #776, opened 2019-09-29
Water symbols

Karry:
Hi Tim. 

I just miss waterfalls on the map, so here it is... As a way, I made it same as weir (160 DPI):

image

More common should be waterfalls as a node (300 DPI):
image

And similar, new node symbol for weir (300 DPI):
image

### DIFF
--- a/libosmscout-map-qt/src/osmscout/MapPainterQt.cpp
+++ b/libosmscout-map-qt/src/osmscout/MapPainterQt.cpp
@@ -759,14 +759,14 @@ namespace osmscout {
         double         radius;
 
         if (circle->GetProjectionMode()==DrawPrimitive::ProjectionMode::MAP) {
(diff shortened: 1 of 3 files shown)


Framstag:
I still think we need some automatic rotation for symbols on ways. This would however require libosmscout to associate nodes and ways (and areas) either during runtime or import time. Which in turn would make the data model more complicate, would impose restrictions (objects on ways are possibly only drawn if the way itself is drawn) but may give the map more semantics and thus better apearance...?

Karry:
It make sense to me, especially in case of these two node symbols. But no idea howto do it simply. Possibly add new feature "angle" and during import evaluate angle of way where node sits? But even Mapnik don't do it, even when it has more expressive styling language and don't target real-time rendering performance.

Issue #777, opened 2019-10-06
Import parameter --langOrder is not working


vyskocil:
I imported the latest france.osm.pbf from geofabrik with the following command :
 
`Import --destinationDirectory . --langOrder fr,# --bounding-polygon ../france.poly ../france.osm.pbf ../france.hgt.osm.pbf`

I expected that `name:fr` tags would be parsed before `name` tags but this isn't working.
Here is a sample, a peak at the French and Italian border, the "Grand Cimon de Rabuons" in french and "Becco Alto d'Ischiator" for the Italian people. Here is the OpenStreetMap data : 

<img width="599" alt="Screenshot 2019-10-06 at 12 08 36" src="https://user-images.githubusercontent.com/452704/66267649-f0a78f80-e833-11e9-91c6-44be06166bed.png">

Here follow screenshots from OSMScout2 where we could see that this is always the Italian name that is drawn and we could not find the french name : 

<img width="484" alt="Screenshot 2019-10-06 at 12 06 21" src="https://user-images.githubusercontent.com/452704/66267676-572cad80-e834-11e9-9ca0-c2452f2a7a9d.png">

Searching for "Grand Cimon de Rabuons"... it is not found. We could also see that the search text is overlapping the textfield : 

<img width="481" alt="Screenshot 2019-10-06 at 12 27 19" src="https://user-images.githubusercontent.com/452704/66267709-b094dc80-e834-11e9-926e-31bb6e2310e1.png">

Searching for "Becco Alto d'Ischiator" : 

<img width="480" alt="Screenshot 2019-10-06 at 12 07 36" src="https://user-images.githubusercontent.com/452704/66267685-6f9cc800-e834-11e9-9d60-dfff8ebd80e5.png">

vyskocil:
I also noticed this repeating warning in the logs :

```
Error in UTF8StringToWString()Illegal byte sequence
Error in UTF8StringToWString()Illegal byte sequence
Error in UTF8StringToWString()Illegal byte sequence
Error in UTF8StringToWString()Illegal byte sequence
Error in UTF8StringToWString()Illegal byte sequence
Error in UTF8StringToWString()Illegal byte sequence
...
```

Framstag:
[...]

PR #784, opened 2019-10-26
fix tag priority while parsing name* features

Karry:
Priority of name tags is increasing with its order.
When lang order is `fr,#`
```
 - tag name:fr has priority 0
 - tag place_name:fr     => 1
 - tag brand:fr          => 2
 - tag name              => 3
 - ...
```
It means that tag with lowest priority should win during parsing.
This commit fixes that.

@vyskocil  can you check that this branch is fixing your issue? https://github.com/Framstag/libosmscout/issues/777

@Framstag it is correct to prioritize `name` before `brand`? Makes it sense to you?

### DIFF
--- a/libosmscout/src/osmscout/TypeFeatures.cpp
+++ b/libosmscout/src/osmscout/TypeFeatures.cpp
@@ -94,14 +94,14 @@ namespace osmscout {
                           FeatureValueBuffer& buffer) const
   {
     std::string name;
-    uint32_t    namePriority=0;
+    uint32_t    namePriority=std::numeric_limits<uint32_t>::max();
 
     for (const auto &tag : tags) {
       uint32_t ntPrio;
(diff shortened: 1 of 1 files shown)


vyskocil:
I imported the france.osm.pbf map again with `--lang fr,#`  and with this fix it works now, the peak in this map was previously named with its Italian name found in the `name` tag but now the lib use the `name:fr` tag as intended ! Thank you for the fix. 

IMG_6988

vyskocil:
For the `name` vs `brand` tag priority I think it's logical to use the `name` tag first and use `brand` as a fallback if there is no name

Framstag:
In principle name and brand are two different things. If you want to brand you should write your own feature for this. However we want high quality name information, so a fallback to the brand does make sense IMHO, thing of McDonalds ;-)

PR #792, opened 2019-11-07
Add support for Imperial units (elevation label)

Karry:
Hi.

I was expecting that this request will arrive some day https://github.com/Karry/osmscout-sailfish/issues/174 :-)

Question is if would like to add Metrics to `FeatureValue::GetLabel` (that I did) or create `LabelProvider` that will handel Units just fro elevation...

### DIFF
--- a/Demos/src/LocationDescription.cpp
+++ b/Demos/src/LocationDescription.cpp
@@ -67,7 +67,7 @@ void DumpFeatures(const osmscout::FeatureValueBuffer &features, const std::strin
         if (feature->HasValue()) {
           osmscout::FeatureValue *value=features.GetValue(featureInstance.GetIndex());
           if (feature->HasLabel()) {
-            std::string label=value->GetLabel(osmscout::Units::Metrics, 0);
+            std::string label=value->GetLabel(osmscout::Locale(), 0);
             if (!label.empty()) {
               std::cout << ": " << osmscout::UTF8StringToLocaleString(label);
             }
--- a/Demos/src/LocationDescription.cpp
+++ b/Demos/src/LocationDescription.cpp
@@ -67,8 +67,9 @@ void DumpFeatures(const osmscout::FeatureValueBuffer &features, const std::strin
         if (feature->HasValue()) {
           osmscout::FeatureValue *value=features.GetValue(featureInstance.GetIndex());
           if (feature->HasLabel()) {
-            if (!value->GetLabel(0).empty()) {
-              std::cout << ": " << osmscout::UTF8StringToLocaleString(value->GetLabel(0));
+            std::string label=value->GetLabel(osmscout::Units::Metrics, 0);
+            if (!label.empty()) {
+              std::cout << ": " << osmscout::UTF8StringToLocaleString(label);
             }
           }
           else {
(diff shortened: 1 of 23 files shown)


Karry:
I am not sure if to put space between elevation and unit... is correct typographically, but it waste the space little bit

Framstag:
The original idea to support this or similar feature requests was to use the labelIndex parameter of FeatureValue::GetLabel() together with Feature::RegisterLabel(). This way to can choose between various label variants via their index name in the style sheet (height_in_meter, height_in_feet,...). This feature though was up to now never been actually used (because of lack of feature requests).
For ele one could define three labels variants: One for "based on locale", one for metric and one for imperial.

For *scanning* the patch it looks like you did not choose this way, but added a central flag for unit. Correct?

Judgment of your solution:
* Central configuration flag, no changes in the style sheet
* Thus simpler for the developer of applications
* All Values have the unit in their parameter list, while most of them do not actually react to it.

Judgment of the initial solution:
* High control, support for various variants for label beyond units
* Nice names in style sheet, but you must change the style sheet.
* Better internal API (more "locality" of actual unit feature, more generic approach), values are more in control in what they actually offer, but needs more documentation and naming rules.

Disadvantages of both solutions:
* Complexity if there a multiple inputs to the labels with result in either many parameters or many label names.

I'm of course willing to path through your path, but I would like you to at least be aware of the alternative and weight them in your role as user of libosmscout. I'm of course also interested in your judgment.

Sorry for looking this late into the path :-(
*

Karry:
> For scanning the patch it looks like you did not choose this way, but added a central flag for unit. Correct?

Exactly. Beside this customisation of elevation label, I want to use this configuration option (by default selected by locale) for display correct distances in UI.

In this case I would prefer to not use stylesheet customisation, because used units are tied with user configuration (locale). I think that is more closer to choice of primary or alternative language, where special LabelProvider is used. 

> All Values have the unit in their parameter list, while most of them do not actually react to it.

Yes, this is one thing that I don't like much...

Framstag:
We can still place in the stylesheet the label index for "UNIT_BY_LOCALE" or even make it index 0 and thus the default, and make "NO_UNIT, UNIT_METER, UNIT_IMPERIAL..." alternatives. This way we have a good default and still flexibility via stylesheets and a better API? The same would work for UNIT_BY_CONFIG, which falls back to UNIT_BY_LOCALE

UNIT_BY_LOCALE would either evaluate locale directely or would get passed a config object which hold the unit to choose - still config object would though be much more general than the explicit unit enumeration. I have no problem with an  additional parameter, but it should hold all options that are likely common for multiple features.

so class FeatureContext with an attribute unit, instead of passing unit directly?

Your opinion?

Karry:
Agree in principle. Just instead of `FeatureContext` I would name it just `Locale` :-) Because right now I can imagine just few things that may be there... Units, decimal separator, thousands separator, unit separator... All of them belongs to locale. It may be renamed later when this name will not cover functionality...

With label index, agree that sometimes may be useful to define units in stylesheet. I may imagine that map will display peak elevation by locale - in feet and elevation of contour lines in meters - because these were imported with 10 meter density...

I will try to implement it...

Framstag:
Fine :-)

PR #804, opened 2019-12-17
Area refactoring

Karry:
Continuation of https://github.com/Framstag/libosmscout/pull/802

It is refactoring how rings are iterated in `MapPainter` and Qt `LookupModule`

 - render nested outer rings properly
 - fix clipping for areas with nested rings
 - provide utility for visiting rings in breadth-first manner
 - use this visitor pattern in MapPainter to separate complicated ring iteration and painter code

Right now, this ruin seems to be rendered properly: https://www.openstreetmap.org/relation/7281899

Screenshot_20191217_074040


Framstag:
Very nice :-)

PR #809, opened 2020-01-07
nicer tree symbol

Karry:
- remove tree icon
 - make tree symbol bigger, with semitransparent treetop
 - mark tree symbol as overlay to allow overlaps

Screenshot_20200107_124544


Framstag:
Neat :-)

PR #813, opened 2020-01-12
Fix SRTM filenames and calculation

vyskocil:
Fix SRTM filenames and calculation for south hemisphere and west of Greenwich meridian
Deactivate for now STRM bilinear height interpolation


Framstag:
Manually restarting the wercker build...

PR #828, opened 2020-03-10
UI models for navigation voice

Karry:
UI models for managing local voices: download, remove, set preferred...


Karry:
AppVeyor error with Ubuntu1804 image seems to be temporary issue with repository...

PR #831, opened 2020-03-17
navigation: add unitialised position state

Karry:
Greetings from quarantine!

Here is fix for one annoying bug - "random" jumps to zero-island (gps coordinates 0,0) during navigation, when route is re-computed... I just added "uninitialised" entry to position state enum, logic is now following:

 - `Unitialised`, position newer should be used in UI
 - `NoGpsSignal` now means that last know position is used, may be inaccurate

other enum entries are unchanged:

 - `OnRoute`, vehicle is on the planned route
 - `OffRoute`, vehicle is out off planned route, route should be re-computed
 - `EstimateInTunnel`, vehicle position is estimated in tunnel

### DIFF
--- a/libosmscout-client-qt/include/osmscout/NavigationModel.h
+++ b/libosmscout-client-qt/include/osmscout/NavigationModel.h
@@ -135,7 +135,7 @@ public:
 
   inline VehiclePosition* getVehiclePosition() const
   {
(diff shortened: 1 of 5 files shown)


Framstag:
Home office with children, wife still needs to drive to work...

PR #835, opened 2020-03-29
add junction delay penalty to fastest routing profile

Karry:
With this penalty, fastest profile avoids route thru city center that is not faster usually

### DIFF
--- a/libosmscout/include/osmscout/routing/AbstractRoutingService.h
+++ b/libosmscout/include/osmscout/routing/AbstractRoutingService.h
@@ -196,7 +196,8 @@ namespace osmscout {
     virtual double GetCosts(const RoutingState& state,
                             DatabaseId database,
                             const RouteNode& routeNode,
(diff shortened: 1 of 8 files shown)


Framstag:
This one is interesting. What are you doing exactly. It looks like you increase the cost in cases of junctions? Junctions are detected by comparing the cost of incoming and outgoing junction speed and just that you are turning in case of a speed change?

Karry:
in recent version I am doing it more easier:
```
      // add penalty for junction
      // it is estimate without considering real junction geometry
      double junctionPenalty{0};
      if (applyJunctionPenalty && inObjIndex!=outObjIndex){
        double minSpeed=std::min(GetMaxSpeed(inPathVariant),GetMaxSpeed(outPathVariant));
        junctionPenalty = 0.160 / minSpeed;
      }
```
Every time when you leave junction by different way than you enter in, add penalty that corresponds to travel 160 meters with minimum speed on these two ways.

PR #837, opened 2020-04-05
Lane turns

Karry:
Hi. 

This is small addition to map renderer - support displaying of lane turns. Screenshot better than words:

Screenshot_20200405_104127

To achieve it, I added:
 - support for relative offset for way icons
 - relative offset by turn type (laneForwardLeft, laneForwardThroughLeft, laneForwardThrough...), one offset may be multiple times on the way
 - symbols for turns, add it to ways


Framstag:
Wow, what a cool feature :+1:

PR #842, opened 2020-04-13
Vs2019

Framstag:
Further improvements to the MSVC builds, improving calls to vcpkg, avoiding failed builds.


Framstag:
Seems like qt-base builds longer than 1 hour. Hmm....

PR #843, opened 2020-04-13
std::optional

Karry:
Use `std::optional` in navigation (instead of std::shared_ptr) and gpx library (instead of custom Optional class). 

This is small step with usage of c++17


Framstag:
std::optional seems to make problems under OS X. I made a new branch to switch from XCode 9.3 to 11.3 since std::optional should in principle be available...

PR #846, opened 2020-04-13
Use imported CMake GLEW target

Karry:


Karry:
Not sure if it will be working on Windows, OpenGL backend is not build on AppVeyor...

Framstag:
Looks like I can merge this one first and then the pulls request regarding std::optional?

Karry:
Order is no important, these should be independent I believe.

Framstag:
OK, merged this one. Regarding the Windows build I'm lost. The idea I had was nice and works in principle but actual hardware resources seem that reduced, that even building qt5-base takes too long. Looks like I have three options:
* Wait for vcpkg supporting central caching
* Skip building Qt5 using MSVC (possibly legitimate, since we already have a msys build for this)
* Try to install a pre-build Qt5 (from Appveyor or somewhere else), mixing vcpkg and foreign Qt5 builds.

Framstag:
Waiting for your go for the std::optional pull request merge.

Karry:
I think that skip building Qt5 is the simplest solution...

> Waiting for your go for the std::optional pull request merge.

I just rebase it, you may merge it when builds finish

PR #853, opened 2020-04-17
Initial build for github and msys

Framstag:


Framstag:
OK, looks like MSYS2 is not installed (but only MINGW).

See https://github.com/marketplace/actions/setup-msys2-environment for a temporary alternative and https://github.com/actions/virtual-environments/issues/30 for status of official support.

PR #854, opened 2020-04-19
Do not inherit from std::iterator

Framstag:


libosmscout/include/osmscout/TypeInfoSet.h
```diff
@@ -28,15 +29,22 @@
...
 
-  class OSMSCOUT_API TypeInfoSetConstIterator CLASS_FINAL : public std::iterator<std::input_iterator_tag, const TypeInfoRef>
+  class OSMSCOUT_API TypeInfoSetConstIterator CLASS_FINAL
   {
+  public:
+    typedef TypeInfoSetConstIterator self_type;
+    typedef TypeInfoRef              value_type;
+    typedef const TypeInfoRef&       reference;
+    typedef TypeInfoRef              pointer;
+    typedef std::input_iterator_tag  iterator_category;
```
Karry, line 39: please consider to use `using Alias = OriginalType` instead of `typedef`. Meaning is the same, but I think that `using` is more natural, because alias is on the left side of expression.

Framstag: OK

PR #855, opened 2020-04-19
use std::sort (stable_sort) intead of qSort (qStableSort)

Karry:


Framstag:
Hmm, where are the github builds?

Karry:
Weird. It was build in my fork, here: https://github.com/Karry/libosmscout/actions?query=branch%3Astl-sort but it is not displayed in merge request...

PR #856, opened 2020-04-19
Tests

Karry:
Hi. I added small region database (500 KiB) to git repository (`Tests/data/testregion`) and add meson and cmake tests that requires database.


Framstag:
(Haven't loked into the details...) Isn't libosmscout-test the way to go if you just want to have a simple scenario of nested regions? This way one does not need to work an real-life data file but can make an own.

I initially had the idea to add further primitives to allow adding ways, nodes and areas for doing tests on certain routing scenarios.

The problems with data processing tools is, that there is no business logic. So tests quickly mutate to integration tests based on real-life data. Which of course for various reason is an anti-pattern (in principle, speed, complexity, builds break often, undefined or implicitly changing test scenarios...). The DSL defined by the test library was a way to still have integration tests but have a simple way to define custom data.

Karry:
Well, I probably was not precise. I just cut small region from Czechia, import database and put it to git repository. It is not imported during the test, to avoid fragile outputs and run test with the same data every time. I am expecting that data will needs to be re-imported just in case of database format change.

Many bugs are hard to discover with synthetic data or preparation of such data is complicated. Newly added tests are `ThreadedDatabase`, `MultiDBRouting` and `CoordinateEncoding`. Especially routing is not covered by automated tests and preparation of synthetic data would be tricky for it.

PR #861, opened 2020-05-05
Initial support for routes.

Framstag:
* Plan is to support public transport V2 and make basic
  information available.
* Added besides MULTIPOLYGON two new subtypes ROUTE_MASTER and ROUTE
* Added new map.ost types for common routes
* Store these relations into two new rawXXX.dat files.
* Added new features operator and network

My initial goal is to make basic data available for further processing (up to a *.dat file).
Personal plan is to generate some visual representation of the network of an operator, but we will see if I have time for this.


Karry:
Great to see work in this topic! I want to visualize [route=bicycle](https://wiki.openstreetmap.org/wiki/Tag:route%3Dbicycle) for long time. How do you want to provide these data to renderer? Add another object type (beside area, way, node) or make route linked to way somehow?

Framstag:
My plan is to extract all raw data to be able to generate something like this: https://mapa-metro.com/mapas/Dortmund/mapa-metro-dortmund.png - of course using a different and simpler geometry.

Data would come from the enhanced database and a seperate tool would likely generate a SVG (if possible).

So no rendering and no routing. I hope to have some time for the data file and simple experiments with the actual map. It is unlikely that  there is more time for the below soon.


However, the raw data (in form of a new data file) should be a good start to play with it and add further features on top and make experiments with how to integrate such stuff in rendering and possibly even routing. I hope this can be done by adding index files on top and enhancing existing preprocessor steps.

Framstag:
I had not read your reference to route=bicycle. I'm currently writing code based on the public transport V2 schema.

route=bicycle is likely similar but not the same. However most of the code can likely be copied and adapted.

PR #878, opened 2020-06-06
Import progress

Karry:
Hi. When I was thinking about extending import pipeline, I had problem that my knowledge about current state is limited. So, here is first attempt to visualise it. New progress class is collecting statistics and metadata about import modules and at the end, it dumps it to [diagram in "dot" format](http://osmscout.karry.cz/tests/stats.dot) (Graphviz project). Later you may convert it to image...

You can see it here:
http://osmscout.karry.cz/tests/stats.svg


Framstag:
The meson build files are not updated, thus meson builds fail.

I'm also currently pushing another bunch of clang-tidy improvements. I assume that you have to rebase your patch afterwards.

Else: nice addition :-)

Framstag:
clang-tidy branch is now merged...

Karry:
Ok, I just rebased it...

Issue #887, opened 2020-07-01
Routing issue with roundabouts


vyskocil:
Quite recently something was broken with routing, I hadn't used the car profile for some time because I usually use foot itineraries calculations in my app but yesterday I couldn't calculate a route using the car profile which used to work before...
I tried to narrow the problem and found that the culprit seems to be  roundabouts between the start point and the destinations.
Here are two screenshots made with the latest libosmscout code and france.osm.pbf data around location (43.61598, 6.97264), the first one is calculated with the car profile when the destination is just before the roundabout (at the bottom of the picture), if the destination is in the roundabout or after it the route could not be calculated :

![IMG_8318](https://user-images.githubusercontent.com/452704/86243673-12595b80-bba7-11ea-8cc8-e3605540873f.jpg)

Here is a route calculated with the pedestrian profile, it is ok even when the destination is after 2 roundabouts but we could see that the second one is taken on the whole around ! : 

![IMG_8320](https://user-images.githubusercontent.com/452704/86243882-649a7c80-bba7-11ea-8560-66eaf51d35ac.jpg)

vyskocil:
Maybe it is not the roundabout the cause of this issue, I found that if the starting point is after the black dot on the following picture the route could be calculated but if it is before it in the red zone on the picture the route calculation fail : 

![Screenshot 2020-07-01 at 19 29 05](https://user-images.githubusercontent.com/452704/86274400-ed2e1280-bbd1-11ea-946a-0dc996b44080.JPG)

Karry:
Can you provide me exact coordinates when routing fails? I tried for example `43.61630 6.97148` as a start and `43.61457 6.97354` as destination and it works fine... 

Here is last step from routing animation:
```
./Demos/RoutingAnimation \
  --zoom 18 \
  --style ../stylesheets/standard.oss \
  ~/Maps/europe-france-20200701-200659 \
  43.61630 6.97148 \
  43.61457 6.97354
```

![0000079](https://user-images.githubusercontent.com/309458/86326226-ad067880-bc41-11ea-96a9-ef440caab1ae.png)

vyskocil:
Hi Karry !
Thank you for investigating !
I tried with the from/to coordinates of your sample and it also works for me with ./Demos/RoutingAnimation but not in my app... I'll check what's wrong with my code because the same calculation used to work...

Karry:
[...]

PR #890, opened 2020-07-09
limit junction penalty to 10 seconds

Karry:
When way speed limit is low (~10 km/h), junction penalty
is computed unrealistically high (250 meters / 10 km/h = 90 seconds).

We should limit maximum time for junction penalty.
Default maximum penalty is setup to 10 seconds now.

It is solving issue https://github.com/Framstag/libosmscout/issues/887

### DIFF
--- a/libosmscout/include/osmscout/routing/RoutingProfile.h
+++ b/libosmscout/include/osmscout/routing/RoutingProfile.h
@@ -23,6 +23,7 @@
 #include <map>
 #include <memory>
 #include <vector>
+#include <iostream>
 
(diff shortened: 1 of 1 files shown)


Framstag:
Just for better understanding, does the penalty just lead to unrealistic costs and thus to "just" bad routes or does it actualyl beak A* constraints?

Karry:
It just leads to bad time estimates. So, not the best route may be chosen or in the worst case, when cost (time) limit was reached, no route was computed at all.

Issue #892, opened 2020-07-25
Bug: Missing header for string_view_literals


OgreTransporter:
The following errors occurred when compiling with Visual Studio 2017:
```
libosmscout\src\osmscout\routing\Route.cpp(497): error C2039: "string_view_literals": Ist kein Element von "std"
libosmscout\src\osmscout\routing\Route.cpp(497): error C2871: "string_view_literals": Es ist kein Namespace mit diesem Namen vorhanden
libosmscout\src\osmscout\routing\Route.cpp(501): error C3688: ungültiges Literalsuffix "sv"; Literaloperator oder Literaloperator-Vorlage "operator ""sv" nicht gefunden
libosmscout\src\osmscout\routing\RoutePostprocessor.cpp(1483): error C2039: "string_view_literals": Ist kein Element von "std"
libosmscout\src\osmscout\routing\RoutePostprocessor.cpp(1483): error C2871: "string_view_literals": Es ist kein Namespace mit diesem Namen vorhanden
libosmscout\src\osmscout\routing\RoutePostprocessor.cpp(1536): error C3688: ungültiges Literalsuffix "sv"; Literaloperator oder Literaloperator-Vorlage "operator ""sv" nicht gefunden
libosmscout\src\osmscout\routing\RoutePostprocessor.cpp(1539): error C3688: ungültiges Literalsuffix "sv"; Literaloperator oder Literaloperator-Vorlage "operator ""sv" nicht gefunden
libosmscout\src\osmscout\routing\RoutePostprocessor.cpp(1540): error C3688: ungültiges Literalsuffix "sv"; Literaloperator oder Literaloperator-Vorlage "operator ""sv" nicht gefunden
libosmscout\src\osmscout\routing\RoutePostprocessor.cpp(1541): error C3688: ungültiges Literalsuffix "sv"; Literaloperator oder Literaloperator-Vorlage "operator ""sv" nicht gefunden
libosmscout\src\osmscout\routing\RoutePostprocessor.cpp(1542): error C3688: ungültiges Literalsuffix "sv"; Literaloperator oder Literaloperator-Vorlage "operator ""sv" nicht gefunden
libosmscout\src\osmscout\routing\RoutePostprocessor.cpp(1545): error C3688: ungültiges Literalsuffix "sv"; Literaloperator oder Literaloperator-Vorlage "operator ""sv" nicht gefunden
```

Reference page: https://en.cppreference.com/w/cpp/string/basic_string_view/operator%22%22sv

[...]

PR #893, opened 2020-07-25
Bugfixes and improvements

OgreTransporter:
- Bugfix for issue #892: Add missing header for string_view_literals
- Remove unused GLUT in CMake (issue #894)
- Bugfix missing include directories for GLM and GLFW (issue #895)
- Improvement: Automatically load GLM via GIT if it cannot be found. GLM is a header only library and therefore does not need to be compiled first. There is a switch `OSMSCOUT_DOWNLOAD_GLM_IF_NOT_FOUND` (default: `OFF`) to enable this feature.
- Bugfix: Fix different parameter declaration in h and cpp (issue #896)
- Update Qt search:
  - The first part is the _modern_ search for Qt with the desired components (see https://blog.kitware.com/cmake-finding-qt5-the-right-way/).
  - The second part adds the program _windeployqt_ under Windows to install the Qt DLLs if necessary.
- Improved search for Direct2D in the Windows 10 SDKs


Framstag:
In general I do not like build script to load data during builds. At least, if the build system does not active support for this (meson has, but we do not use it currently, vcpkg is and conan are pre-build package managers).

I understand the decision to do so for header only libraries. Still, would be vcpkg an option?

Everything is OK, thanks for the patches! We moved the test builds from MSVC 2017 to 2019, seems like we should have kept 2017, too?

OgreTransporter:
I can remove the GLM download part if you like. Or I'll add a switch with which you have to turn on the automatic download.

The error #896 also appears here under Visual Studio 2019. Error #892 only with Visual Studio 2017, but adding the header _string_view_ does not cause any errors in Visual Studio 2019 and should not be a problem under Linux.

Framstag:
I prefer a switch, but as I saw, you already changed this.

Thanks also to additional changes to the cmake file to make use of more modern features. I know that there is some room and appreciate more additions :-)

Since you are still adding commits, please give me a sign, if I should merge.

Framstag:
It would also be interesting what your plan is. Looks like you are plan to use libosmscout with Visual Studio under Windows :-) Of your you can use the mailing lust for this.

OgreTransporter:
Well, my plan is to actually use libosmscout on Windows. Preferably with Visual Studio, because that is installed on all machines here. If I build libosmscout with Visual Studio 2017, it will also work with 2019 - unfortunately, the other way around does not always work. You should use libosmscout from C/C++, Matlab and Python - optionally also C#. I can extend Matlab with C/C++ and for Python and C# I can write routines as wrappers in C/C++. Map-matching, routing and topography are the required application areas.

Currently I'm not planning any further commits, because there are still some serious bugs in CMake for which I would like to make a new patch series. For example the following parts do not fit together:

`BUILD_SHARED_LIBS` is actually the modern variant between SHARED and STATIC.
https://github.com/Framstag/libosmscout/blob/75dcc5be3895f6facf3e36ebbe574b9464fd98c3/CMakeLists.txt#L24

Depending on how `BUILD_SHARED_LIBS` is set, a project file/makefile for a static or shared library is created at this point.
https://github.com/Framstag/libosmscout/blob/75dcc5be3895f6facf3e36ebbe574b9464fd98c3/libosmscout/CMakeLists.txt#L246

With this you force the export under Windows no matter if it is a static or shared library.
https://github.com/Framstag/libosmscout/blob/75dcc5be3895f6facf3e36ebbe574b9464fd98c3/libosmscout/CMakeLists.txt#L302

Always export or nothing.
https://github.com/Framstag/libosmscout/blob/75dcc5be3895f6facf3e36ebbe574b9464fd98c3/libosmscout/include/osmscout/CoreImportExport.h#L27

Unfortunately, this schema runs through all libraries and must therefore be repaired _global_ once. I would rewrite this similar to https://gcc.gnu.org/wiki/Visibility. Also build one import-export definition for the whole project.

Framstag:
There are build errors that seem to be caused by FindDirec2D. I'll merge, once this is fixed.

I did implement the original export handling (years ago) and would claim that is uses gcc visibility features and the suggested preprocessor structure. From your quotes I do see, that symbols would be exported even if building static libraries. That should be corrected. The meson build files do handle this correctly.

But it sound like you plan further changes? Currently each library uses it own import/Export mechanism triggered by its own XXX_EXPORT_SYMBOLS define. I'm rather sure that is correct, but it sounds like you do not think so?

OgreTransporter:
> There are build errors that seem to be caused by FindDirec2D. I'll merge, once this is fixed.

I'll fix that mistake.

> But it sound like you plan further changes? Currently each library uses it own import/Export mechanism triggered by its own XXX_EXPORT_SYMBOLS define. I'm rather sure that is correct, but it sounds like you do not think so?

That is correct, but I think it's dangerous. So it is possible to add the main library dynamically and for example a renderer statically. But then it goes wrong with the linking because the target program tries to link the dependencies of the renderer (static main library) and also the dynamic main library. I would prefer an export-import macro that applies to all parts of libosmscout. You can keep your own macros for all parts (OSMSCOUT_API, OSMSCOUT_IMPORT_API, OSMSCOUT_MAP_CAIRO_API, etc.) but control it with a central switch. At POCO (https://github.com/pocoproject/poco/blob/master/Crypto/include/Poco/Crypto/Crypto.h#L71) for example, POCO_DLL applies to all libraries, but each library has its own macros. It is even easier to create a central separate header file for import-export macros and use it everywhere. Then you don't have to adapt all parts even if changes are made.

Framstag:
I'm with you that at least by default, either all libraries are build static or dynamic. It does not need to be easy to generate a mix. We do have libraries and executables that are used on a "server side" for data import and we do have libraries and executables that are used on the client. There could be reasons to build differently but in this case, some libraries would have to get build shared and static. Note also that static libraries and LGPL do have consequences. So yes, either static or shared or both consistent, every else is tricky and people are on their own. Does this help you? Do we aggree?

OgreTransporter:
Yes, I agree. I hadn't thought about the licenses, though. But that is an important point.

PR #898, opened 2020-07-26
Search transliterate

Karry:
For long time, I have been frustrated that search isn't diacritic-agnostics. When you enter pattern "jindrichuv hradec" it was not able to find "Jindřichův Hradec" town. This merge request ads new string matcher that is able to do that. It is using `iconv` ascii transliteration (there is fallback implementation when iconv is not available). It is a bit slower, but it is still possible to use current string matcher...

### DIFF
--- a/Demos/src/LocationLookup.cpp
+++ b/Demos/src/LocationLookup.cpp
@@ -43,6 +43,7 @@ struct Arguments
   size_t                 limit=30;
   size_t                 repeat=1;
   std::list<std::string> location;
+  bool                   transliterate=false;
 };
 
 bool GetAdminRegionHierachie(const osmscout::LocationServiceRef& locationService,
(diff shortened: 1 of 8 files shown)


Framstag:
Some builds fail. There seems to be test failures. Does your change correlate with the test failures?

Nice feature! Thanks!

Karry:
> Does your change correlate with the test failures?
Yes, it seems that MacOS iconv library don't support transliteration for some tested characters... Trying to hardcode en_US locale to the test...

Karry:
MacOS iconv behaves differently than gnu implementation... https://stackoverflow.com/questions/806199/how-to-fix-weird-issue-with-iconv-on-mac-os-x I added small workaround for that to get same results from search.

PR #899, opened 2020-07-28
Larger CMake update

OgreTransporter:
Update of the CMake scripts to CMake 3.10.

Correct static/shared export problems (see https://github.com/Framstag/libosmscout/pull/893#issuecomment-663995108)

Probably some corrections still need to be made. I gotta keep an eye on the CI's.

Edit: Visual Studio 2017 and 2019 are working well.


CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.10)
```
Karry, line 1: I am worry, that CMake available on Travis CI is 3.9.2

OgreTransporter: > I am not CMake expert, so I cannot comment all changes, but it looks good overall. I am fine with the changes when it doesn't break my or CI builds :-) Can you look why they are failing? Github's "MSYS using gcc and cmake" is broken for long time (tool installation issue), you may ignore that one I believe.

I'm not finished yet. There are a lot of weird bugs in the CI systems. ;-)

Karry: Ok, thanks :-) Please consider to split it to smaller isolated changes if possible...

Framstag:
Historical background:

Initially the two MSYS GitHub Actions jobs worked. After some time the both broke during MSYS setup phase. Now for some time only the cmake build breaks (during setup still, which I find strange). I researched and found bugs that hinted at possibly MSYS problems. So I hoped that build would heal after some time (MSYS update, and possibly GitHub Action for MSYS  update) automatically. I though have not activly tracked that. Also the MSYS action is some kind of hack, I hope that GitHub itself will offer a better solution sometime (haven't tracked that though becaus eof time constraints).

Yes, cmake version is an issue, we already are sideloading some packages and it is always a pain and a risk. Alternative would be to take a look for
* a newer base image available
* CI provider already offering builds tools under a different installation directory

CMake (or meson) version on the developer build system I would asume to be "near current".

And yes, build should keep working afterwards.

@OgreTransporter Please tell if I should merge. I wait for a explicit "GO!".

OgreTransporter:
> I am worry, that CMake available on Travis CI is 3.9.2

Check it out: https://travis-ci.org/github/Framstag/libosmscout/jobs/712622413#L222 ;-)
```
Unpacking cmake-data (3.10.2-1ubuntu2.18.04.1) ...
```
I chose 3.10 because I found this version to be the most current in the CI-Logs. I can also use 3.18.

> And yes, build should keep working afterwards.
> @OgreTransporter Please tell if I should merge. I wait for a explicit "GO!".

If it works I will do a rebase once and then it can be merged.

Currently there are still some mistakes that I cannot follow up.

https://ci.appveyor.com/project/Framstag/libosmscout/builds/34348397/job/s3c6vna9w0fs45jc#L718
```
C:\projects\libosmscout\libosmscout\src\osmscout\util\String.cpp(521,8): error C2664: 'size_t libiconv(libiconv_t,char **,size_t *,char **,size_t *)': cannot convert argument 2 from 'const char **' to 'char **' [C:\projects\libosmscout\build\libosmscout\OSMScout.vcxproj]
```
I have neither changed the iconv version nor this file. Why does it suddenly not work anymore?

Framstag:
> I have neither changed the iconv version nor this file. Why does it suddenly not work anymore?

The use of const or not in the signature is implementation dependend. One needs an test against the dependency during build configuration for this to be compatible on all systems (autoconv had a macro for this, see https://www.gnu.org/software/gettext/manual/html_node/AM_005fICONV.html). Why it now fails I do not know (would have to look deeper into this). Lukas worked on using iconv recently. Possibly your changes trigger better detection of iconv on Windows where it was not detected before? The builds were green after Lukas changes...

CMakeLists.txt
```diff
@@ -52,54 +87,38 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANGXX OR CMAKE_COMPILER_IS_GN
...
+    add_definitions(-ldl -lm -lpthread)
+  endif()
 endif()
-
-# Enable Large file support
-include(TestLargeFiles)
-OPJ_TEST_LARGE_FILES(OPJ_HAVE_LARGEFILES)
-
+if(APPLE AND NOT ANDROID)
+  set(CMAKE_SIZEOF_VOID_P 4)
```
cmdrf, line 95: `APPLE` and `ANDROID` are unlikely to occur at the same time, so the `if` statement doesn't seem too useful. And both have 64 bit variants, so forcing `CMAKE_SIZEOF_VOID_P` to 4 seems odd.

If this makes sense, maybe it needs a comment to explain it.

Documentation/CMakeLists.txt
```diff
@@ -52,6 +48,8 @@ add_custom_target(doc ALL
     COMMENT "Generating API documentation with Doxygen"
     VERBATIM)
 
+set_target_properties(doc PROPERTIES VERSION ${OSMSCOUT_LIBRARY_VERSION} SOVERSION ${OSMSCOUT_LIBRARY_VERSION} FOLDER "Documentation")
```
cmdrf, line 51: Does setting `VERSION` and `SOVERSION` on a custom target do anything useful? Same question for executable targets.

OgreTransporter: Useful for package generation.

CMakeLists.txt
OgreTransporter: Right, that makes no sense.

Framstag:
@OgreTransporter Looks like the builds are stabilizing now, only the iconv problem is left. Do you need help in this or will you supply a fix for this ourself?

Since the patches are huge, it would be nice to have  small summary, which conceptional changes you did. This might help keep your intentions alive :-)

OgreTransporter:
@Framstag Of course I am grateful for any help, but I will also try it myself.

Yes the patch is big. I will try to use rebase to merge the commits and re-label them.

OgreTransporter:
Unfortunately I cannot solve the iconv problem like this. Debugging does not work because the vcpkg scripts appear to be incomplete or incorrect. I.e. I cannot create the same version of iconv that AppVeyor uses. I pack the patches together with comments to merge.

```
Error: Building package libiconv:x86-windows failed with: BUILD_FAILED
Please ensure you're using the latest portfiles with `.\vcpkg update`, then
submit an issue at https://github.com/Microsoft/vcpkg/issues including:
  Package: libiconv:x86-windows
  Vcpkg version: 2020.06.15-nohash
```

Framstag:
Let's sumarize to get an overview of the current state:
* MSYS+gcc+cmake fails because of `D:/a/_temp/msys/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -ldl`
* MSYS+gcc+meson fails because of some problems in the opengl backend
* VisualStudio 2019 fails because of iconv (!)
* msys2+cmake fails because of DirectX API problems?
* The Ubuntu+cmake build fails because of -llzma failing

I can try to check the msys+gcc+meson problem regarding the opengl problems. I'll first check master...
I can also try to check the VisualStudio build, I'll also will start with master

The other error may or not may be caused by your cmake changes? The difference between the two msys+cmake build may be differences in the ci setup? The meson problem is obviously is not caused by your cmake changes ;-)

Th Ubuntu problem should be reproducible by using the (possibly slightly modified) Docker images in ci/docker?

If my analysis is wrong, please correct me, by copying and fixing the above list :-)

Framstag:
Regarding the MSYS+meson error. The import/Export attribute mismatch warnings do not occur in the master build. I assume that there is some (not so obvious) typo in the MapOpenGLImporExport.h. If I have time, I'll checkout the branch and check myself. But before I'll check for the MSVC+cmake+vcpkg problem...

Framstag:
The VS2019+cmake+vcpkg build in master works.

The by cmake generated Config.h contains

```c++
/* iconv detected */
#ifndef HAVE_ICONV
#define HAVE_ICONV 1
#endif
#ifndef ICONV_CONST
#define ICONV_CONST 
#endif
```

Framstag:
The msys+meson build can be fixed by setting OSMScoutMapOpenGL_EXPORTS in the meson.build file for the OpenGL Map backend instead of the original one (you changed the define, but missed change).

Framstag:
OK, now to the VS2019+cmake+vcpkg build in your branch.

The Config.h file in your branch contains:

```c++
/* iconv detected */
#ifndef HAVE_ICONV
#define HAVE_ICONV 1
#endif
#ifndef ICONV_CONST
#define ICONV_CONST const
#endif
```

Obviously there is a check with a different (and obviously wrong) result. I'll look further...

Framstag:
Sorry, due to the big diff, I cannot see the change that results into wrong content for the Config.h.

So, now I need help. I have a local build, so if should try something out or take a look at a special file...just give a hint!

Framstag:
Hmm...

```
Determining if the prototype iconv exists for ICONV_SECOND_ARGUMENT_IS_CONST passed with the following output:
Change Dir: C:/projects/libosmscout/build/CMakeFiles/CMakeTmp

Run Build Command(s):C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/MSBuild/Current/Bin/MSBuild.exe cmTC_aa250.vcxproj /p:Configuration=Debug /p:Platform=x64 /p:VisualStudioVersion=16.0 /v:m && Microsoft (R)-Build-Engine, Version 16.5.0+d4cbfca49 für .NET Framework

Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.



  Microsoft (R) C/C++-Optimierungscompiler Version 19.25.28612 für x64

  CheckPrototypeDefinition.c

  Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.

  cl /c /I"C:\projects\vcpkg\installed\x64-windows\include" /Zi /W3 /WX- /diagnostics:column /Od /Ob0 /D WIN32 /D _WINDOWS /D "CMAKE_INTDIR=\"Debug\"" /D _MBCS /Gm- /RTC1 /MDd /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /Fo"cmTC_aa250.dir\Debug\\" /Fd"cmTC_aa250.dir\Debug\vc142.pdb" /Gd /TC /errorReport:queue C:\projects\libosmscout\build\CMakeFiles\CMakeTmp\CheckPrototypeDefinition.c

C:\projects\libosmscout\build\CMakeFiles\CMakeTmp\CheckPrototypeDefinition.c(14,39): warning C4028: Formaler Parameter "2" unterscheidet sich von der Deklaration [C:\projects\libosmscout\build\CMakeFiles\CMakeTmp\cmTC_aa250.vcxproj]

  cmTC_aa250.vcxproj -> C:\projects\libosmscout\build\CMakeFiles\CMakeTmp\Debug\cmTC_aa250.exe
```

Should this not be an error?

OgreTransporter:
Thanks a lot for your help!

> Ah, and the Config.h is not in a private directory anymore as the old one was (which could be dangerous, since it has the same name in all libraries)

That's strange, because my Config.h are still in private directories ...
```
D:\Libosmscout\build\libosmscout\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-client-qt\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-gpx\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-import\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-map\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-map-agg\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-map-cairo\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-map-directx\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-map-opengl\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-map-qt\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-map-svg\include\osmscout\private\Config.h
D:\Libosmscout\build\libosmscout-test\include\osmscout\private\Config.h
```
The source code is located in `D:\Libosmscout\src` and the dependencies in `D:\Libosmscout\deps`.

I could build my iconv without const, but this does not solve the problem with the faulty check. I'll try to fix it.

OgreTransporter:
> Should this not be an error?

Yes! I fixed it. Let's see if it's working now.

Framstag:
Appveyor starts a build for every checkin...I have killed the older builds to speed up the result.

OgreTransporter:
So, finally it works. To the CI errors:

```
C:/Program Files (x86)/Windows Kits/10/Include/10.0.19041.0/um/winnt.h:105:10: fatal error: kernelspecs.h: No such file or Directory
```
That cannot be. If the SDK was installed correctly, the file is located in the same directory.

```
C:/Program Files (x86)/Microsoft DirectX SDK/Include/d2d1helper.h:344:15: error: expected primary-expression before 'radiusY
7654
  344 | FLOAT radiusY
7655
      | ^~~~~~~
7656
```
Maybe the DirectX SDK is too old. Meanwhile I use DirectX from the Windows SDK - which also works in the other CI builds.

And at https://ci.appveyor.com/project/Framstag/libosmscout/builds/35237941/job/y4cph9m66ujfjmpk?fullLog=true there is no error message.

Translated with www.DeepL.com/Translator (free version)

Framstag:
I'm concerted regarding the DirectX builds, since I fear that this is not an installation problem. I'll pass the patch though, since the alternatives are no fun either (patch getting even bigger, difficult to help, possibly down spiraling motivation...). I'll check the MSYS/cmake build afterwards locally and will give feedback for this.
It would be nice, if you could check things afterwards and see if you can fix/improve the DirectX situation.
IT would also helpful I you give some concreted hints regarding to changes. You could enhance the Hugo based documentation for example for example, which is simply markup based (and I know, documentation is the not fun part of programming).

Framstag:
I can reproduce the MSYS DirectX build problem on my system. The include path it uses is C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um and actually points to a existing directory.

Build fails with:

```
[ 44%] Building CXX object libosmscout-map-directx/CMakeFiles/OSMScoutMapDirectX.dir/src/osmscout/MapPainterDirectX.cpp.obj
In file included from C:/projects/libosmscout/libosmscout/include/osmscout/util/FileScanner.h:42,
                 from C:/projects/libosmscout/libosmscout/include/osmscout/TypeFeature.h:30,
                 from C:/projects/libosmscout/libosmscout/include/osmscout/TypeConfig.h:35,
                 from C:/projects/libosmscout/libosmscout/include/osmscout/TypeFeatures.h:27,
                 from C:/projects/libosmscout/libosmscout-map/include/osmscout/MapPainter.h:30,
                 from C:/projects/libosmscout/libosmscout-map-directx/include/osmscout/MapPainterDirectX.h:29,
                 from C:/projects/libosmscout/libosmscout-map-directx/src/osmscout/MapPainterDirectX.cpp:20:
C:/Program Files (x86)/Windows Kits/10/Include/10.0.18362.0/um/windows.h:31: warning: ignoring '#pragma region Application' [-Wunknown-pragmas]
   31 | #pragma region Application Family or OneCore Family
      |
In file included from C:/msys64/mingw64/x86_64-w64-mingw32/include/minwindef.h:163,
                 from C:/msys64/mingw64/x86_64-w64-mingw32/include/windef.h:8,
                 from C:/Program Files (x86)/Windows Kits/10/Include/10.0.18362.0/um/windows.h:171,
                 from C:/projects/libosmscout/libosmscout/include/osmscout/util/FileScanner.h:42,
                 from C:/projects/libosmscout/libosmscout/include/osmscout/TypeFeature.h:30,
                 from C:/projects/libosmscout/libosmscout/include/osmscout/TypeConfig.h:35,
                 from C:/projects/libosmscout/libosmscout/include/osmscout/TypeFeatures.h:27,
                 from C:/projects/libosmscout/libosmscout-map/include/osmscout/MapPainter.h:30,
                 from C:/projects/libosmscout/libosmscout-map-directx/include/osmscout/MapPainterDirectX.h:29,
                 from C:/projects/libosmscout/libosmscout-map-directx/src/osmscout/MapPainterDirectX.cpp:20:
C:/Program Files (x86)/Windows Kits/10/Include/10.0.18362.0/um/winnt.h:105:10: fatal error: kernelspecs.h: No such file or directory
  105 | #include <kernelspecs.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [libosmscout-map-directx/CMakeFiles/OSMScoutMapDirectX.dir/build.make:83: libosmscout-map-directx/CMakeFiles/OSMScoutMapDirectX.dir/src/osmscout/MapPainterDirectX.cpp.obj] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:785: libosmscout-map-directx/CMakeFiles/OSMScoutMapDirectX.dir/all] Fehler 2
make: *** [Makefile:161: all] Fehler 2
```

This is the compiler call meson uses:

```
build libosmscout-map-directx/libosmscout_map_directx.dll.p/src_osmscout_MapPainterDirectX.cpp.obj: cpp_COMPILER ../libosmscout-map-directx/src/osmscout/MapPainterDirectX.cpp
 DEPFILE = "libosmscout-map-directx/libosmscout_map_directx.dll.p/src_osmscout_MapPainterDirectX.cpp.obj.d"
 DEPFILE_UNQUOTED = libosmscout-map-directx/libosmscout_map_directx.dll.p/src_osmscout_MapPainterDirectX.cpp.obj.d
 ARGS = "-Ilibosmscout-map-directx/libosmscout_map_directx.dll.p" "-Ilibosmscout-map-directx" "-I../libosmscout-map-directx" "-Ilibosmscout-map-directx/include" "-I../libosmscout-map-directx/include" "-Ilibosmscout-map/include" "-I../libosmscout-map/include" "-Ilibosmscout/include" "-I../libosmscout/include" "-fdiagnostics-color=always" "-pipe" "-D_FILE_OFFSET_BITS=64" "-Wall" "-Winvalid-pch" "-Wnon-virtual-dtor" "-Wextra" "-Wpedantic" "-std=c++17" "-O2" "-g" "-D_USE_MATH_DEFINES" "-pthread" "-D_UNICODE" "-DOSMScoutMapDirectX_EXPORTS" "-fvisibility=hidden"
```

IMHO the answer is, that MINGW/MSYS comes with its own DirectX headers:

```
bash-4.4$ find /mingw64/ -name "d2d1.h"
/mingw64/x86_64-w64-mingw32/include/d2d1.h
```

so in this cases the cmake build files seem to do more than they should? Can you supply a fix?

Framstag:
Have now made https://github.com/Framstag/libosmscout/issues/910 for this.

PR #902, opened 2020-08-13
setup minimum supported version of Qt version to 5.6

Karry:
It should be safe to merge. Version 5.6 is pretty old and it was there before recent CMake change. @OgreTransporter which Qt version are you using for your build?

### DIFF
--- a/cmake/features.cmake
+++ b/cmake/features.cmake
@@ -126,7 +126,7 @@ find_package(OpenGL)
 find_package(GLEW)
 find_package(GLM)
 find_package(GLFW)
-find_package(Qt5 COMPONENTS Core Gui Widgets Qml Quick Svg Positioning Multimedia LinguistTools)
+find_package(Qt5 5.6 COMPONENTS Core Gui Widgets Qml Quick Svg Positioning Multimedia LinguistTools)
 if(Qt5_FOUND AND WIN32 AND TARGET Qt5::qmake AND NOT TARGET Qt5::windeployqt)
   get_target_property(_qt5_qmake_location Qt5::qmake IMPORTED_LOCATION)
   execute_process(


OgreTransporter:
I use 5.14.2. I would not set a Qt version or use a version that is either active or marked as LTS. The support end of Qt 5.6 was on 2019-03-16. So I would use at least 5.12 (end of life 2021-12-05) or set to the current LTS version 5.15 (2023-05-26).

Karry:
I am stuck with Qt 5.6 on Sailfish OS platform unfortunately :-( Newer versions of Qt are licensed under terms GPLv3, that guarantee users possibility to update such library on their devices... For Jolla it is problematic, when they provides certified system for governments. Their lawyers needs to solve this issue first (question is who is user, if government that owns the device or employee).

OgreTransporter:
Qt 5 is licensed under both GPL 3 and LGPL 3. Alternatively you can buy a commercial license. The difference between 5.6 (https://doc.qt.io/archives/qt-5.6/licensing.html) and 5.15 (https://doc.qt.io/qt-5/licensing.html) is that 5.6 was also licensed under LGPL 2.1. So I don't understand your problem.

Karry:
You are right. It is LGPL v2 vs. v3. But problem is still the same. You cannot sell locked device which contains software licensed under terms of LGPLv3. And Jolla need to have locked device when they want certification from some governments. Community is pushing Jolla to solve this issue, but it seems that it will not be solved soon :-(

OgreTransporter:
Sounds familiar. Many companies and also some projects don't worry about version and license upgrades until it is already too late. But if you don't want to update, the security of the customers doesn't matter. For example, BigBlueButton is only built for Ubuntu 16.04, which expires at the end of the year. All requests for upgrades were rejected with the statement "16.04 is stable and reliable". Now they are trying to upgrade to 18.04 in a hurry.

PR #909, opened 2020-09-19
Use std::sort parallel version

Framstag:
Add some code that makes use of std parallel sort. Added TBB library to get it to work for gcc. Let's see what the other environments say.

Could be a way to get rid of omp code.


Framstag:
Apple clang doe snot yet support the parallel extensions of c++17. See https://en.cppreference.com/w/cpp/compiler_support#cpp17

Framstag:
@Karry Does this compile in your Sailfish environment? Currently only Mac OS/iOS cannot compile this.

Karry:
It fails in Sailfish OS SDK unfortunately. It is using GCC 8.3.0 that don't have `<execution>` header. But you can merge it. I may patch it in branch that I am using...

Framstag:
I also fails on iOS and Mac OS. If we use the new feature, we will likely spreed them all over the place, possibly even remove some or all OMP usage. I do not think, that you want to path this "away" everytime. This would also result in some performance lost.

IMHO we can only add an check and make this n optional performance boost and do not change places where a performance loss would be intolerable.

OgreTransporter:
You could write a switch in the configuration header to enable or disable this feature. I can quickly add a corresponding selection option to CMake. What is TBB needed for? Up to now this is not used in the code, only the C++17 features are used. If you can do without TBB, I am for it. The library already has enough dependencies, there is no need for more.

Framstag:
> You could write a switch in the configuration header to enable or disable this feature. I can quickly add a corresponding selection option to CMake. What is TBB needed for? Up to now this is not used in the code, only the C++17 features are used. If you can do without TBB, I am for it. The library already has enough dependencies, there is no need for more.

I already did on the meson side (via master push by mistake). libosmscout-import defines HAVE_STD_EXECUTION if tbb is available and the <execution> header is available. Strictly only gcc uses tbb to implement the execution header functionality.

I currently only target the importer to improve performance (though code has the get refactored first to make use of <execution>), though the map library of course could also make use of the execution header.

Framstag:
Most of the code has been pushed by other pull requests. Only the updates to more recent compilers were not merged, but that can wait...

Issue #870, opened 2020-05-24
Meson unity build fails


Framstag:
Meson supports unity builds, allowing to join all source files and build them in one go. This can reduce compile times since header parsing is reduced (and possibly also allows some more compiler optimisations).

Currently the unity build fails:

```
FAILED: libosmscout-client-qt/068bd2a@@osmscout_client_qt@sha/meson-generated_osmscout_client_qt-unity24.cpp.o 
[...]

PR #911, opened 2020-09-20
Unity

Framstag:
Try to get meson unity builds running. Should be faster, allows more compiler optimizations.


Framstag:
Fix for https://github.com/Framstag/libosmscout/issues/870

PR #914, opened 2020-09-21
Small route fix, subway route rendering

Karry:
Hi. 

Here is small fix that was causing troubles during data loading... And initial support for routes in public transport stylesheet. 

Here are Prague subways:

obrazek

I will try to add tram and bus routes later (evening today?)...


Framstag:
That looks really cool. What do I have to do, to get such rendering? Is it part of the default style sheet?

Karry:
> Is it part of the default style sheet?

Nope. There is alternative stylesheet: `stylesheets/public-transport.oss`

PR #920, opened 2020-09-24
Route redering performance

Karry:
Hi. Real-live usage shows that route preparison for rendering is slow. Perf shows that performance problem is in `map` copying... This change speedup rendering 10x ;-) It is still 25% slower than without routes... But usable. I want to use routes just in specialized stylesheets: public transport and hiking/cycling.


Framstag:
I suggest to write a blog article about this, where the blog appear son the OSM blog list :-)

Issue #923, opened 2020-09-28
Missing headers


OgreTransporter:
```
1>D:\libosmscout\src\osmscout\util\Logger.cpp(121): error C2679: Binärer Operator "<<": Es konnte kein Operator gefunden werden, der einen rechtsseitigen Operanden vom Typ "const std::string_view" akzeptiert (oder keine geeignete Konvertierung möglich)
1>D:\libosmscout\include\osmscout/Database.h(422): error C2039: "string_view_literals": Ist kein Element von "std" (Quelldatei wird kompiliert D:\libosmscout\src\osmscout\routing\MultiDBRoutingService.cpp)
1>D:\libosmscout\include\osmscout/Database.h(422): error C2871: "string_view_literals": Es ist kein Namespace mit diesem Namen vorhanden (Quelldatei wird kompiliert D:\libosmscout\src\osmscout\routing\MultiDBRoutingService.cpp)
1>D:\libosmscout\include\osmscout/Database.h(423): error C3688: ungültiges Literalsuffix "sv"; Literaloperator oder Literaloperator-Vorlage "operator ""sv" nicht gefunden (Quelldatei wird kompiliert D:\libosmscout\src\osmscout\routing\MultiDBRoutingService.cpp)
1>D:\libosmscout\include\osmscout/Database.h(423): error C2672: "osmscout::Database::GetObjectsByOffset": keine übereinstimmende überladene Funktion gefunden (Quelldatei wird kompiliert D:\libosmscout\src\osmscout\routing\MultiDBRoutingService.cpp)
1>D:\libosmscout\include\osmscout/Database.h(423): error C2780: "bool osmscout::Database::GetObjectsByOffset(DataFile,const OffsetsCol &,DataCol &,const std::string_view &) const": Erwartet 4 Argumente - 3 unterstützt (Quelldatei wird kompiliert D:\libosmscout\src\osmscout\routing\MultiDBRoutingService.cpp)
1>D:\libosmscout\include\osmscout/Database.h(405): error C2039: "string_view_literals": Ist kein Element von "std" (Quelldatei wird kompiliert D:\libosmscout\src\osmscout\routing\MultiDBRoutingService.cpp)
1>D:\libosmscout\include\osmscout/Database.h(405): error C2871: "string_view_literals": Es ist kein Namespace mit diesem Namen vorhanden (Quelldatei wird kompiliert D:\libosmscout\src\osmscout\routing\MultiDBRoutingService.cpp)
[...]

Issue #879, opened 2020-06-06
Qt 5.15 deprecations


Framstag:
Qt 5.15 seems to have new deprecations that result in compiler warnings

```
QPoint QWheelEvent::pos() const
bool QVariant::operator>(const QVariant&) const
int qrand()
```

OgreTransporter:
You missed one ;-)

`/Users/travis/build/Framstag/libosmscout/libosmscout-client-qt/src/osmscout/MapWidget.cpp:108:65: warning: 'QFlags' is deprecated`

Unfortunately I have not (yet) been able to find a fix for this.

Framstag:
Currently no warnings anymore :-)

Issue #310, opened 2017-05-07
Rewrite tests to make use of an existing tets framework


Framstag:
I would suggest to use https://github.com/philsquared/Catch:
* Looks good
* Allows to use it with one header copied into the project

Framstag:
Added catch header. Rewrote 3 tests. More have to follow. Since verybody can now convert further tests, I'm givingh away ownership of the issue.

Framstag:
Test still likely to get transformed to new test framework:
* EncodeNumber.cpp
* FileScannerWriter.cpp
* GeoBox.cpp
* GeoCoordParse.cpp
* Geometry.cpp
* MapRotate.cpp
* NumberSet.cpp
* OSTAndOSSCheck.cpp
* ScanConversion.cpp
* TransPolygon.cpp
* WStringStringConversion.cpp

Others tests are designed to be called manually or have no success/fail criteria or just measure performance.

Framstag:
Updated list:

* EncodeNumber.cpp
* FileScannerWriter.cpp
* GeoCoordParse.cpp
* MapRotate.cpp
* NumberSet.cpp
* OSTAndOSSCheck.cpp
* TransPolygon.cpp
* WStringStringConversion.cpp

OgreTransporter:
I looked at the code of the tests and found even more tests without catch framework. Or did I do something wrong?
 - CachePerformance
 - CalculateResolution
 - ClientQtThreading
 - CoordinateEncoding
 - DrawTextQt
 - EncodeNumber
 - FeatureLabelTest
 - FileScannerWriter
 - GeoCoordParse
 - LocationLookupTest
 - MapRotate
 - MultiDBRouting
 - NumberSet
 - NumberSetPerformance
 - OSTAndOSSCheck
 - QtFileDownloader
 - ReaderScannerPerformance
 - ThreadedDatabase
 - TransPolygon
 - WorkQueue
 - WStringStringConversion

Framstag:
There are some tests, that act as classical automatic test suites because they do not have a defined result but act as base for manual tests to see if thing have degraded (like e.g. CachePerformance).

I appreciate that you have taken the time to move tests to catch. This is "no fun" work. Thank you!

Note that some tests will fail under MSYS, because of limitations of MSYS - like not having proper locale support. Because of this tests are not executed under all platforms - which is something we should fix. Exclude only those that under a platform that are known to fail and enable everything else.

OgreTransporter:
[...]

PR #924, opened 2020-09-28
Bugfixes

OgreTransporter:
**Fix missing headers for std::string_view (issue #923)**
The GCC and CLANG forgive much easier if you forget headers from the standard library. MSVC is very strict about this. Header `<string_view>` is required to compile with VS2017 (https://en.cppreference.com/w/cpp/string/basic_string_view/operator%22%22sv).

**Resolve 2/3 Qt deprecation warnings (issue #879)**
Somewhere we already had the discussion about the QT version. As far as I'm concerned, you can assume 5.14, but on some old systems there is only QT 5.6 for licensing reasons, so the old code is still included. I could not (yet) find the problem `bool QVariant::operator>(const QVariant&) const`.

**Update Tests (issue #310)**
I have switched some tests to the catch framework.

### DIFF
--- a/Tests/meson.build
+++ b/Tests/meson.build
@@ -77,7 +77,7 @@ endif
 
 EncodeNumber = executable('EncodeNumber',
              'src/EncodeNumber.cpp',
-             include_directories: [osmscoutIncDir],
+             include_directories: [testIncDir, osmscoutIncDir],
              dependencies: [mathDep, openmpDep],
              link_with: [osmscout],
              install: false)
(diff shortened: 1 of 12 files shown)


Framstag:
Regarding the meson based builds. The tests that (up to know) did not need catch.hh did not add "testIncDir" as include dir. In the meson build ou have to change for them

```
             include_directories: [osmscoutIncDir],
```

to

```
             include_directories: [testIncDir, osmscoutIncDir],
```

Karry (changes requested):
After fast review, it looks good. But Meson - based build is broken. Make sure that you add `testIncDir` (with catch.hpp) to all tests that are using Catch...

OgreTransporter:
Sorry, I did not think of meason. With CMake you should find the right directory everywhere in it. I will correct this later.

Karry (approved):
LGTM. Meson build is working now. But `ninja test` executes 33 tests, and `ctest` 37. So, some tests are missing in Meson build ;-) It may be solved later...

PR #944, opened 2020-10-13
Android qt dirty

Karry:


Framstag:
Interesting idea, though it still does not seem to detect the Qt installation. i just heard in an older podcast, that the Debian Qt maintainer stopped in the wake of Qt 6 ;-)

PR #947, opened 2020-10-14
allow raw html (tables, extended formating...) in webpage markdown

Karry:


Karry:
It seems that webpage don't contains raw html (tables) from Hugo upgrade at 29.8.2020 ( 9c3cbe9 ) http://libosmscout.sourceforge.net/features/rendering/

Framstag:
I have seen that problem at the time I hinted @OgreTransporter at the documentation - but did not have time to look at. Thanks for doing that!

PR #964, opened 2020-10-24
skip preparison of way labels/shields when there is no style for it

Karry:
Rendering should be faster when there is no way labels or way shields.

With standard stylesheets, there are no way labels until zoom level 12. But during my performance testing (used Prague area), zoom level 11 contained 14145 ways and there was no measurable improvement...

What do you think? Is code change simple enough to accept it even when there is no measurable improvement?

### DIFF
--- a/libosmscout-map/include/osmscout/StyleConfig.h
+++ b/libosmscout-map/include/osmscout/StyleConfig.h
@@ -760,8 +760,21 @@ namespace osmscout {
                                std::vector<PathSymbolStyleRef> &symbolStyles) const;
     PathTextStyleRef GetWayPathTextStyle(const FeatureValueBuffer& buffer,
                                          const Projection& projection) const;
+
+    /**
+     * @param projection
+     * @return true when some way text style is defined on provided projection
+     */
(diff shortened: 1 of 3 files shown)


Framstag:
For me it is OK. We have the general problem (as I see it) that optimisations are not measurable under Desktop situations. I recently read an article about optimisation that are not measurable under desktop but have measurable (and non-neglectable) impact under Raspberry PI. Did you make measurements on mobile?

This raises question how we in principle can measure performance optimisations at all?

Karry:
Yes, I measured it with release build (`-O2`) on my phone :-) I believe that this optimization will have effect with style that render a lot of ways, but no way labels... It is not the case for our standard stylesheet. For low zoom levels is amount of ways relatively small...

> This raises question how we in principle can measure performance optimisations at all?

It is true that results from x86_64 may be significantly different than arm* platforms. Arm don't like unaligned memory access, some chips has slow unit for float computations, jump prediction is limited... For manual comparisons, I am running slightly modified `PerformanceTest` demo on my phone... But for some continuous measurements, it would require some dedicated hardware...

Issue #968, opened 2020-10-31
LocationIndex performance issues


Karry:
Location index is written to save memory at runtime. It is not used whole application live time usually (it was the motivation probably). It don't cache loaded entries and file is opened with every method call. For example [`LocationIndex::VisitLocations`](https://github.com/Framstag/libosmscout/blob/master/libosmscout/src/osmscout/LocationIndex.cpp#L666). 

But this implementation has performance issues in some cases. For example when DescribeLocation is called for place in Los Angeles. Following command takes 4 minutes on x86, and ages on mobile phone...
```
./Demos/LocationDescription \
  ~/Maps/north-america-us-california-20201016-061601 \
  33.9417 -118.26719
```

Here is flame graph from previous command:
![Screenshot_20201031_124833](https://user-images.githubusercontent.com/309458/97778438-6edd7f00-1b77-11eb-8b60-98548b8139ab.png)

Two highlights:
 - 18% of the time is consumed in `osmscout::FileScanner::Open`
 - 46% of the time is consumed in copying of objects

I will try to address the most expensive issue...

Karry:
I made small mistake in measurement. I was using debug build on x86. But flamegraph is similar... With release build:

 - original code takes 1:39
 - with recursive visitor 54 seconds
 - with scanner pool 1:01
 - with both optimisations 1.2 seconds ;-)

Karry:
Closing this issue. Reverse lookup in Los Angeles takes ~5 seconds on mobile now.

PR #971, opened 2020-11-01
filter candidates for reverse lookup

Karry:
Improves https://github.com/Framstag/libosmscout/issues/968 by skipping candidates that cannot be location index. If I understand it correctly, location index is composed from nodes processed in `NodeLocationProcessorFilter` and areas in `AreaLocationProcessorFilter`.

There may be just:

 - objects with address feature
 - POIs with name feature

### DIFF
--- a/libosmscout/include/osmscout/LocationDescriptionService.h
+++ b/libosmscout/include/osmscout/LocationDescriptionService.h
@@ -383,13 +383,17 @@ namespace osmscout {
 
     void AddToCandidates(std::vector<LocationDescriptionCandicate>& candidates,
                          const GeoCoord& location,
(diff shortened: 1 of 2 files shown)


Framstag:
I do not like the method API growing by a (rather huge) number of parameters. I haven't checked the code in details, so the more general question. Doe sit make sense to encapsulate the objects into lightweight context or parameter objects to get nicer and better APIs?

Karry:
I understand your concerns. But in this case, I can imagine just filtering candidates later, or add some "filter" lambda as a argument. But it would complicate usage more IMHO. Luckily, it is not public API, `AddToCandidates` is private...

PR #973, opened 2020-11-01
add util class that cleanup LocationIndex cache when destructed

Karry:
...and use it in LocationService and LocationDescriptionService.
That way, it is not necessary to explicitly cleanup cache
before every function exit point and it keep memory footprint
small when location index is not used.

___

I am still not sure if it makes sense to try cleanup location index cache (file scanner pool) after every usage, or keep it to application to decide when to cleanup cache and free resources...


Framstag:
If it is unclear what the user/developer want, I would suggest to add a callback which 1 default implementation, too. This way, the user can implement its own version (if the callback get enough information for this) and the default code is not cluttered with alternatives and options.

Framstag:
Looks like the other merges resulted in merge conflicts :-/

PR #979, opened 2020-11-03
fix return type of extectedSize method

Karry:
size_t is 32 bit unsigned number on 32 bit architectures,
this is not enough for when download job has more than 4 GiB


Framstag:
I suggest to use FileOffset as a datatype for this, but will pass the merge anyway...

Issue #63, opened 2016-07-04
Write style sheet analysis tool


Framstag:
Manual style sheet optimizations show, that an explicit tool may be helpful:

Features of the tool:
* Analysis of actual data: Which types in the database will be visible at which zoom level. If style will try to draw the type from a lower zoom level, output some warning. Example: Do not draw buildings at state level, they will not be visible at all.
* Analyse for each zoom level the amount of nodes drawn to visualize data by type.  Hint at types, that have many nodes and are expensive to render.

PR #983, opened 2020-11-19
add initial stylesheet analysis to OSTAndOSSCheck

Karry:
When I am playing with map stylesheet, sometimes found a issue that may be detected automatically by tool. For example when important label visible on low zoom disappear on higher zoom because of another label... Tool should check that labels that are visible on lower zoom levels should have higher priority. But we don't have such tool unfortunately. Yet :-) There is ticket with similar idea - even include real database data to analysis https://github.com/Framstag/libosmscout/issues/63 . 

This commit extends existing `OSTAndOSSCheck` tool by `--analyze` option. I just show database types and zoom levels when these types are visible... Later, we may extend this test or move code to separate code. Or Stylesheet editor possibly. Not sure.

Screenshot_20201119_090432

### DIFF
--- a/Tests/src/OSTAndOSSCheck.cpp
+++ b/Tests/src/OSTAndOSSCheck.cpp
@@ -26,12 +26,147 @@
 
 #include <osmscout/util/File.h>
 #include <osmscout/util/CmdLineParsing.h>
+#include <iomanip>
 
 using namespace std;
+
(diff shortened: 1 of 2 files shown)


Framstag:
Exactly such things I had in mind. I suggest to move that part to top level command line tool, but will pass the merge anyway.

PR #988, opened 2020-12-09
Fixed Qt deprecation warnings

Framstag:


Karry:
Hi. `QNetworkRequest::RedirectPolicyAttribute` was introduced in Qt 5.9. It cannot be used on SailfishOS unfortunately. I am still stuck with Qt 5.6. Only possibility would be usage of some conditional macro for me :-(

Framstag:
Damned :-/ I was not aware that your version is _that_ old. I got deprecation warnings under 5.15.2, preparing the code for Qt 6.0 (which is out now). Can you extend the patch with the necessary conditionals?

vyskocil:
If I’m right Qt had big licence changes impacting commercial code when using the free version around that version...

Framstag:
I haven't checked for license restrictions changes yet. Though IMHO for Qt5 there were already restrictions at place for commercial applications. I have the statement in mind, that if you handle all deprecations in Qt5 you code likely already will run under Qt6.

Questions is how Linux distributions will switch to Qt 6 and when.

Karry:
Conditionals are there... 

Jolla (company behind Sailfish OS) has multiple problems. Their team is relatively small, so many libraries in this OS is very old. And Qt recently (2016 ? :-) ) starts transition from LGPLv2 to LGPLv3 in free edition. (L)GPL version three adds guarantee to user that he may build the software for his own (sources are available), but also upgrade that software on the device! And there is problem. Jolla's biggest consumer is Russia government and Sailfish powered devices are used on the state institutions, for example post. And administrators for these institutions don't want to allow employees to reinstall core libraries on mobile devices. So, if I understand the issue, there is question for the lawyers: who is user from (L)GPL perspective? Government who owns the devices or its employees who use them?

But I am optimistic. Support of Qt 5.6 ended last year and Jolla will have to upgrade soon...

Karry (approved):

PR #1006, opened 2021-02-07
import leisure=fitness_station as leisure_pitch

Karry:
What you think about merging [fitness_station leisure](https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dfitness_station) type to existing one - pitch? Or you would rather create new type for it?

### DIFF
--- a/stylesheets/map.ost
+++ b/stylesheets/map.ost
@@ -1003,7 +1003,7 @@ TYPES
       PATH[FOOT] PIN_WAY OPTIMIZE_LOW_ZOOM
 
   TYPE leisure_pitch
-    = NODE AREA ("leisure"=="pitch")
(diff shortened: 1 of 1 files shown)


Framstag:
Hmm, compare

https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dpitch with https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dfitness_station

IMHO they are not the same, but of course in a rough version similar enough to visualize.

So different types but similar visualization?

To avoid type explosion I once though of have rather simple generic types and do style sheet differentiation via Features. I have however never explored this in deep.

In the end there is complexity and you have to distribute it in some way over types, features etc... so that it is a "best fit".

Framstag:
Obsolete because of #1008 ?

Karry:
yes :-)

PR #1007, opened 2021-02-07
Routing tests

Karry:
I added possibility to dump route to json file and prepared simple bash script that verify if route computation was changed. I want to add more testing routes later and use it to verify how routing tunning change existing routes. Current routing is not ideal but I need to be sure that my changes don't makes routing worse.


Framstag:
Hi @Karry Please explicitly tell me to merge, if the merge is final :-)

Karry:
The merge is final :-) I will try to optimise routing later.

Issue #962, opened 2020-10-23
DirectX character encoding


OgreTransporter:
The DirectX renderer ist not working with non-ascii characters. It should be possible to fix, because GDI+ is working ;-)

ghpl210312:
I had a quick look into the DirectX-renderer and I changed the lines 130 to 140 in MapPainterDirectX.cpp as follows:

	std::wstring text2 = UTF8StringToWString(text);
	double horizontalOffset = 0;
	//for (size_t ch = 0; ch < text.length(); ch++) {
	for (size_t ch = 0; ch < text2.length(); ch++) {
		result.emplace_back();

#ifdef MBUC
		//std::wstring sample = s2w(text.substr(ch,1));
		std::wstring sample = text2.substr(ch,1);

That did the trick for me.

Karry:
nice :-) may you send a merge request please?

ghpl210312:
Hi,
I'm completely new to github and despite of reading various READMEs I am totally lost.
I have checked out the project with SVN. Do I have to create a branch from the trunk, apply the changes to it and commit?
Sorry.

Karry:
Ok then, I prepared merge request. But I have no Windows machine with setup development environment. Can you review it and test it please?

Thank you!

ghpl210312:
Sure, I will do.
Thank you.

PR #1012, opened 2021-03-15
fix DirectX character encoding

Karry:
try to fix https://github.com/Framstag/libosmscout/issues/962

@ghpl210312 may I ask for review please? I have no windows machine.

### DIFF
--- a/libosmscout-map-directx/src/osmscout/MapPainterDirectX.cpp
+++ b/libosmscout-map-directx/src/osmscout/MapPainterDirectX.cpp
@@ -131,14 +131,15 @@ namespace osmscout
     std::vector<Glyph<MapPainterDirectX::DirectXNativeGlyph>> result;
 
     double horizontalOffset = 0;
-    for (size_t ch = 0; ch < text.length(); ch++){
-      result.emplace_back();
-
 #ifdef MBUC
-      std::wstring sample = s2w(text.substr(ch,1));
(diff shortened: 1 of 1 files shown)


Framstag:
@ghpl210312

As you can see from the CI/CD builds, you should be able to retrieve the patch by:

```
git clone -q https://github.com/Framstag/libosmscout.git C:\projects\libosmscout
git fetch -q origin +refs/pull/1012/merge:
git checkout -qf FETCH_HEAD
Restoring build cache
```

Karry:
No response from @ghpl210312 :-( I would merge it @Framstag , builds are green and this change should fix described issue...

ghpl210312:
Hi guys,
I am really sorry, but I my work keeps me quite busy at the moment.
I tested it only with my version, but these lines from MapPainterDirectX do the same:

#ifdef MBUC
    std::wstring wText = UTF8StringToWString(text);
    for (size_t ch = 0; ch < wText.length(); ch++) {
      std::wstring sample = wText.substr(ch,1);
#else
    for (size_t ch = 0; ch < text.length(); ch++) {
      std::string sample = text.substr(ch,1);
#endif

Framstag:
I'm not sure that this is the optimal solution, but I see that it might improve the situation. I in principle can test it myself but also lack time, Thus I'll merge it now :-)

PR #1013, opened 2021-03-20
Routing parameters demo

Karry:
Hi. I am not satisfied with routing often, and tuning routing profile in c++ code is slow. Moreover, I would like to have parametrized routing in application. For example have checkbox for avoiding main highways on bicycle... So, I made few changes:

 - allow to specify different speed by route grade in routing profile
 - add QML type for routing profile that allows customize routing (from QML UI)
 - add QML demo for simple experiments with route profiles

Screenshot_20210320_173516


Karry:
Thank you :+1:

PR #1014, opened 2021-03-21
fix crash SIGSEGV caused by referencing deleted iterator in iteration

janbar:
Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x8 in tid 28485 (qtMainLoopThrea), pid 28463
(QScopedPointer<QObjectData, QScopedPointerDeleter<QObjectData>>::operator->() const+12)
(QObject::parent() const+28)
(osmscout::LocationEntry::LocationEntry(osmscout::LocationEntry const&)+80)
(osmscout::LocationListModel::onSearchResult(QString, QList<osmscout::LocationEntry>)+1552)

Platform: Qt5.12.9, arm64

The fix erase merged locations outside of the loop

### DIFF
--- a/libosmscout-client-qt/src/osmscout/NearPOIModel.cpp
+++ b/libosmscout-client-qt/src/osmscout/NearPOIModel.cpp
@@ -43,12 +43,6 @@ NearPOIModel::NearPOIModel()
 
 NearPOIModel::~NearPOIModel()
 {
-  for (QList<LocationEntry*>::iterator location=locations.begin();
-       location!=locations.end();
-       ++location) {
-      delete *location;
-  }
-
   locations.clear();
 
   if (breaker){
@@ -68,7 +62,7 @@ QVariant NearPOIModel::data(const QModelIndex &index, int role) const
     return QVariant();
   }
 
-  LocationEntry* location=locations.at(index.row());
+  LocationEntryRef location=locations.at(index.row());
 
   switch (role) {
     case Qt::DisplayRole:
@@ -139,7 +133,7 @@ QObject* NearPOIModel::get(int row) const
     return nullptr;
   }
 
-  LocationEntry* location=locations.at(row);
+  LocationEntryRef location=locations.at(row);
   // QML will take ownership
   return new LocationEntry(*location);
 }
@@ -169,7 +163,7 @@ void NearPOIModel::onLookupResult(int requestId, QList<LocationEntry> newLocatio
     if (distance > maxDistance){
       continue;
     }
-    for (LocationEntry* secondLocation:locations) {
+    for (const LocationEntryRef& secondLocation:qAsConst(locations)) {
       if (distance < osmscout::GetSphericalDistance(secondLocation->getCoord(), searchCenter)){
         break;
       }
@@ -182,7 +176,7 @@ void NearPOIModel::onLookupResult(int requestId, QList<LocationEntry> newLocatio
     }
 
     emit beginInsertRows(QModelIndex(), position, position);
-    locations.insert(position, new LocationEntry(location));
+    locations.insert(position, std::make_shared<LocationEntry>(location));
     qDebug() << "Put " << location.getObjectType() << location.getLabel() << " to position: " << position << "(distance" << distance.AsMeter() << " m)";
     emit endInsertRows();
   }
@@ -196,12 +190,6 @@ void NearPOIModel::lookupPOI()
 
   if (!locations.isEmpty()){
     beginResetModel();
-    for (QList<LocationEntry*>::iterator location=locations.begin();
-         location!=locations.end();
-         ++location) {
-      delete *location;
-    }
-
     locations.clear();
     endResetModel();
     emit countChanged(locations.size());
--- a/libosmscout-client-qt/src/osmscout/SearchLocationModel.cpp
+++ b/libosmscout-client-qt/src/osmscout/SearchLocationModel.cpp
@@ -64,12 +64,6 @@ LocationListModel::LocationListModel(QObject* parent)
 
 LocationListModel::~LocationListModel()
 {
-  for (QList<LocationEntry*>::iterator location=locations.begin();
-       location!=locations.end();
-       ++location) {
-      delete *location;
-  }
-
   locations.clear();
 
   if (breaker){
@@ -102,19 +96,23 @@ void LocationListModel::onSearchResult(const QString searchPattern,
   int equalityCall=0;
   int comparisonsCall=0;
 
-  QList<LocationEntry> foundLocations = foundLocationsConst;
+  QList<LocationEntryRef> foundLocations;
+  for (const LocationEntry& e : foundLocationsConst)
+    foundLocations.push_back(std::make_shared<LocationEntry>(e));
+
   QQmlEngine *engine = qmlEngine(this);
   if (equalsFn.isCallable()){
     // try to merge locations that are equals
-    auto Equal = [&](LocationEntry &a, const LocationEntry &b) -> bool {
-      if (a.getLabel()==b.getLabel() &&
-          a.getDatabase()==b.getDatabase() &&
-          a.getType()==LocationEntry::typeObject &&
-          b.getType()==LocationEntry::typeObject){
+    auto Equal = [&](const LocationEntryRef& a, const LocationEntryRef& b) -> bool {
+      assert(a != nullptr && a->parent() == nullptr && b != nullptr && b->parent() == nullptr);
+      if (a->getLabel()==b->getLabel() &&
+          a->getDatabase()==b->getDatabase() &&
+          a->getType()==LocationEntry::typeObject &&
+          b->getType()==LocationEntry::typeObject){
 
         QJSValueList args;
-        args << engine->newQObject(new LocationEntry(a));
-        args << engine->newQObject(new LocationEntry(b));
+        args << engine->newQObject(new LocationEntry(*a));
+        args << engine->newQObject(new LocationEntry(*b));
         QJSValue result = equalsFn.call(args);
         equalityCall++;
         if (result.isError()){
@@ -133,7 +131,7 @@ void LocationListModel::onSearchResult(const QString searchPattern,
       bool merged=false;
       for (auto locB = locA+1; locB != foundLocations.end(); ++locB) {
         if (Equal(*locA, *locB)){
-          locB->mergeWith(*locA);
+          (*locB)->mergeWith(**locA);
           merged=true;
           break;
         }
@@ -148,10 +146,10 @@ void LocationListModel::onSearchResult(const QString searchPattern,
     for (auto newLocation = foundLocations.begin(); newLocation != foundLocations.end(); ) {
       int index=0;
       bool merged=false;
-      for (LocationEntry *modelLocation:locations) {
-        if (Equal(*modelLocation, *newLocation)){
+      for (LocationEntryRef& modelLocation:locations) {
+        if (Equal(modelLocation, *newLocation)){
           // qDebug() << "Merge " << b.getLabel() << " to location " << a.getLabel();
-          modelLocation->mergeWith(*newLocation);
+          modelLocation->mergeWith(**newLocation);
           merged=true;
 
           // emit data change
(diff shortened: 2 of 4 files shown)


Karry:
Hi. Are you able to reproduce this crash? Can try it will full debug symbols, to see source line numbers? I don't see problem in current code. Element in the list is always erased and iterator to is overwritten... `iterator = list.erase(iterator);`

janbar:
Hi, yes I able to reproduce. The crash occurs on arm64 (Android). My app running on my pc x86_64 doesn't crash for the same search.

Karry:
As the crash is happening in `QObject::parent()` method called from `LocationEntry` copy constructor, I am worry that the issue may be somewhere else than erasing list elements... 

When you are able to reproduce it, your change fixed the issue, right? Getting full stracktrace may be tricky, "println debugging" may be faster :-) 

I am using library on x86_64 (Qt 5.12.8) and armv7 (Qt 5.6.3) without any sign of trouble, even with address sanitizer...

janbar:
Yes that's right, my change fixed the issue. I have tried to investigate on the copy constructor but nothing found. I deduced the copy was made in an container operation and a double free occurred somewhere.

Framstag:
@Karry @janbar Merge or not merge?

Karry:
I would like to know exact reason of the crash, because I don't see issue in current code. I am worry that this patch is just workaround, not a solution. I hope that I will be able to build and test code on aarch64 device soon, or @janbar will be able to dig deeper into the issue.

janbar:
@karry, seems it is not fully fixed, @Framstag succeeded to crash the search even with this patch. I retry to build the android apk to be sure this patch was applied in my branch.

janbar:
So no fixed yet. The crash occurs when searching for place "Bla". Also it crashes also with arch armv7. I have to enable the debug with symbols to understand what happens.

janbar:
After putting a lot of qWarning(__FUNCTION__  + __LINE), I saw the crash occurs here:
```
std::sort(foundLocations.begin(), foundLocations.end(), Compare);
```

janbar:
Inside the lambda "compare", calling `qWarning("%s: %d b = %s", __FUNCTION__, __LINE__, b.getLabel().toUtf8().constData());` , it crashes with SIGSEGV, SIGV_MAPERR on b.getLabel() + QString(QString)
:
```
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 185 comparaisonCall = 569
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 187 compareFN return OK
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 179 a = Gleisdreieck e.V.
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 180 b = Gleisdreieck e.V.
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 185 comparaisonCall = 570
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 187 compareFN return OK
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 179 a = Gleisdreieck e.V.
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 180 b = Gleisdreieck e.V.
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 185 comparaisonCall = 571
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 187 compareFN return OK
03-25 12:46:07.489  9113  9138 W osmin   : operator(): 179 a = Gleisdreieck e.V.
03-25 12:46:07.494  9113  9138 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x18 in tid 9138 (qtMainLoopThrea), pid 9113 (ub.janbar.osmin)
03-25 12:46:07.657  9188  9188 F DEBUG   : pid: 9113, tid: 9138, name: qtMainLoopThrea  >>> io.github.janbar.osmin <<<
03-25 12:46:07.671  9188  9188 F DEBUG   :     #00 pc 00000000003e42fc  /data/app/io.github.janbar.osmin-Q81UK4nYiDHvh_6hCYoEvQ==/lib/arm64/libosmin.so (QString::QString(QString const&)+48)
03-25 12:46:07.671  9188  9188 F DEBUG   :     #01 pc 00000000004f681c  /data/app/io.github.janbar.osmin-Q81UK4nYiDHvh_6hCYoEvQ==/lib/arm64/libosmin.so (osmscout::LocationEntry::getLabel() const+32)
03-25 12:46:07.671  9188  9188 F DEBUG   :     #02 pc 0000000000530004  /data/app/io.github.janbar.osmin-Q81UK4nYiDHvh_6hCYoEvQ==/lib/arm64/libosmin.so (_ZZN8osmscout17LocationListModel14onSearchResultE7QString5QListINS_13LocationEntryEEENK3$_1clERS3_RKS3_+108)
```

janbar:
I guess the issue is in the QML engine itself. I don't know how are handle compareFn and equalsFn , who is set these pointer and when ... ???

janbar:
@karry, unassigning compare() and equals() from the LocationListModel in QML code then the issue doesn't occurs anymore. Those JS function are inspired from your original code in OSMScout. I have to investigate what happens here with Qt >= 5.12 ...

janbar:
@Karry, I found !!!

janbar:
The root cause seems to be a mix with iterator and iterator of pointer. A sample as follow:
https://github.com/Framstag/libosmscout/blob/fcc882e32347c7c77e07ba30080a060a984eae80/libosmscout-client-qt/src/osmscout/SearchLocationModel.cpp#L197
But fixing this one is not enought. Finally I changed the container foundLocations to store new allocated pointer instead copying entry from the const list. Then I free (delete) merged entries and all they aren't inserted into the list of the model.

Karry:
You are right. Problem is in iterator to pointer! `LocationEntry` constructors (with `QObject* parent = nullptr`) should be explicit, compiler then shows where the problem is. Current code works, because pointer to location entry is used as a parent (implicit constructor). 

And problem starts. When entry copy is tranferred to the QML, its ownership is setup (QML don't takes ownership in that case), when parent is deleted from c++ code, Qt is deleting even children that may be referenced from QML that time... and it crash!

To correct it
 - problematic `LocationEntry` constructor have to be explicit
 - iterator have to be de-referenced twice 
 - we should add `assert(a.parent() == nullptr && b.parent() == nullptr);` to `Compare` function

Thank you @janbar !

Framstag:
@Karry Waiting for your approval
@janbar Please offer I new osmin download afterwards, so I can test https://github.com/janbar/osmin/issues/1 and routing, too :-)

janbar:
The current commit works perfectly. I tested many search and the crash no longer occurs.

Now I try to make only the change you mentioned ... and it crash again ;-).

Rolbacking to this commit and no longer crash.

janbar:
@framstag, sure , a couple of minutes ...

Framstag:
> @Framstag, sure , a couple of minutes ...

I already got a release message, however there are only source assets available? Should I wait?

janbar:
@Framstag , yes you should, the apk are uploading ....

janbar:
@Framstag , the apk are available now.

janbar:
Using pointers for QObject in functor (lambda) seems to resolve the issue. But I don't know why ...

Karry:
I would like to offer more simpler fix: https://github.com/Framstag/libosmscout/pull/1024
Problem is described there. Thank you @janbar for investigation.

Karry:
I tried to reproduce this crash on my raspberry (armv7) with llvm libc++, but unsuccessfully. I would really love to know where the problem is, but I am giving up for now. You are right that QList copies share underlying data that are copied on modification, it may be tricky on arm memory model... When some simplification in code solve the issue, I am ok with it. We waste many hours on it right now, there are more interesting things to do.

I would like to ask you for last modification - please use `std::shared_ptr` on places with `LocationEntry` - even slot/signal api, raw pointers are painful for usage. Just create some new type like `using LocationEntryPtr = std::shared_ptr<LocationEntry>`...

janbar:
@Karry, right I will make those update using shared pointers.

janbar:
@karry,
Using shared pointer is not feasible because the ownership cannot be transferred when we push entry to the model list.
Even using unique pointer is feasible, but we cannot mix pointer and std::unique_ptr when calling the functor "compare". Currently the list model store pointer aka (*LocationEntry) and if you want use unique_ptr we have to create a second functor compare(std::unique_ptr& a, std::unique_ptr& b) to handle the sort on entries of the working list foundLocation.

Karry:
> Using shared pointer is not feasible because the ownership cannot be transferred when we push entry to the model list
why? :-) You may share one `shared_ptr` from multiple collections...

janbar:
> > Using shared pointer is not feasible because the ownership cannot be transferred when we push entry to the model list
> 
> why? :-) what lines do you mean? You may share one `shared_ptr` from multiple collections...

@karry, using shared pointer you cannot release the ownership when transfering the pointer into the list model:
`positionIt = locations.insert(positionIt, location)`
Using unique pointer you can:
`positionIt = locations.insert(positionIt, location.release())`
But also we need to create a specialized functor "compare" to handle the sort of the list `foundLocations<std::unique_ptr<LocationEntry> >`.

janbar:
I do it for sport ;-) ... and you will see

janbar:
Need test on arm64 and armv7 ...

janbar:
All work great on x86 and arm64. No armv7 around me but it should be same as arm64.

janbar:
I squash the commits to make only one for merging if you agree ;-)

janbar:
Please let me know if you want I change the comment in the commit to give us a better explanation of this weird issue.

janbar:
@Karry, sorry I haven't read you full message before, because I am working at the the time :-D. So understood now I can make changes everywhere is required ...

janbar:
Don't see where I can do more. We should keep pointer for QML bindings in others places.

libosmscout-client-qt/src/osmscout/NearPOIModel.cpp
```diff
@@ -182,7 +176,7 @@ void NearPOIModel::onLookupResult(int requestId, QList<LocationEntry> newLocatio
     }
 
     emit beginInsertRows(QModelIndex(), position, position);
-    locations.insert(position, new LocationEntry(location));
+    locations.insert(position, LocationEntryRef(new LocationEntry(location)));
```
Karry, line 179: Instead of allocating with `new` I would use:
```
locations.insert(position, std::make_shared<LocationEntry>(location));
```

libosmscout-client-qt/src/osmscout/SearchLocationModel.cpp
```diff
@@ -102,19 +96,23 @@ void LocationListModel::onSearchResult(const QString searchPattern,
   int equalityCall=0;
   int comparisonsCall=0;
 
-  QList<LocationEntry> foundLocations = foundLocationsConst;
+  QList<LocationEntryRef> foundLocations;
+  for (const LocationEntry& e : foundLocationsConst)
+    foundLocations.push_back(LocationEntryRef(new LocationEntry(e)));
```
Karry, line 101: Instead of allocating with `new` I would use:
```
foundLocations.push_back(std::make_shared<LocationEntry>(e));
```

libosmscout-client-qt/src/osmscout/SearchLocationModel.cpp
```diff
@@ -293,7 +286,7 @@ void LocationListModel::setPattern(const QString& pattern)
       QString name=QString::fromLocal8Bit(coord.GetDisplayText().c_str());
       QString label=name;
 
-      LocationEntry *location=new LocationEntry(label, coord);
+      LocationEntryRef location=LocationEntryRef(new LocationEntry(label, coord));
```
Karry, line 289: `std::make_shared<LocationEntry>(label, coord)` please :-)

libosmscout-client-qt/src/osmscout/NearPOIModel.cpp
```diff
@@ -169,7 +163,7 @@ void NearPOIModel::onLookupResult(int requestId, QList<LocationEntry> newLocatio
     if (distance > maxDistance){
       continue;
     }
-    for (LocationEntry* secondLocation:locations) {
+    for (LocationEntryRef& secondLocation:locations) {
```
Karry, line 166: `secondLocation` may be const here, right?

Karry (changes requested):
LGTM otherwise, thank you!

Karry (approved):

janbar:
@karry, bad news it crashes again using shared pointer.  I investigate.....

janbar:
@Karry, bad news it crashes again with shared pointers. II iinvestigate and retry with pointers...

janbar:
@Karry , bad news it crashes again with shared pointers. I investigate and retry with pointers ...

janbar:
Once at home I retry passing primitve pointers and let you know

PR #1015, opened 2021-03-21
Signal the route request should be broken on no route

janbar:
Hi, first many thanks for this project. Yesterday I consumed the day to rebase my changes with the master and I decide to propose some changes to be definitively in the master branch. You can refuse it, then I will continue to rebase. I explained the purpose of each commit, and there are really small and non intrusive.

### DIFF
--- a/libosmscout/include/osmscout/navigation/DataAgent.h
+++ b/libosmscout/include/osmscout/navigation/DataAgent.h
@@ -67,6 +67,16 @@ namespace osmscout {
 
   using RoutableObjectsMessageRef=std::shared_ptr<RoutableObjectsMessage>;
 
+  /**
+   * Message created when we should break.
+   */
+  struct OSMSCOUT_API NoRoutableObjectsMessage CLASS_FINAL : public NavigationMessage
+  {
+    NoRoutableObjectsMessage(const Timestamp& timestamp);
+  };
+
+  using NoRoutableObjectsMessageRef=std::shared_ptr<NoRoutableObjectsMessage>;
+
   template <typename DataLoader>
   class DataAgent CLASS_FINAL : public NavigationAgent
   {
@@ -105,6 +115,8 @@ namespace osmscout {
         if (GetSphericalDistance(requestMessage->bbox.GetMinCoord(),
                                  requestMessage->bbox.GetMaxCoord()) > Kilometers(2)){
           log.Warn() << "Requested routable data from huge region: " << requestMessage->bbox.GetDisplayText();
+          result.push_back(std::make_shared<NoRoutableObjectsMessage>(requestMessage->timestamp));
+          return result;
         }
 
         auto msg=std::make_shared<RoutableObjectsMessage>(requestMessage->timestamp, std::make_shared<RoutableObjects>());
--- a/libosmscout-client-qt/include/osmscout/NavigationModel.h
+++ b/libosmscout-client-qt/include/osmscout/NavigationModel.h
@@ -90,6 +90,8 @@ signals:
 
   void laneUpdate();
 
+  void breakRequest();
+
 public slots:
   void locationChanged(bool locationValid,
                        double lat, double lon,
@@ -116,6 +118,8 @@ public slots:
   void onMaxAllowedSpeed(double maxAllowedSpeed);
 
   void onLaneUpdate(osmscout::LaneAgent::Lane lane);
+
+  void onBreakRequest();
 
 public:
   using Roles = RouteStep::Roles;
--- a/libosmscout-client-qt/include/osmscout/NavigationModule.h
+++ b/libosmscout-client-qt/include/osmscout/NavigationModule.h
@@ -77,6 +77,8 @@ signals:
 
   void laneUpdate(osmscout::LaneAgent::Lane lane);
 
+  void breakRequest();
+
 public slots:
   void setupRoute(QtRouteData route,
                   osmscout::Vehicle vehicle);
--- a/libosmscout-client-qt/src/osmscout/NavigationModel.cpp
+++ b/libosmscout-client-qt/src/osmscout/NavigationModel.cpp
@@ -69,6 +69,10 @@ NavigationModel::NavigationModel():
           this, &NavigationModel::routeAheadChanged);
   connect(this, &NavigationModel::arrivalUpdate,
           this, &NavigationModel::routeAheadChanged);
+
+  connect(navigationModule, &NavigationModule::breakRequest,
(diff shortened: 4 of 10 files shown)


libosmscout/include/osmscout/navigation/DataAgent.h
```diff
@@ -105,6 +115,8 @@ namespace osmscout {
         if (GetSphericalDistance(requestMessage->bbox.GetMinCoord(),
                                  requestMessage->bbox.GetMaxCoord()) > Kilometers(2)){
```
Framstag, line 116: I do not like a hardcoded value like this in generic code. The value should be configurable (with a sensible default).

Framstag (changes requested):

Framstag:
@janbar Thanks for your path. Most of it looks OK, but since most changes are in qt client library it is up to @Karry :-)

Please try to offer smaller patches in future. This allows use to let all patches pass quickly that are OK, without waiting for corrections for the complete patch while only a minor fix for a small part is required to change.

janbar:
I can split the PR for each commit If you prefer.

Framstag:
If it is OK for @Karry it is OK for me with this one the way it is. This was just meant as a hint for future patches (which I'll enjoy to see of course :-)).

Karry:
Hi @janbar. Thank you for contribution! I just start reviewing changes commit by commit. Some of them is absolutely fine and we may merge them as it. But I would like to discuss other commits. For example HiDPI change in MapPainter...

I would really love to have these changes as separate merge requests.

janbar:
@Karry , Please let me know which commits must be moved to new PRs.

janbar:
Voila I splitted the initial PR.

Framstag:
We now moved much of the changes to individual patches. I assume that I reject this patch, but is there anything left not merged via other patches?

janbar:
@Framstag , still here the important commits:

- "introduce a signal to break request on no route" 
This one send a new message to the navigation module frontend : "BreakRequest" to inform it there is no route available around the limited area. If the navigation module handles this new message, it no longer try to reroute automatically and it should ask to the user to reroute on demand i.e by pressing a button.
Sending the message we don't try to calculate a new route because in many case I tested on real courses, it will crash the lib after stucking the cpu ... to find a way the browsed area is huge.

- "fix uninitialized position"
 This one fix an issue during navigation when the current position is unavailable for an instant. In this case the map move to the unintialized position (lat 0, lon 0) in the middle of atlantic ocean, then on next change of position the map move again to the valid position. So here I initialize the new position the the last valid position.

Framstag:
I suggest to make individual pacthes for these, too. Simply because I have the impression that this work good for all participians. 

I encurrange you also to use the GitHub mechanismen for marking reviewer, acking reviews for this. I'm currently the only one to merge, but do not have the time to do in deep reviews myself. Having clear status, allows to you "blindly" merge (fast).

OK for you?

janbar:
Ok. I move the commit "fix uninitialized position" in a new PR.

libosmscout/include/osmscout/navigation/DataAgent.h
```diff
@@ -105,6 +115,8 @@ namespace osmscout {
...
           log.Warn() << "Requested routable data from huge region: " << requestMessage->bbox.GetDisplayText();
+          result.push_back(std::make_shared<NoRoutableObjectsMessage>(requestMessage->timestamp));
+          return result;
```
Karry, line 119: When requested area is huge, it doesn't mean that there are no routable objects, right? If I understand problem correctly, main issue is in `PositionAgent` that is requesting data again and again and cause infinite loop that leads to stack overflow, right? Would not be better to fix issue in `PositionAgent` itself instead of moving responsibility to UI code or worse - user, to re-try whole routing?

Karry:
Btw @janbar, do you have some recorded gpx track where this problem happened? It should be possible to reproduce it in `NavigationSimulation` demo...

Karry:
> Ok. I move the commit "fix uninitialized position" in a new PR.

Thank you. This commit is simple and straightforward.

janbar:
> Btw @janbar, do you have some recorded gpx track where this problem happened? It should be possible to reproduce it in `NavigationSimulation` demo...

I haven't track record to check it. But it is simple to reproduce:
- Make a route from your position to a place
- Launch the routing
- Start to follow the route for i.e 1 km
- Finally take an other way in the land (a non-drivable path): after few meters (100 or more) the router recalculate a new route ... but in this case it cannot find it ... I let you see what happens ...

Karry:
Hi, I tried simulation with "car" vehicle and use record from [our hike](https://osmscout.karry.cz/tests/Lecice.gpx) to simulate gpx. When position leave planned route, position state switched to OffRoute state, route was re-planned... When position goes to the track (not routable for a car), re-routing fails, route state switched to Unitialized and nothing special happened then. So I am not able to reproduce your issue. At least with NavigationSimulation demo :-/

janbar:
@Karry, the next week I will try the app without this patch and with the latest libosmscout. May be some changes since last months have fixed this issue.

janbar:
@Karry, running a simulation with the app and latest libosmscout the issue does not occur on my desktop. No high cpu usage like before when the position is in the land. So like you. I will confirm during the week by a real test on road / off road.
Great news if it has been resolved !

Karry:
It sounds promising! Btw, I may recommend you to start tracker when you are using navigation. It is not fun to do debugging while you are driving :-D I found this simulator very useful...

janbar:
I will give a response this week-end if the issue has been resolved. Until now I have no longer the issue with the latest master.

libosmscout/include/osmscout/navigation/DataAgent.h
janbar: Yes you are right. Please see my next comments.

janbar:
The initial issue no longer exits when you move in an area without possible route. Still one case where a similar issue come back (high cpu usage by the navigation module). It is tricky but you can reproduce easilly as follows:

- disconnect the gps sensor: so the position is invalid (i.e lat=0 , lon=0, posValid=false)
- make a route from a location you choose on the map, to an other location as destination.
- start navigation by setup navigation module with the calculated route, and the destination.

A similar issue occurs.

I am able to reproduce this case on a PC without positionning module. Or setuping the navigation module when sensor become disconnected.

Framstag:
Sounds like you should make an issue with a simple reproducer for the issue you mentioned. If you do not need this pull request anymore, please delete it.

Many thanks for the various improvements that were initially part of this request!

janbar:
I will create an issue about the last test case

PR #1016, opened 2021-03-21
fix size of painted items using HiDPI device

janbar:
Using device with HiDPI, painted items are not positioned correctly.
Only the client app knows the ratio to compute the right position. The pixel ratio can be
configured with "setPixelRatio()". Obviously the default is 1.0.

### DIFF
--- a/libosmscout-client-qt/src/osmscout/MapWidget.cpp
+++ b/libosmscout-client-qt/src/osmscout/MapWidget.cpp
@@ -277,9 +277,8 @@ void MapWidget::paint(QPainter *painter)
       }
 
       painter->save();
-      QTransform t=QTransform::fromTranslate(x, y); // move to rotation center
-      t.rotateRadians(iconAngle.AsRadians());
-      painter->setTransform(t);
+      painter->translate(x,y);
+      painter->rotate(iconAngle.AsDegrees());
       // draw vehicleIcon center on coordinate 0x0
       painter->drawImage(QPointF(vehicleIcon.width()/-2, vehicleIcon.height()/-2), vehicleIcon);
       painter->restore();
@@ -782,7 +781,7 @@ QImage MapWidget::loadSVGIcon(const QString &directory, const QString fileName, 
 
 void MapWidget::loadVehicleIcons()
 {
-  double iconPixelSize=getProjection().ConvertWidthToPixel(vehicle.iconSize);
+  double iconPixelSize=getProjection().ConvertWidthToPixel(vehicle.iconSize * vehicleScaleFactor);
   QString iconDirectory=OSMScoutQt::GetInstance().GetIconDirectory();
 
   vehicle.standardIcon=loadSVGIcon(iconDirectory, vehicle.standardIconFile, iconPixelSize);
@@ -850,6 +849,13 @@ bool MapWidget::toggleInfo()
     osmscout::log.Info(!osmscout::log.IsInfo());
 
     return osmscout::log.IsInfo();
+}
+
+void MapWidget::setVehicleScaleFactor(float factor)
+{
+  vehicleScaleFactor = factor;
+  loadVehicleIcons();
+  redraw();
 }
 
 QString MapWidget::GetRenderingType() const
--- a/libosmscout-client-qt/include/osmscout/MapWidget.h
+++ b/libosmscout-client-qt/include/osmscout/MapWidget.h
@@ -140,6 +140,8 @@ private:
   };
   Vehicle vehicle;
 
+  float vehicleScaleFactor{1.0};
+
 signals:
   void viewChanged();
   void lockToPossitionChanged();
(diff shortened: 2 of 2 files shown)


janbar:
This one try to fix an issue on device HiDPI.
Sailfish OS is not, so you cannot test on it. Android is, and any desktop too.
The issue is the vehicule icon and all others painted items over the map haven't the expected size on device with screen HiDPI. My fix allows to set the ratio when initializing the map engine. Probably I used the wrong way...

janbar:
The pixel ratio to apply is retrieved as follow:

https://github.com/janbar/osmin/blob/3a3b9c05944c48a749af91eb927f75113bcad105/src/osmin.cpp#L330

janbar:
I will add screenshots to show you what happens with HiDPI screen...

janbar:
**Without the patch**, the vehicule icon is positionned at  hard (x,y) in the screen, and its size is the hard (h,w): (the blue arraow in the left/top corner):
Screenshot_20210322-192125

janbar:
With the HiDPI patch applied, all is right now:
Screenshot_20210322-191747

janbar:
Now you know where I live ;-).

libosmscout-client-qt/src/osmscout/MapWidget.cpp
```diff
@@ -277,7 +277,7 @@ void MapWidget::paint(QPainter *painter)
       }
 
       painter->save();
-      QTransform t=QTransform::fromTranslate(x, y); // move to rotation center
+      QTransform t=QTransform::fromTranslate(x * pixelRatio, y * pixelRatio); // move to rotation center
```
Karry, line 280: I don't understand where is the problem here (always have problem with these transformation matrices :-) ). Why map icons, that are SVG images too, are rendered on correct place but vehicle icon is on the wrong place? Or map tiles during tiled rendering where Transformation is used too?

May it be caused by the fact that icon and painter have different device pixel ratio? Can you try to add these debug lines please?
```
qDebug() << "icon pixel ratio: " << vehicleIcon.devicePixelRatio();
qDebug() << "painter pixel ratio: " << painter->device()->devicePixelRatioF();
```

Karry (commented):
I would like to have sensible default value. It can be evaluated in `DBThread` singleton for example, there is obtained physical dpi already... 
```
physicalDpi = (double)srn->physicalDotsPerInch();
```

But first, I would like to understand why Transformation is wrong, it doesn't make sense to me now.

libosmscout-client-qt/src/osmscout/MapWidget.cpp
Karry: both are 1.0 on my desktop...

janbar: Not the correct place and not the correct size. In my sample the screen pixel ratio is 3.

janbar: I try the debug output on macos with retina screen which has a HiDPI ratio of 2...

janbar: Debug output shows:
```
icon pixel ratio: 1
painter pixel ratio: 1
```

Karry: But other usages of `projection.GeoToPixel` have to be wrong too, or not? Where is the translation between map painter coordinates and upscaled screen coordinates? How it is possible that tile rendering on rotated map is correct?! Did you understand my concerns?

janbar: I try to understand ;).
I your desktop is Linux/Gnome i.e ubuntu, try to upscale the screen with the factor 200% and you will see what happens... the HiDPI will be 2.0

janbar: Also the tile rendering is correct

janbar: How do you make your test ? Have you an app for desktop I can test here with the HiDPI ?

janbar: If your desktop is Linux, you could try with my app. I guess you have Qt headers installed. To build it, clone the branch dev/20210323:
```
git clone -b dev/20210323  https://github.com/janbar/osmin.git
cd osmin
git submodule init
git submodule update
mkdir build
cd build
cmake -B build .
cmake --build build/ -j8
```
finally launch the app: ./build/osmin
Download a map or copy an existing map into the folder $HOME/Maps/
In the file `src/osmin.cpp` line 330 you can set a fixed value for ScreenScaleFactor (i.e QVariant(1.0))
Then scale your screen in the Display/Parameters to 200%, restart the app ...

Karry: I was able to reproduce it with your procedure and NavigationSimulation demo. It is really strange. When I replaced usage of QTransform with these two lines:
```
      painter->translate(x,y);
      painter->rotate(iconAngle.AsDegrees());
```
it is working perfectly... I would guess that there is some bug in Qt or I just don't understand QTransform :-)

So, I would advice you to replace transform in your branch and rename `pixelRatio` to "vehicle icon scale" or something like this :-)

janbar:
@Karry , OK I make those changes

janbar:
@karry, that works great now.

Karry (approved):

Karry:
Thank you for the changes and your patience with my comments and questions :-)

PR #1018, opened 2021-03-21
allow to change map style from qml frontend

janbar:
Introduce GetStyleSheetFile() and SetStyleSheetFile() to get/set the map stylesheet
from QML frontend. Setting style will change the map style online.


Karry (approved):
LGTM. By the way, there is `MapStyleModel` that provides list of `*.oss` files from style sheet directory to UI and allows to change the style. But it is fine to have this possibility in settings directly.

I am just curious, what UI components you are using on Android?

janbar:
I use QtQuickControl2 for the android app. The source is https://github.com/janbar/osmin/tree/master/gui/controls2.

janbar:
@Karry , I checked why I didn't use the functions get/setStyle from MapStyle. Those functions are not INVOKABLE, and so we can'ot use them from QML.

Karry:
These methods don't have to be invokable, it is exposed as a property:
```
Q_PROPERTY(QString  style  READ getStyle  WRITE setStyle  NOTIFY styleChanged)
```

So, in QML you may just write to the property...
```
MapStyleModel {
  id: styleModel
  Component.onCompleted: {
    styleModel.style = "standard.oss"
  }
}
```

janbar:
Reviewing my code (one year passed), I used the property style from MapStyle to set the stylesheet. As you understood this change was made to keep the change in settings for next restart.

Karry:
It should work from MapStyle too :-)
```
void MapStyleModel::setStyle(const QString &styleFile)
{
  SettingsRef settings=OSMScoutQt::GetInstance().GetSettings();

  settings->SetStyleSheetFile(styleFile);

  emit loadStyleRequested(settings->GetStyleSheetAbsoluteFile(),
                          settings->GetStyleSheetFlags());
}
```

PR #1019, opened 2021-03-21
enable result limit for NearPOIModel

janbar:
Break lookup result after reached limit

### DIFF
--- a/libosmscout-client-qt/include/osmscout/NearPOIModel.h
+++ b/libosmscout-client-qt/include/osmscout/NearPOIModel.h
@@ -67,7 +67,7 @@ Q_OBJECT
   Q_PROPERTY(double   maxDistance READ GetMaxDistance WRITE SetMaxDistance)
 
   /**
-   * Limit of results for each database.
+   * Limit of lookup results.
    */
   Q_PROPERTY(int      resultLimit READ GetResultLimit WRITE SetResultLimit)
 
--- a/libosmscout-client-qt/src/osmscout/NearPOIModel.cpp
+++ b/libosmscout-client-qt/src/osmscout/NearPOIModel.cpp
@@ -176,6 +176,11 @@ void NearPOIModel::onLookupResult(int requestId, QList<LocationEntry> newLocatio
       position++;
     }
 
+    // do not care of result outside the limit
+    if (position > resultLimit) {
+      continue;
+    }
+
     emit beginInsertRows(QModelIndex(), position, position);
     locations.insert(position, new LocationEntry(location));
     qDebug() << "Put " << location.getObjectType() << location.getLabel() << " to position: " << position << "(distance" << distance.AsMeter() << " m)";


Karry (changes requested):
It make sense. Kind of. Just in header is comment that property `Limit of results for each database.`. It was not true, because this limit was never implemented, but now it is true either, because it is limit for whole model. Can you update documentation in header file please?

Moreover, entries from `POILookupModule` are not sorted (by distance from search center), so it may happen that this limit will remove POIs that are closest. Would not be better to do NOT add entries which target position is greater than limit?
```
if (position > resultLimit) {
  continue;
}
```
That way, model size may exceed the limit, but it will be guaranteed that first X nearest entries will be displayed...

janbar:
Ok I will change this evening asap

janbar:
I anderstand what you means. The goal of this commit was to stop very long queries returning a lot of results and blocking the UI thread by inserts in the model. The human user doesn't care to see thousand POI. I think he want a quick response with readable result.
Your suggestion make sense, discarding update in the model if position is over the limit. I will try on the mobil device a query with this code. On a desktop with a powered cpu we have always an acceptable response time ...

janbar:
Bingo, you are right.
My commit didn't reduce drastically the response time for an huge query. With your recomandations the gain is better:
Initial response time: 20 sec, and discarding result the response time is 16 sec.

janbar:
I updated the documentation in the header.

Karry (approved):

Karry:
Great. But still, 16 seconds for the poi result? It is long. It is instant on Jolla phone (single poi type, up to distance 10km).

janbar:
@karry, my test was made for searching an address "avenue robert". Looking for a particular POI type (gas station or medical center etc near the current position) is instant.
The same type of search on Jolla was very long, near two/four times longer. I also gave up using Jolla because the performance of the map renderer was bad compared to android due to missing of decent implementation of openmp. The map rendering on android (I use LineageOS and Android 8 to 10) is fore times faster than Jolla on the same devices (Sony Xperia XA2 and 10).

Framstag:
A performance degration form desktop to mobbile is to be expexted, but from the few videos Lukas posted over the years, performance seemed to be OK.

The overall answer of libosmscout in this case of course is not "do'nt do it", but customize your stylsheet to reduce the number of rendered objects per zoom level (in average) :-)

Not that adress search and POI search may use different indexes, dependning on the APIs you use. It would be interesting to get more details on this.

Btw., since I have a Pixel 3, is there an easy way to test on an actual device :-)?

janbar:
@Framstag , which version of Android for Pixel 3 ? and what is the arch (arm64 or armv7) ?

I will post in the release page of the app an APK for your device to test it ....

janbar:
@Framstag , you can download the APK for your pixel 3 from the release page of the app :

https://github.com/janbar/osmin/releases/tag/1.3.0-alpha

Get the file **osmin-1.3.0-arm64-nougat.apk** which is for your device (Android 9 arm64). I let you discover the app but it is simple. On first startup you will need to go to the "map download page" (same as osmscout), then choose the map for your location, sweep the screen to the top to see download progress ...

Framstag:
@janbar Installation and app starting works. I personally find map rendeirng on the Pixel 3 definitely fast :-)
 
 However I canot check further because you only offer germany, which is a rather big map :-) Is it possibly to add nordrhein-westfalen as download (https://download.geofabrik.de/europe/germany.html)? That map should be much smaller and is where I live. This would allow me to further test the application and give feedback.
 
 Btw., it is only the third time I actually saw "my" software on a mobile device, so please continue (first time my own experiments years ago under android, second time a Jolla phone (good job, @Karry!) of a friend and now my own mobile!)! Thanks!

Framstag:
I'll merge now :-)

janbar:
@Framstag , you can "build" the map for the area you want, then upload it manually in your mobile in the folder:

[Internal_Storage] /Android/data/io.github.janbar.osmin/files/Maps/

Or if you have an additional storage card (SD card) the MAPs are stored in :

[SD card] /Android/data/io.github.janbar.osmin/files/Maps/

Framstag:
I'll try. However since USB is not working anymore, this might get a little bit tricky...

Is the map data source yours or the (raspberry pi) server @Karry offers?

janbar:
@Framstag , Yes it is. The same data source (see map-provider file in resource)

PR #1023, opened 2021-03-25
fix uninitialized position

janbar:
In some case position could be uninitialized. So here I initialize the new position the the last valid position.

### DIFF
--- a/libosmscout/src/osmscout/navigation/PositionAgent.cpp
+++ b/libosmscout/src/osmscout/navigation/PositionAgent.cpp
@@ -118,6 +118,7 @@ namespace osmscout {
       double nearest=std::numeric_limits<double>::max();
 
       PositionAgent::Position position;
+      position.coord=coord;
 
       for (auto &e:routableObjects->dbMap){
         DatabaseId dbId=e.first;


janbar:
@Karry, I moved the fix relates "fix uninitialized position" here.

libosmscout/src/osmscout/navigation/PositionAgent.cpp
```diff
@@ -118,6 +118,7 @@ namespace osmscout {
       double nearest=std::numeric_limits<double>::max();
 
       PositionAgent::Position position;
+      position.coord.Set(coord.GetLat(), coord.GetLon());
```
Framstag, line 121: Isn't this just `position.coord=coord`? A direct assign of coord object is safer, because it avoid lat/lon getting passed inverted.

Possibly PositionAgent should get an explicit constructor to enforce passing a valid coord?

janbar: You are right, the operator=() is preferable

Karry (approved):
LGTM. 

It is possible to handle `Uninitialised` car position in UI to avoid jumps to "zero island" (0,0 coord), but this solution is more error prone. Thank you!

PR #1025, opened 2021-03-26
Keep sub directory structure in OSMScout framework headers

vyskocil:
Keep the subdirectory structure in libosmscout  headers when building Apple framework.
This is the first step to fix osmscout frameworks compilation, the second step would be to fix osmscout-map and osmscout-map-iosx because there is still a issue to merge the osmscout top dir accross several frameworks. 
Actually I use the following commands to copy manually osmscout-map and osmscout-map-iosx headers to osmscout framework : 

 cp libosmscout-map/OSMScoutMap.framework/Headers/*.h libosmscout/OSMScout.framework/Headers
 cp libosmscout-map-iosx/OSMScoutMapIOSX.framework/Headers/*.h libosmscout/OSMScout.framework/Headers 
 cp libosmscout-map/include/osmscout/MapFeatures.h  libosmscout/OSMScout.framework/Headers

### DIFF
--- a/libosmscout-map-iosx/include/osmscout/MapPainterIOS.h
+++ b/libosmscout-map-iosx/include/osmscout/MapPainterIOS.h
@@ -25,6 +25,7 @@
 
 #include <string>
 #include <unordered_map>
+#include <utility>
 
 #if TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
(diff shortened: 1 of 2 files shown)


Framstag:
Hello @vyskocil, nice to see you again :-) What is the cref_ptr template class about?

vyskocil:
Hi @Framstag ! Yes I'm back in the business :-) Seeing all the great work done on the Android side motivated me to address some of the issues on the Apple side !
The cref_ptr is a small template for doing as Mozilla says : "A CoreFoundation smart pointer" (https://hg.mozilla.org/mozilla-central/file/tip/dom/media/platforms/apple/AppleUtils.h), it deals with managing pointers to CoreFoundation objects in Objective C++ (C++ and ObjectiveC bridged) to avoid memory leaks.
I also found it here : https://github.com/kbinani/cfref_ptr
I used this years ago in osmscout-map-iOSX in a separate private include file but it has to be copied along so I decided to include it in the MapPainterIOS.h header file.
I looked how to split the the MapPainterIOS.h in private/public files but it is beyond my C++ knowledge with the templates,...
but the MapPainterIOS.h is not really a public header, it is only used by osmscout-map not by the final library user.

Framstag:
Yes, while I'm currently making some issues for the Android app, I'm overall very pleased with the current version of the Android app. Things are snappy and fast in general with already good quality (@Karry, @janbar Excellent work!). It looks like libosmscout in fact allows people to concentrate on visualization and interaction with the user with libsosmcout giving a good base framework for the original "nav" stuff.

As such I would be happy to be have a good support and app for iOS, too :-)

Since Qt should support iOS, too, I see the option to join forces. The Android and Sailfish apps already share a huge amount of medium layer code via the libosmscout-client-qt library. But of course I take a native iOS, too :-)

Thanks for the explanation of the new header. I'll merge now. The broken builds seems to be unrelated to your changes.

vyskocil:
We were using Qt a lot at work for Windows/MacOS multi-platform developments 3-4 years ago but not any more. Since then I also switched to mobile development team and now I'm using Objective-C/Swift only on iOS, but 3-4 years ago I followed the Qt development for mobiles, at this time it started to be usable on Android but iOS was still very experimental, now I hope it would as advanced as for Android.
I think OSMScout should at first provide a good native iOS/OSX libraries because I don't see a big momentum for Qt in iOS world, it seems as it never really take off... but I'm still interested in it and will have a look on using qt bindings on iOS and MacOS and provide some help on this.
I also think it's time to provide a complete Swift bindings and I started thinking about it.

Issue #1026, opened 2021-03-27
Compilation issue for NavigationSimulation on MacOS


vyskocil:
CMake did detect marisa lib : 

-- core library:                    ON
--  - marisa support:               TRUE

but when the project is compiled on MacOS are error occur when compiling NavigationSimulation : 
```
Scanning dependencies of target NavigationSimulation
[ 60%] Building CXX object Demos/CMakeFiles/NavigationSimulation.dir/src/NavigationSimulation.cpp.o
[ 60%] Building CXX object Demos/CMakeFiles/NavigationSimulation.dir/src/PositionSimulator.cpp.o
[ 60%] Building CXX object Demos/CMakeFiles/NavigationSimulation.dir/__/OSMScout2/src/Theme.cpp.o
[ 61%] Building CXX object Demos/CMakeFiles/NavigationSimulation.dir/src/QtDemoApp.cpp.o
In file included from /Users/vyskocil/Dev/libosmscout-vyskocil/Demos/src/QtDemoApp.cpp:22:
In file included from /Users/vyskocil/Dev/libosmscout-vyskocil/libosmscout-client-qt/include/osmscout/OSMScoutQt.h:30:
In file included from /Users/vyskocil/Dev/libosmscout-vyskocil/libosmscout-client-qt/include/osmscout/SearchModule.h:29:
/Users/vyskocil/Dev/libosmscout-vyskocil/libosmscout/include/osmscout/TextSearchIndex.h:29:10: fatal error: 'marisa.h' file not found
#include <marisa.h>
         ^~~~~~~~~~
1 error generated.
make[2]: *** [Demos/CMakeFiles/NavigationSimulation.dir/src/QtDemoApp.cpp.o] Error 1
make[1]: *** [Demos/CMakeFiles/NavigationSimulation.dir/all] Error 2
make: *** [all] Error 2
```

vyskocil:
Marisa lib is installed in /usr/local, the includes could be found in /usr/local/include : 
```
% ls -al /usr/local/include/
total 408
drwxr-xr-x  10 root      wheel     320 Nov  4 21:14 .
drwxr-xr-x   8 root      wheel     256 Oct  7 15:51 ..
drwxr-xr-x  10 root      wheel     320 Oct  4 19:09 fuse
-rw-r--r--   1 root      wheel     246 Oct  4 19:09 fuse.h
drwxr-xr-x   3 vyskocil  staff      96 Oct 17  2014 google
[...]

PR #1027, opened 2021-03-27
expose marisa include dir as public from libosmscout-client-qt

Karry:
Hi @vyskocil can you check if this change fixes your issue https://github.com/Framstag/libosmscout/issues/1026 please?


vyskocil:
hi @Karry ! I checked this and It fix the issue, thank you !

PR #1028, opened 2021-03-27
GenMergeAreas refactoring

Framstag:
GenMergeAreas is one of the most expensive import steps for larger maps. Optimisations here should improve import time.

Goal is to refactor GenMergeAreas in a way, that we can make use of the new Thread/Work management code (likely with some necessary enhancements for thread pooling).

### DIFF
--- a/libosmscout-import/include/osmscout/import/GenMergeAreas.h
+++ b/libosmscout-import/include/osmscout/import/GenMergeAreas.h
@@ -71,29 +71,6 @@ namespace osmscout {
 
   private:
     /**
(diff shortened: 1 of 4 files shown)


Framstag:
For the curious: Initial version needed >1000 seconds for germany.osm.pbf, we are already down to <400 seconds, using most of the time all 8 CPUs for merging on my system.

* The code needs further cleanup (WorkerPool, reducing function complexity)
* There are also other currently sequential file operations that could be parallelized, further reducing running time.

PR #1042, opened 2021-04-01
do not use qAsConst introduced with qt 5.7

janbar:
As features require Qt >= 5.6, and 'qAsConst' was defined from qt 5.7, it reverts
its usage that I introduced in the commit 5569e50edd1cf59ef4f32a4ed4775ce503d51954.
So it allows to compile again the lib using qt 5.6, which still the mainline for
few OS.


janbar:
@Karry,

Here I revert the usage of qAsConst because it has been introduced in qt 5.7. I didn't care the features require qt 5.6.

Karry (approved):

PR #1043, opened 2021-04-02
Some cleanup of the render and transformation code

Framstag:


Karry (approved):
I don't test it, but looks good.

Framstag:
The main change is the assertion in GenerateParallelWay which before hat a bool return value.

In the end further changes seem to make sense. I would like to separate the caches from the methods that operate on them. All the transform methods could be functions that read write to streams/buffer/caches. That would likely simplify the code.

Recent Clion versions now show function complexity warnings. There also a number of methods in MapPainter that have such alarm. This is another place we have to clean up before things get out of control ;-)

But since I was sidetracked form hillshading I only did the obvious and quick to improve...

Framstag:
See #1044 and #1045

PR #1046, opened 2021-04-02
Hillshading infra

Framstag:
This is some code to setup some infrastructure for hillshading rendering.

It adds a command line option for hillshading to the draw demos and renderer.

It also draws a checkboard pattern of the size of the SRTM tiles (SRTM3) to get a feeling, what it is about.

Next steps:
* Do the same for contour lines
* --- different patch
* Extend MapData with SRTM data
* Add some code to get elevation lines from SRTM data

### DIFF
--- a/Demos/include/DrawMap.h
+++ b/Demos/include/DrawMap.h
@@ -30,13 +30,15 @@
 #include <iostream>
 
 struct Arguments {
-  bool help{false};
+  bool       help=false;
 #ifdef NDEBUG
-  bool debug{false};
+  bool       debug=false;
 #else
-  bool debug{true};
+  bool       debug=true;
 #endif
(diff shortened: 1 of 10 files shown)


Framstag:
You can now do:

(note the new command line options!)

```bash
Demos/DrawMapCairo --contourlines true --hillshading true ../maps/nordrhein-westfalen ../stylesheets/standard.oss 1024 800 51.565 7.455 30000 test.png && display test.png
```

and get 

Bildschirmfoto von 2021-04-02 14-58-58

Where each chess board field is an srtm tile.

I'll merge after this one. Next step would be to load actual SRTM data and start with primitive contour lines calculation.

janbar:
Awesome !

PR #1047, opened 2021-04-02
fix LocationEntry comparison

Karry:
compare function should return true just in case that first argument
is less than second. in other words, comp(a,a) have to be false

I tested it and it seems to fix crashes on Android so far... @janbar


janbar:
It WORKs for me !!!

PR #1050, opened 2021-04-03
fix crash in CoordBuffer::GenerateParallelWay

Karry:
Source way have to have at least two nodes.


Framstag:
Looks like the recent exchange of an if(...) return false to assert now bubbles up the call stack ;-)

I added another assertion in the latest pull request...

Karry:
There is always space for more assertions :-)

Problem is that during low-zoom area import rings may be reduced just two nodes (fixed by https://github.com/Framstag/libosmscout/pull/1052 ), later during rendering, when this ring/area type contains parallel line in stylesheet, it may crash. Because `GenerateParallelWay` expect at least two points...

Karry:
I added more defensive checks to renderer:
 - ways have to have at least two nodes
 - clipping rings cannot be empty

PR #1060, opened 2021-04-06
add missing roles for route step

janbar:
It add roles to bind some properties of the route step for usage with qt abstract list view. The new bound properties are:
- lat
- lon
- distance
- distanceDelta
- distanceTo
- time
- timeDelta

We need them to display extra info on a route overview.


janbar:
@Karry, some news to bind missing properties of the route step

janbar:
@Framstag , it is linked with your request of the route overview. With those info we can show all you need.

Karry (approved):
Thank you!

PR #1061, opened 2021-04-06
Zoom gesture with one finger

Karry:
Gesture is activated by finger tap followed by dragging.
similar gesture is used by some other map applications,
dragging to the top - zoom out the map, dragging to the bottom - zoom in.


Karry:
@janbar you possibly may want to test it on your device :-) Are you fine with this gesture? I am not sure if anyone was using "tap followed by press" for zoom-out, it is replaced by this new gesture. It is more common in other map applications, but still many users don't know about this trick...

PR #1062, opened 2021-04-07
fix parse error with qt moc 5.6

janbar:
This allows to compile project linked with libosmscout-gpx on SailfishOS 3.0/4.0. Seems Qt MOC 5.6 fails to parse concatenated namespace.


janbar:
@Karry, I succeeded to build again my app on Sailfish 4.0. But I had to fix little things to make MOC working with the latest master of libosmscout-gpx.

PR #1066, opened 2021-04-14
add the color feature to custom poi types

janbar:
Allows to set the preferred color for an overlay object.
Then if the type config is set accordingly as follows, the color will be applied.

Config in oss stylesheet:

[TYPE _custom] WAY { preferColorFeature: true; color: @trackColor; }


janbar:
@Karry, I proposed a change to add the color feature to the registered custom poi type.

Karry (approved):
LGTM

Karry:
I like the idea of custom waypoint and track colors, I need to extend UI in my app to allow that :-) Moreover I will need ability to export custom color to gpx...

janbar:
@Karry, cf PR #1067 as purpose for adding extensions element

PR #1067, opened 2021-04-14
add extensions to gpx file

janbar:


janbar:
I think it needs the color name i.e Red, Blue ... instead the hex string. I will check that ...

janbar:
I make some change to better support ...

janbar:
At least It works with qmapshake

janbar:
See https://docs.safe.com/fme/html/FME_Desktop_Documentation/FME_ReadersWriters/gpxx/Feature_Types.htm for better explaination of extensions.

janbar:
One thing, probably important, in the form of output. The gpxx extensions must be declared under the element "extensions" to be discarded by any app that doesn't support garmin extensions. In the original xsd, those extensions are placed in the main element. But it is preferable to stage them in "extensions" for the reason explained below.

janbar:
Due to githib bug (invalid ref) , i needed to drop the branch. The new PR is #1068

PR #1068, opened 2021-04-14
add extensions to gpx file

janbar:
The output looks like as the following.

```
 <trk>
  <name>track</name>
  <extensions>
   <gpxx:TrackExtension>
    <gpxx:DisplayColor>#8b008b</gpxx:DisplayColor>
   </gpxx:TrackExtension>
  </extensions>
  <trkseg>
...
```

### DIFF
--- a/libosmscout-gpx/CMakeLists.txt
+++ b/libosmscout-gpx/CMakeLists.txt
@@ -1,19 +1,21 @@
 set(HEADER_FILES
-	include/osmscout/gpx/GpxFile.h
-	include/osmscout/gpx/GPXImportExport.h
-	include/osmscout/gpx/Route.h
-	include/osmscout/gpx/Track.h
-	include/osmscout/gpx/Waypoint.h
-	include/osmscout/gpx/TrackPoint.h
-	include/osmscout/gpx/TrackSegment.h
-	include/osmscout/gpx/Utils.h
+    include/osmscout/gpx/GpxFile.h
+    include/osmscout/gpx/GPXImportExport.h
+    include/osmscout/gpx/Route.h
+    include/osmscout/gpx/Track.h
+    include/osmscout/gpx/Waypoint.h
+    include/osmscout/gpx/TrackPoint.h
+    include/osmscout/gpx/TrackSegment.h
+    include/osmscout/gpx/Utils.h
+    include/osmscout/gpx/Extensions.h
 )
 
 set(SOURCE_FILES
     src/osmscout/gpx/GpxFile.cpp
     src/osmscout/gpx/Track.cpp
-	src/osmscout/gpx/TrackSegment.cpp
-	src/osmscout/gpx/Utils.cpp
+    src/osmscout/gpx/TrackSegment.cpp
+    src/osmscout/gpx/Utils.cpp
+    src/osmscout/gpx/Extensions.cpp
 )
 
 if(LIBXML2_FOUND)
--- a/libosmscout-gpx/include/osmscout/gpx/Extensions.h
+++ b/libosmscout-gpx/include/osmscout/gpx/Extensions.h
@@ -1,9 +1,9 @@
-#ifndef OSMSCOUT_GPX_GPXFILE_H
-#define OSMSCOUT_GPX_GPXFILE_H
+#ifndef OSMSCOUT_GPX_EXTENSIONS_H
+#define OSMSCOUT_GPX_EXTENSIONS_H
 
 /*
   This source is part of the libosmscout-gpx library
-  Copyright (C) 2017 Lukas Karas
+  Copyright (C) 2021 Lukas Karas
 
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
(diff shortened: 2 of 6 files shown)


janbar:
See https://docs.safe.com/fme/html/FME_Desktop_Documentation/FME_ReadersWriters/gpxx/Feature_Types.htm for better explaination of extensions.
The full xsd is here: https://www8.garmin.com/xmlschemas/GpxExtensionsv3.xsd

libosmscout-gpx/CMakeLists.txt
```diff
@@ -1,19 +1,20 @@
...
+    include/osmscout/gpx/Extensions.h
 )
 
 set(SOURCE_FILES
     src/osmscout/gpx/GpxFile.cpp
     src/osmscout/gpx/Track.cpp
-	src/osmscout/gpx/TrackSegment.cpp
-	src/osmscout/gpx/Utils.cpp
+    src/osmscout/gpx/TrackSegment.cpp
+    src/osmscout/gpx/Utils.cpp
```
Framstag, line 17: We normally add an empty *.cpp file - even for header-only *.h files.

libosmscout-gpx/CMakeLists.txt
```diff
@@ -1,19 +1,20 @@
...
-	include/osmscout/gpx/Utils.h
+    include/osmscout/gpx/GpxFile.h
+    include/osmscout/gpx/GPXImportExport.h
+    include/osmscout/gpx/Route.h
+    include/osmscout/gpx/Track.h
+    include/osmscout/gpx/Waypoint.h
+    include/osmscout/gpx/TrackPoint.h
+    include/osmscout/gpx/TrackSegment.h
+    include/osmscout/gpx/Utils.h
+    include/osmscout/gpx/Extensions.h
```
Framstag, line 10: Please also add the header to the meson build files. Meson would also need it for installation.

libosmscout-gpx/include/osmscout/gpx/Extensions.h
```diff
@@ -0,0 +1,64 @@
...
+*/
+
+#include <osmscout/gpx/GPXImportExport.h>
+
+#include <string>
+#include <vector>
+
+#define GPX_EXTENSIONS_V3 "gpxx" /* http://www.garmin.com/xmlschemas/GpxExtensions/v3 */
+
+namespace osmscout {
```
Framstag, line 30: It is possible that we get compiler errors on some platforms with export but no constructor/destructor/virtual table. We'll see ;-)

Framstag (changes requested):

libosmscout-gpx/include/osmscout/gpx/Extensions.h
```diff
@@ -0,0 +1,64 @@
...
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307  USA
+*/
+
+#include <osmscout/gpx/GPXImportExport.h>
+
+#include <string>
+#include <vector>
+
+#define GPX_EXTENSIONS_V3 "gpxx" /* http://www.garmin.com/xmlschemas/GpxExtensions/v3 */
```
Karry, line 28: Instead of global macro, it would be nicer to use `constexpr` in `osmscout::gpx` namespace (`char const *` or `std::string_view`).

libosmscout-gpx/src/osmscout/gpx/Export.cpp
```diff
@@ -479,6 +491,70 @@ bool GpxWritter::WriteWaypoints(const std::vector<Waypoint> &waypoints)
   return true;
 }
 
+bool GpxWritter::WriteExtensions(const Extensions &ext)
```
Karry, line 494: this is not used right? do you plan to use it somehow in future?

libosmscout-gpx/include/osmscout/gpx/Extensions.h
```diff
@@ -0,0 +1,64 @@
...
+  Extensions() = default;
+
+  std::vector<Extension> elements;
+
+};
+
+class OSMSCOUT_GPX_API TrackExtensions : public Extensions {
+public:
+  TrackExtensions() : Extensions(), xmlns(GPX_EXTENSIONS_V3) {}
+  TrackExtensions(const char* ns) : Extensions(), xmlns(ns) {}
```
Karry, line 56: please make this constructor explicit

Karry (commented):
looks good, just few minor comments. 

Moreover, it is possible to use color even for waypoints (with Garmin extensions)?

janbar:
Hi, if the concept suits you I will make the required update tomorrow, introducing the cpp file and constexpr.

libosmscout-gpx/src/osmscout/gpx/Export.cpp
janbar: Yes not necessary because we don't fill simple extensions. To be removed.

libosmscout-gpx/include/osmscout/gpx/Extensions.h
janbar: agreed

libosmscout-gpx/include/osmscout/gpx/Extensions.h
janbar: I will replace it with constexpr and it will be initalized in the object file, as the others constructors.

libosmscout-gpx/CMakeLists.txt
janbar: Don't know Meson, but I will learn ...

libosmscout-gpx/CMakeLists.txt
janbar: oki

libosmscout-gpx/include/osmscout/gpx/Extensions.h
janbar: will be

libosmscout-gpx/CMakeLists.txt
Framstag: Meson files rather intuitive and simple in structure. Adding a source file is a simple addition to a list of files.

janbar:
@Karry, about color  for waypoint. There is no extension for it. In fact the color is defined in the symbol itself, i.e:

"Flag, Red" , "Flag, Blue", "Pin, Red" ...

Only Route and Track have the extensions DisplayColor.

libosmscout-gpx/include/osmscout/gpx/Extensions.h
janbar: There is a tons of manner to handle const, I choosen one.

Karry (approved):
LGTM, thank you!

PR #1075, opened 2021-04-20
introduce utf8 helper

janbar:
Hi, it is a way to resolve issue caused by conversions from utf8 to localized wchar (upper/lower), in particular for location service (search). There is no longer dependency with system locale to handle operations on unicode string, similary using libICU.
The characters map has been generated from UTF8. 8.0, and I include all codepoint related to Latin, European languages + armenian + georgian and some others known to have upper/lower matching. The others codepoint are simply pass throught.
The code is fully static with minimal footprint. For instance it provides 3 functions impacted by the issue ( upper, lower, normForLookup ). Also now the function "normalize" remove duplicate spaces and breakers.
It just for preview, surely need rework.

### DIFF
--- a/libosmscout/include/osmscout/util/utf8helper_charmap.h
+++ b/libosmscout/include/osmscout/util/utf8helper_charmap.h
@@ -1,0 +1,106 @@
+/*
+  This source is part of the libosmscout library
+  Copyright (C) 2021  Jean-Luc Barriere
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 2.1 of the License, or (at your option) any later version.
+
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307  USA
+*/
+
+#ifndef UTF8HELPER_CHARMAP_H
+#define UTF8HELPER_CHARMAP_H
+
+#include <cstdint>
+
+namespace utf8helper
+{
+
+constexpr int None          = 0x00;     // no category
+constexpr int IsSpace       = 0x01;     // charcater is space (breaking or non-breaking)
+constexpr int IsBreaker     = 0x02;     // character is breaker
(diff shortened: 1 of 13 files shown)


libosmscout/include/osmscout/util/utf8helper_charmap.h
```diff
@@ -0,0 +1,78 @@
...
+{
+
+constexpr char None         = 0x00;
+constexpr char IsSpace      = 0x01;
+constexpr char IsBreaker    = 0x02;
+constexpr char IsControl    = 0x04;
+
+using byte = unsigned char;
+
+using codepoint = unsigned int;
```
Framstag: Shouldn't you not better use something like uint8_t...uint32_t?

libosmscout/src/osmscout/util/utf8helper.cpp
```diff
@@ -0,0 +1,276 @@
+#include <osmscout/util/utf8helper.h>
+#include <osmscout/util/utf8helper_charmap.h>
+
+#include <string>
+#include <cstring>
+
+namespace utf8helper
+{
+
+static std::string parse(const std::string& text, codepoint (*func)(const character*, char context)){
```
Framstag, line 10: It would be nice to have some description for what the code is doing. Also some external link to a more in deep documentation would be helpful.

What about some tests, especially for invalid input to make sure that corner cases are handled correctly?

Framstag:
Reviewed, please give me a hint, if I should merge.

Framstag:
Also, Visual Studio breaks on your code.

libosmscout/include/osmscout/util/utf8helper_charmap.h
janbar: Yes It would be better. It is an old habit to avoid compiler failure with old platforms. To be fixed including stdint.

janbar:
@framstag, I have to fix the issue you had with Visual Studio, and add some code for test. I will let you know when it will ready.

Framstag:
@janbar Take your time. I should stress also, that this is a great new feature! Thanks for implementing it.

janbar:
@Framstag , I fixed the issue with the microsoft compiler. Seems `char[]` is not char* on this platform. Anyway I made it better by testing with windows. Still to make some primitives to test/stress it. As the utf8 standard is backward compatible with ascii/iso8, you can make it eat a binary file without get an issue. The best effort is to maintain the charmaps to do what we need, i.e generate efficient tokens with transliterate. Probably removing punctuation etc.

janbar:
@Framstag , I have no more change to add. I tested this successfully with android app and now searching works as expected, and it should work anywhere for any body (locale).
About performance it processes more than 25GB per second per core i7, so much more than iconv. Also I tried to design it as cpp class, but the overhead is 25%. IMO no need for overkill engineering, it is OK as is.

Issue #1074, opened 2021-04-20
Enormous memory usage during Belarus import


Karry:
During recent import of [Belarus OSM extract provided by Geofabrik](http://download.geofabrik.de/europe/belarus-latest.osm.pbf) import tool was killed by OOM (out of memory killer) on my machine (16 GiB ram + 22 GiB swap). It happened during optimizing areas for low zoom levels (step 21). Problematic type is `boundary_administrative`...

Commands to reproduce:
```
wget http://download.geofabrik.de/europe/belarus-latest.osm.pbf
wget http://download.geofabrik.de/europe/belarus.poly

Import \
  -d \
  --eco true \
  --typefile map.ost \
  --rawCoordBlockSize 60000000 \
  --rawWayBlockSize 4000000 \
  --relMaxWays 4096 \
  --altLangOrder en \
  --destinationDirectory europe-belarus \
  --bounding-polygon belarus.poly \
  belarus-latest.osm.pbf
```

Question is, if there is some bug created recently or data was just modified the way that is causing troubles? Other countries seems to be fine. Removing flag `OPTIMIZE_LOW_ZOOM` from `boundary_administrative` type workaround this issue.

Framstag:
I reworked the complete OptimizeAreas step, but not any of the GenOptimizeXXX steps. Looking at the history https://github.com/Framstag/libosmscout/commits/master/libosmscout-import/src/osmscout/import/GenOptimizeAreasLowZoom.cpp there were recent smaller changes. If there is a bug, than because of the (in principle structural) changes to the transformation code I also recently did. However I did not change the transformation code itself, I just did heavy restructuring, so if this broke it, it should also be easy to fix.

Karry:
I am expecting that problem is in the data file. But have no time for investigating it yet.

Karry:
I randomly pickup four months old commit e185728cecce05fef59e12ec5ee928e2c3fff612 and it imports this data sets without issues. I will continue with git bisect, because I have no idea where is the problem.

Karry:
I made simple bisect script, import is executed in cgroup with limited memory to 8 GiB (to avoid huge swap usage and fail fast):

```bash
#!/bin/bash
set -xe
cd $(dirname $0)
mkdir -p cmake-build-release
cd cmake-build-release

cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_UNITY_BUILD=OFF ..
make -j 8 Import

[...]

PR #1076, opened 2021-04-23
Import merge areas fix

Karry:
Fix for https://github.com/Framstag/libosmscout/issues/1074

It was hidden properly :-D


Framstag:
Interesting that my changes have triggered that, should have been a problem at other places, too.

Nevertheless, thanks for finding it and sorry for crearting the problem at all!

PR #1077, opened 2021-04-26
Fix/utf8 charmap

janbar:
- I found a bug for string containing only 1 codepoint.
- Small letter 'I' with Diaeresis was translated to capital letter 'I' instead small letter
- Remove transliterate to us7ascii: no usage

### DIFF
--- a/Tests/src/StringUtils.cpp
+++ b/Tests/src/StringUtils.cpp
@@ -126,6 +126,26 @@ TEST_CASE("Parse illegal UTF8 sequence")
   REQUIRE(transformed == u8"\xef\xbb\xbf\x2f\x2e\x2f");
 }
 
+TEST_CASE("Parse one UCS1 sequence without BOM")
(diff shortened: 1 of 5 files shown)


Framstag:
Since it sounds like you found an error, does it make sense to add a specific test for this error?

janbar:
It is a case we don't use for now: processing one letter. I found it checking the letter 'I' with Diaeresis. I did tests on unlikely cases, but I missed the simplest one (see Murphy's laws). You must always remember these laws, thanks to Murphy.
So I add a new test for the case ...

Framstag:
@janbar 

See https://github.com/Framstag/libosmscout/issues/1082

It seems like some of the StringUtils fail under MSVC/Windows

Can you take a look what is the problem? We can of course disable the tests if there is some fundamental problem.

janbar:
I will check that ...

PR #1079, opened 2021-04-27
fix coord buffer leak, make it member of MapPainter

Karry:
When I tried to run unittests with address sanitizer, it shows me that coord buffer is leaking from MapPainter. It seems to me that there is no reason to have it as pointer passed by constructor, it may be (protected) member of MapPainter.

### DIFF
--- a/libosmscout-map-agg/include/osmscout/MapPainterAgg.h
+++ b/libosmscout-map-agg/include/osmscout/MapPainterAgg.h
@@ -207,7 +207,6 @@ namespace osmscout {
     explicit MapPainterAgg(const StyleConfigRef& styleConfig);
     ~MapPainterAgg() override;
 
-
     bool DrawMap(const Projection& projection,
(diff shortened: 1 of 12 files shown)


Framstag:
I have seen this recently, too, when I refactored the transformation code. I was unsure, why we passed the coordBuffer to the base class at all. Looks like some problem we solved at that time, that now has disappeared. I can only thing of the OpenGL backend enforcing this, but that one does not implement MapPainter (anymore?).

Framstag:
Btw.: I configured codescene.io to scan the libosmscout repository. I was not able to extract a report from it, but the findings are interesting. It shows that parts of the code are starting to get worse, especially MapPainter. The findings are in line with my expectations. The findings also show, that large parts of the system are in rather good state :-)

I haven't yet found time to make in deep use of the tools, but perhaps you might take a look, too (if you can).

PR #1084, opened 2021-05-10
Improve utf8

janbar:
Hi, here an overkiller for utf8helper. It is modular and allows injection of customized tranformation.
@Karry please let me know if it could help to improve postal address processing or others. In some case you could need to customize some transformation (i.e normalize/translate etc...).

### DIFF
--- a/libosmscout/include/osmscout/util/utf8helper.h
+++ b/libosmscout/include/osmscout/util/utf8helper.h
@@ -20,7 +20,10 @@
 #ifndef UTF8HELPER_H
 #define UTF8HELPER_H
 
+#include <osmscout/util/utf8helper_charmap.h>
+
 #include <string>
+#include <vector>
 
 namespace utf8helper
 {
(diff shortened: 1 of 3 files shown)


libosmscout/include/osmscout/util/utf8helper.h
```diff
@@ -35,6 +38,209 @@ extern std::string UTF8Capitalize(const std::string& text);
...
+ * discard the sequence.
+ */
+using transform = codepoint (*)(const character*, int context);
+
+extern codepoint transform_nop(const character*, int);
+extern codepoint transform_upper(const character*, int);
+extern codepoint transform_lower(const character*, int);
+extern codepoint transform_capitalize(const character*, int);
+extern codepoint transform_normalize(const character*, int);
+extern codepoint transform_transliterate(const character*, int);
```
Karry, line 56: Just nitpicking - in library is usually used CamelCase for methods and class names with first big letter.

libosmscout/include/osmscout/util/utf8helper.h
```diff
@@ -35,6 +38,209 @@ extern std::string UTF8Capitalize(const std::string& text);
...
+   * @param pos
+   * @param n
+   * @return the std::string object
+   */
+  std::string substr(size_t pos, size_t n = (-1)) const;
+
+private:
+  parser m_p;
+  storage_type m_data;
+  size_t m_rawSize;
```
Karry, line 241: Hungarian notation (`m_` prefix for members) is not usually used in library.

Karry (approved):
I am fine with the change in general, but don't review it deeply. 

 - just minor nitpicking that method and fields names don't match to style used in library
 - first two commits are merged already, right? Can you rebase this MR please?

To your question. I am not aware of any place where custom transformation would be useful. But we may enable "transliteration" as default now (it is not depending on iconv anymore) and we may try to use transliteration even for Marisa-based search (prefix name search).

PR #1100, opened 2021-07-20
via_ferrata_scale tag may be used with -/+ suffix

Karry:
see https://wiki.openstreetmap.org/wiki/Key:via_ferrata_scale

via_ferrata_scale is used describe the difficulty of a via ferrata...
... You can add a + or a - after the value to indicate that the via
ferrata is on bottom or top of the grade.


Framstag:
Looks like we someday need to implement regular expressions for this ;-)

PR #1105, opened 2021-08-08
Merge osmscout-map and osmscout-map-iOSX public header with osmscout framework headers

vyskocil:
In order to make osmscout header resolve to work with the include/osmscout directory split into the osmscout, osmscout-map and osmscout-map-iOSX framework we have to copy the public header in the include directory of osmscout framework

### DIFF
--- a/libosmscout/CMakeLists.txt
+++ b/libosmscout/CMakeLists.txt
@@ -137,6 +137,31 @@ set(HEADER_FILES_NAVIGATION
     include/osmscout/navigation/SpeedAgent.h
     include/osmscout/navigation/VoiceInstructionAgent.h
     include/osmscout/navigation/LaneAgent.h)
+	
+set(HEADER_FILES_MAP
+	../libosmscout-map/include/osmscout/MapImportExport.h
+	../libosmscout-map/include/osmscout/oss/Parser.h
+	../libosmscout-map/include/osmscout/oss/Scanner.h
+	../libosmscout-map/include/osmscout/LabelLayouter.h
+	../libosmscout-map/include/osmscout/MapPainter.h
+	../libosmscout-map/include/osmscout/MapParameter.h
+	../libosmscout-map/include/osmscout/MapData.h
+	../libosmscout-map/include/osmscout/MapService.h
+	../libosmscout-map/include/osmscout/LabelProvider.h
+	../libosmscout-map/include/osmscout/LabelPath.h
+	../libosmscout-map/include/osmscout/Styles.h
+	../libosmscout-map/include/osmscout/StyleDescription.h
+	../libosmscout-map/include/osmscout/StyleConfig.h
+	../libosmscout-map/include/osmscout/StyleProcessor.h
+	../libosmscout-map/include/osmscout/DataTileCache.h
+	../libosmscout-map/include/osmscout/MapTileCache.h
+	../libosmscout-map/include/osmscout/MapPainterNoOp.h
+	../libosmscout-map/include/osmscout/MapFeatures.h
+)
+
+set(HEADER_FILES_MAP_IOSX
+	../libosmscout-map-iOSX/include/osmscout/MapPainterIOS.h
+)
 
 set(SOURCE_FILES
     src/osmscout/ost/Parser.cpp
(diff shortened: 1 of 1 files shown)


Framstag:
Hi @vyskocil Please let's discuss the change. Without in deep knowledge of specialities of Mac OS X framework development, your change looks very strange. 

In principle, we have a number of libraries that depend on each other, building in general a tree-like dependency structure. What is special in the case of libosmscout may be that they share the same top level include directory (include/*osmscout*/...). So installed include directories are not disjunct.

I assume that there are additional constraints on OS X that trigger your change? Maybe there is another solution? If I look at Qt for example, they also share the same include/*qt5*/ top level directory, but they have another layer per package (in fact per module which seem to be smaller than my ArchLinux packages):
* include/qt5/*QtCore*
* include/qt5/QTMultimedia

Would this solve possible problems?

vyskocil:
@Framstag this is exactly the point, I didn't found another workaround to the fact that there are several frameworks sharing the same top level include directory without having to modify the include files. But if we could update the directory structure the Qt solution is of course a good one !

Framstag:
I'm in for it. This would be of course some brain-dead effort for the library and all of its users, but if we are more conform to some expectations, it i better than dirty tricks.

@Karry what is our opinion?

vyskocil:
Maybe the simplest way would be to leave osmscout core includes as is and add one or more level of sub directory to the other modules, for example add /map in osmscout root directory for osmscout-map and /map/iosx for osmscout-map-iOSX ?

Karry:
Hi. I don't understand the reason for this change. When I install libraries to some prefix (for example `/opt/osmscout`), all public headers (from libosmscout, libosmscout-map, libosmscout-client-qt) are installed to the same directory on Linux: `<prefix>/include/osmscout`. It is different on MacOS?

But in general, I am fine with Qt-like directory tree. It will be nicer and less error prone.

vyskocil:
hi @Karry !
The issue with Apple (Mac and iOS) is that each module (libosmscout, libosmscout-map, libosmscout-map-iOSX, libosmscout-qt,...) is packaged in a separate "framework" which is a bundle with binary, includes and metadata put aside in a directory structure. 
A iOS app may for example need OSMScout.framework, OSMScoutMap.framework and OSMScoutMapIOSX.framework to display a map. Each of these frameworks contain a Headers folder which contain a osmscout folder. When we want to include a file from the osmscout root dir it would only be searched in the first listed framework (usually OSMScout.framework) and if not there the compiler would not search it in the next framework in the list...

vyskocil:
In fact on OSX and iOS the framework name (without .framework) is the first part of the header search path.

For example the following include will search MapPainter.h at the root of the OSMScout Header folder (OSMScout.framework/Header/MapPainter.h) :  

```#include <osmscout/MapPainter.h>```

But as this file is provided by OSMScoutMap.framework the include should have been : 

```#include <osmscoutmap/MapPainter.h>```

And "traditionally"  the MacOS FS is case sensitive (the file names retain the case) but it is not taken into account...
OSMScout.framework and osmscout.framework should be the same file but there is a option to format the FS with "real" case sensitive mode and then we should take care of the case... then I think I should rename the framework to lowercase to be sure !

vyskocil:
There is also something called "Umbrella Framework" that could merge several frameworks into one but as I understand it, this would not solve our naming problem and using this would mean that we would have to deliver several combination of our framework put together (osmcout alone, osmscout + osmscoutmap + osmscoutmapiosx, osmscout + osmscoutmap + osmscoutqt,...)

Framstag:
What would be the easiest way to test this without a complete big bang over all sources?

vyskocil:
Hello @Framstag I just made another proposition with https://github.com/Framstag/libosmscout/pull/1106. I did the job to rename include/osmscout folder for libosmscout-map includes to include/osmscoutmap and include/osmscoutmapiosx for libosmscout-map-iosx, then I fixed the whole compilation for iOS and OSX (the libs and the samples apps).
But there are still some work to fix targets that are not compiled for iOS or OSX. I don't have the devices or tools to compile for Android, Windows, Linux.

vyskocil:
I will close this pull request as the work is done in another branch

PR #1106, opened 2021-08-15
Modularize OsmscoutMap to fix osmscout, osmscoutmap and osmscoutmapiosx OSX and iOS frameworks

vyskocil:
Move osmscout-map includes from include/osmscout to include/osmscoutmap and move osmscout-map-iosx includes from include/osmscout to include/osmscoutmapiosx, then fix compilation on iOS, OSX and Apple XCode samples.


vyskocil:
There is one thing that I could not solve properly and I had to put the following line in libosmscout-map CMakeLists.txt : 
```file(COPY ${CMAKE_CURRENT_BINARY_DIR}/include/osmscout/MapFeatures.h 
        DESTINATION ${CMAKE_CURRENT_SOURCE_DIR}/include/osmscoutmap)```
This is bad because it copy a file from build dir to src include dir... but the TEMPLATE line : 
```TEMPLATE ${CMAKE_CURRENT_SOURCE_DIR}/include/osmscoutmap/MapFeatures.h.cmake```
generate the MapFeatures in the wrong build subdir and I couldn't find how to fix it.

Karry (commented):
It is huge change, but mostly straightforward. I believe that it make sense. I send small fix to cmake macro and another batch of import fixing: https://github.com/vyskocil/libosmscout/pull/1 Before merge even Meson build will need some tunning I guess. 

I'm still thinking about naming, it is little bit odd to me:
 - directory in the repository is `libosmscout-map`
 - include dir is `osmscoutmap`
 - cmake library is named `OSMScoutMap`
 - output so file is `libosmscout_map.so` (`libosmscout_mapd.so` for debug build)

I would at least remove underscore from library name (in future MR), what do you think @Framstag ? :-)

Later it would be great to do the same change for other libraries, to be consistent...

Framstag:
Yes, this is the chance to fix some naming inconsistencies. I looked for naming conventions for unix library names, but di dnot find any for the actual name.

The Qt libraries use upper case, but no "-". There seems to be some use of "-"..., but most do not use it. Lowercase seems to be an obvious trend. Same for includes.

So for unix no "-" and all lower case in the library name.
For iOSX we should follow their conventions, too. But they are up to @vyskocil to handle ;-)

I also would like to have short include directory names for the basic libraries, osmscoutmapiosx is a monster. We could shorten the common prefix osmscout, but that might collide with other libraries. Something like "osms" is already risky to use. So it is the way it is. I thus would not use additional "-" to make the includes even longer (it also does not look "common").

The "-" in the directory name was likely choosen by me to be better to read. I would not change this.

For cmake we also should stay with the conventions, I found https://stackoverflow.com/questions/48524359/naming-convention-for-components-and-namespaces-in-cmake for this. Using an osmscout namespace sound interesting and might improve readability? But I'm no expert in cmake.

Meson does not know of namespaces. It has modules (for example for Qt), but these would be part of meson itself and not custom. If you look at meson wrap (repository of meson based build depedencies): https://mesonbuild.com/Wrapdb-projects.html lower case and "-" and "_" seem to be OK.

That means (in this order?)?
* Use namespaces, if available, else use osmscout prefix.
* Stay lowercase
* Use "-2 or "_" only if it is common in the environment. Use it only, if the code directories already use it (so no osm_scout_map, but possibly osmscout-map_iosx).
* Use "lib" prefix only if it is common in the environment. It is a technical prefix but not part of the name of the library. It is only done in the filesystem to separate the libraries from the applications.
* Try to stay short but readable.

Regarding adapting meson build files: This is also a simple repetitive task. And does not need in deep know how of meson. I can do it, but it is ok, if somebody else does it.

vyskocil:
@Framstag you should be aware that on Apple platforms the names of the top include dirs (osmscout, osmscoutmap, osmscoutmapiosx,...) should also be the base names of the frameworks but case is not relevant (OSMScout.framework, OSMScoutMap.framework, OSMScoutMapIOSX.framework,...).
I agree with you that osmscoutmapiosx is ugly... maybe I should use apple instead of iosx ?

Framstag:
> I agree with you that osmscoutmapiosx is ugly... maybe I should use apple instead of iosx ?

It is ugly because of its length, not because of iosx. Personally, I find iosx even a nice naming, though their various platform have changed their name recently?

What is missing for this to get more etsted7merged? Just fixing meson?

PR #1110, opened 2021-08-18
GHA: Use Ninja with CMake in msys2

Biswa96:
Ninja is faster than make in msys2

### DIFF
--- a/.github/workflows/build_and test_on_msys.yml
+++ b/.github/workflows/build_and test_on_msys.yml
@@ -23,14 +23,14 @@ jobs:
         with:
           update: true
           install: "mingw-w64-x86_64-toolchain mingw-w64-x86_64-libtool mingw-w64-x86_64-libiconv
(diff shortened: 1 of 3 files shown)


Framstag:
Sorry for the late reply. I just have seen, that I need to activate the builds manually, since this is the first pull request from you.

Patch looks good to me, we'll see, if the builds run.

Thanks for this pull request!

Framstag:
I love windows. Seem like things work with ninja in principle, but we now get a compiler error (which AFAIK we already had before) we did not have before. Is this something you can reproduce or even fix? It can (it likely is) be unrelated to our changes, but a fix would be helpful to do not have an e´increased count of failing builds (since I have no time to look at it myself).

Biswa96:
Done.

Framstag:
Perfect. I love to see more pull requests :-)

PR #1119, opened 2021-09-15
use semitransparent color for leisure_marina and leisure_fishing

Karry:
Some marina objects covers larger areas, sometimes even land partialy.
When it is rendered with opaque color, multiple structures may
be hidden (pier, breakwater...). For that reason, it is better
to render marina with semitransparent color, silimar as Carto style does.

https://wiki.openstreetmap.org/wiki/File:Rendering-leisure_marina.png


Karry:
One example of such problematic marina is in Bari, Italy: https://www.openstreetmap.org/way/338350885

PR #1147, opened 2021-11-05
Opengl includes

Karry:


Karry:
Why linking is failing on Visual Studio 2019 / CMake ...? :-(

Karry:
...and build is green again :tada: 

OpenGL renderer will deserve some love in future

PR #1154, opened 2021-11-11
Install locations

Karry:
I cleaned installed paths a bit, to use more standard locations. At least for Linux... It would be great to do similar review on Windows and MacOS. 

 - remove lib prefix from OSMScout2 organization name and move installation directory of translation files to
`<install prefix>/share/omscout/OSMScout2/translations` accordingly
- use standard install path for cmake configuration files
on unix: `<prefix>/lib/cmake/libosmscout`
otherwise `<prefix>/CMake`
- install stylesheets to `<prefix>/share/osmscout/stylesheets`
- don't install demos and tests to bin directory
instead:
install demos to `<prefix>/share/osmscout/demos`
install tests to `<prefix>/share/osmscout/tests`

so, structure of installed files now looks like:
```
/bin:
  BasemapImport
  BasemapImport-1.1.1
  DumpData
  DumpData-1.1.1
  Import
  Import-1.1.1
  OSMScoutOpenGL
  OSMScoutOpenGL-1.1.1
  OSMScout2
  OSMScout2-1.1.1
  PublicTransportMap
  PublicTransportMap-1.1.1
  StyleEditor
  StyleEditor-1.1.1

/include:
  osmscout
  osmscout-test
  osmscoutclientqt
  osmscoutgpx
  osmscoutimport
  osmscoutmap
  osmscoutmapagg
  osmscoutmapcairo
  osmscoutmapopengl
  osmscoutmapqt
  osmscoutmapsvg

/lib:
  cmake/libosmscout/
    libosmscoutConfig.cmake
    libosmscoutConfigVersion.cmake

  libosmscout*.so*

/share:
  osmscout:
    demos
    docs
    OSMScout2
      translations
    stylesheets
    tests
```


Karry:
Do we need to install `include/osmscout-test` and `include/osmscoutimport` ? I would not expect that users of library will develop  against these libraries... Moreover, `osmscoutimport` contains just one file: `ImportFeatures.h`, because there is `SKIP_HEADER` argument...

PR #1158, opened 2021-11-21
Filter useless turn instructions

Karry:
When there is no other way in junction that can be used for driving from the junction,
we don't have to signalize turn.
There may be junction with one-way in forbidden direction
or way with forbidden type for current vehicle.


Karry:
This change for example removes "Turn right" instruction when entering [link way before roundabout](https://www.openstreetmap.org/way/216863194#map=19/49.99541/14.58717)...

Screenshot_20211122_102833

PR #1210, opened 2022-01-30
Clean up and enhance the Apple sample app

vyskocil:
Add a Readme.md file
Add a sample compiled map of the Greater London
Add sample app icons


Framstag:
It looks like you are also checking in a concrete database including the import log file. Is this by accident?

vyskocil:
Hi @Framstag ! I put a small sample database in the Resources directory of the sample app to ease the start with osmscout for newcomers but if you think It shouldn’t be included, I might add instead some explanations in Readme.md how to compile and install a osm database in the sample project.

Framstag:
Hi @vyskocil 

The problem is, that the database will not be in the current format if we change something in the format. I'll merge now, but it would be nice if switch to the user generating the database.

PR #1211, opened 2022-01-31
Fix/utf8helper

janbar:
Hi, here a fix for the utf8helper. 

The rawsize member should be updated with the size in bytes of the stored code point.
Applying a transform operation could change the original code point to one with a different byte size. Therefore, the parser should store the size of the final code point resulting from the transform operation, not the size of the input code point.

The rawsize is used to size the buffer output string (ToStdString). Actually it is larger than required in some case, or can be shorter for border cases. Added test cases check that.

### DIFF
--- a/Tests/src/StringUtils.cpp
+++ b/Tests/src/StringUtils.cpp
@@ -234,6 +234,19 @@ TEST_CASE("Normalize for lookup")
   REQUIRE(transformed == "baker street");
 }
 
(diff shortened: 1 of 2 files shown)


janbar:
Some build failed. I guess utf8helper is hidden. I am updating the test case ...

Karry (approved):
Good catch. How did you find it? I have limited knowledge of utf8, but it is clear that test case is broken without fix...

PR #1212, opened 2022-01-31
code style in utf8helper

Karry:
just minor code style:
 - brackets even for one-line expression
 - minimize variable scope
 - use `[[maybe_unused]]` annotation, not `(void)` cast


Framstag:
@Karry Still in draft?

Karry:
> @Karry Still in draft?

no. I just wanted to avoid conflicts...

Issue #1108, opened 2021-08-16
Use area visual center for label


Karry:
We are using bounding box center for placing label on map. It is simple and fastest method. But for some natural areas, like islands or lakes its label may be located outside the area. For example ["Císařský ostrov" islet in Prague](https://www.openstreetmap.org/way/543187110):

![Snimek_20210806_001](https://user-images.githubusercontent.com/309458/129634488-48da6ffa-cf64-41dc-8b81-331dac001b3a.png)

It would be great to use some precise method for placing label on areas (rings). We can try to use [Polylabel algorithm from Mapbox](https://blog.mapbox.com/a-new-algorithm-for-finding-a-visual-center-of-a-polygon-7c77e6492fbc). This method should be fast, [code is available in C++](https://github.com/mapbox/polylabel) under terms of ISC license, that should be compatible with LGPL. It should be easy to adapt this code to libosmscout.

Not to degrade performace, this algorithm may be used just for some types. Or label position (visual center) may be computed during import time and stored in database.

Framstag:
Yes, sounds interesting. I think import time would be wise for calculation. We can add it as an optional point to the polygon (triggered by some flag). Since we have to handle a huge number of polygons on import, we would need a very selective filter to even start the calculation - likely size (bigger buildings and bigger?) and possibly even types?

Framstag:
I also think for relation based polygons there is a way to define the center. So this can be also evaluated?

PR #1215, opened 2022-02-07
Evaluate ring center during import, use it for labels

Karry:
Solution for https://github.com/Framstag/libosmscout/issues/1108

 - `Area::Ring` center (pole of inaccessibility) is evaluated  during import, it is stored just in case when it is far enough from bounding box center. Area data file increased just ~1% in size (4 MiB increase for Czech Republic database). Later we may even filter types where there area center is stored in database to reduce size again. But I think that it is fine now. I put the evaluation to `OptimizeAreaWayIdsGenerator` to have it available in area and even optimized areas...
 - Database format was changed, so I increased format version.
 - **When ring have optional "center" value, it is used as label location**

before:
Screenshot_20220207_095344

after:
Screenshot_20220207_094606


Karry:
Hi @vyskocil . You recently added sample database to `./Apple/OSMScoutOSX/OSMScoutiOS/Resources/map.osmscout/`, it should be updated with every database format change, right? It would be great to prepare similar script for update as this one: https://github.com/Framstag/libosmscout/blob/master/Tests/data/testregion-generate.sh

vyskocil:
@Karry It was a bad idea indeed and @Framstag asked if I could remove it from the sample app, I’ll add some explanations how to install a sample database in the right place and remove the embedded one.

PR #1216, opened 2022-02-08
downgrade pip to fix MS VS2019 / Meson github action

Karry:


Karry:
see https://github.com/mesonbuild/meson/issues/9955 
thanks @eli-schwartz

PR #1225, opened 2022-02-27
Refactored drawing steps

Framstag:
Add some more steps, moved some code between steps.

Step "CalculateWayShields" is abviously a "slow one" for drawing levels <= level 10


Karry (approved):
good idea to split it

PR #1231, opened 2022-03-27
clickable map icons - first attempt

Karry:


Framstag:
The code changes are rather complex and difficult to grok. So a few questions (based on assumptions):

I assume the goal is to make things in the map clickable, which requires being able to identify the object under the cursor. It looks like you try to implement this by passing more context information to the objects on the map (LabelLayouter). There is at least one other option: Query the Renderer or even the database regarding the object(s) under the mouse. If there are multiple objects, one can still apply a heuristic based on the type. This way you might get cleaner and simpler APIs and less coupling of things. It looks no got to see that many APIs enhanced regarding a new cross-cutting concern. This way one may separate "looks like a button", "identify clickable object under the cursor" with "trigger some action".

Are my assumptions regarding your goal right? Can you explain your approach and compare it to other possible approaches?

Karry:
Hi Tim. You are right. My goal is to make Qt MapWidget more interactive, allow to detect if on some screen position is icon/symbol and show information about the object belongs to the icon... The way you are describing is really simple, and we already support it, by Qt `MapObjectInfoModel`. But problem is that this model have no information what objects are rendered with icon/symbol. For that it would need to process stylesheet. It is still possible. But on some zoom levels, there may be multiple close objects that each may be rendered with icon. And just the layouter have the information about what icons are rendered and what icons are dropped because of conflicting position...

Karry:
Btw the big commit "Solve sonnar issues" is unrelated regactoring. I may send it in separate MR...

Framstag:
> It is still possible. But on some zoom levels, there may be multiple close objects that each may be rendered with icon. And just the layouter have the information about what icons are rendered and what icons are dropped because of conflicting position...

OK, I see. I see that DatabaseId and ObjectRef together are such id of an object. The problem is, that there are very specific regarding the storage layer. It is likely already broken or at least leaking somewhere but I still find it interesting to let the UI not know too much about this. I still should be able to generate vector tiles from the database and still should be able to switch storage layer and use the renderer.

Since the MapPainter now gets a databaseId. Have we already enforced by the support for multiple databases that each database has its own MapPainter instance (We currently pass a StyleConfig to the Painter, which has a TypeConfig. A TypeConfig is specific for a database...)? Or would this be a new constraint?

Karry:
Closing this one. It is too complicated. Your approach is more clear and works good enough. It is implemented here: https://github.com/Framstag/libosmscout/pull/1233

PR #1233, opened 2022-04-03
Clickable map icons

Karry:
Make Qt MapWidget more interactive - optionally make map icons / symbols active, detect click (tap) on them, highlight them and provide details about map object (name, phone, web...) to QML UI... 

As Tim suggested here https://github.com/Framstag/libosmscout/pull/1231#issuecomment-1081034421 information about icon object is not transferred thru renderer (it is one of the most complicated part of library), but widget is trying to guess it, by `IconLookup` utility. As result, lookup process may choose different icon that is on the map (layouter work differently a bit), but result is good and errors are not so disruptive...

### DIFF
--- a/OSMScout2/qml/main.qml
+++ b/OSMScout2/qml/main.qml
@@ -114,6 +114,7 @@ Window {
             Layout.fillHeight: true
             focus: true
             renderingType: "plane" // or "tiled"
+            interactiveIcons: true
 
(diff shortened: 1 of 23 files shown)


Karry:
How feature looks like in Sailfish UI: https://www.youtube.com/shorts/Nix_32Qc6ys ;-)

PR #1237, opened 2022-04-08
use PerformanceTest demo for automated testing

Karry:
As PerformanceTest demo support various rendering backends
and renderer is used multiple times, it is good tool for
automated testing. It may help us discover introduced
regressions, memory issues (with asan) and other bugs
that are not covered by other tests.


Karry:
btw, it would make sense to me move PerformanceTest to Tests... what do you think @Framstag ?

Framstag:
> btw, it would make sense to me move PerformanceTest to Tests... what do you think @Framstag ?

I'm unsure. It is at the same time a manually called test tool and you also now use it as automatic test tool (which is a goof idea btw.). For the first case I would never look under "tests" ;-) Perhaps we move some stuff under "testtools" to separate them from demos?

Framstag:
Regarding the Visual Studio builds: Perhaps a problem with the internal cache used?

Karry:
So, I decided to move it to test directory. As it is named PerformanceTest, it makes sense to me :-)

> Regarding the Visual Studio builds: Perhaps a problem with the internal cache used?

I have no idea what is wrong with MSYS build. Exit code may signalize dynamic linking issue. But there is no more information in log :-( I just disable these tests for now. The same for opengl backend on Mac. Discussions on internet says that it is issue with VM, but without solution. And tests with asan are disabled because graphical backends (agg, cairo, Qt) are leaking memory... I will look on these issues later...

PR #1241, opened 2022-04-15
Implementation of MODULE statement to the OST files.

Framstag:
Adding some modules for testing (further modularisation likely makes sense, if the feature completely works)


Framstag:
* It tries to lookup modules by adding the module name to the directory of the parent file + ".ost". 
* No recursion or double import checks (could be added, but requires further actions)
* Apply calls in order of scanning to the type config.
* Module imports can be applied between sections

PR #1242, opened 2022-04-16
Implementation of MODULE statement to the OSS files.

Framstag:
Adding some modules for testing (further modularisation likely makes sense, if the feature completely works)

### DIFF
--- a/StyleEditor/src/StyleAnalyser.cpp
+++ b/StyleEditor/src/StyleAnalyser.cpp
@@ -81,7 +81,7 @@ void StyleAnalyser::onContentsChanged()
 void StyleAnalyser::update(QString content)
 {
   osmscout::StyleConfigRef styleConfig=std::make_shared<osmscout::StyleConfig>(typeConfig);
-  styleConfig->LoadContent(content.toStdString());
+  styleConfig->LoadContent("main.oss", content.toStdString());
 
   QSet<int> errorLines;
   QSet<int> warningLines;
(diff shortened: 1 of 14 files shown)


Karry (approved):
I like it :-)

Karry:
Ups, it seems that this change breaks area rendering with standard stylesheet and zoom level >= 19

Framstag:
Hmmm, how? Any idea?

Karry:
Not yet. I am debugging broken label layouter. I introduced bug with my latest refactoring :-/

Framstag:
> Not yet. I am debugging broken label layouter. I introduced bug with my latest refactoring :-/

The changes also may subtly change initialization of StyleConfig. I removed a reset() method which would have effect if the same StyleConfig is loaded again (but I did not see such code). Also Postprocess() is called multiple times  while it might not have an effect, this is something that should be changed (by adding another parameter to LoadContent (subModule true/false).

Karry:
I don't see any obvious bug, but difference is visible even when you run `OSTAndOSSCheck` tool with `--analyze` argument, on style with "modules" and without it...

```
./Tests/OSTAndOSSCheck --analyze ../stylesheets/map.ost ../stylesheets/standard.oss
```
Screenshot_20220418_231958

> ... Also Postprocess() is called multiple times while it might not have an effect

I am not so sure, there is computed `size_t maxLevel=0;`, and that may be the issue... It should be computed just once I guess.

PR #1252, opened 2022-04-24
fix width of the route

janbar:


janbar:
The width of the route is sized to 6 meters for high zoom.
image

Framstag:
For style sheet changes I assign @Karry as reviewer, since he will make use of our changes, too. You can assign him directly for further changes. I can live with the results as long as you two agree (though I loved my color scheme :-()

janbar:
Oki. Also this change doesn't update the color. The screenshot is only a sample to show you the width of the route on high zoom.

Karry (approved):
LGTM. It makes sense to keep _route width similar to street/way under it.

Karry:
Btw, I'm not using upstream stylesheets in my app, but slightly modified versions, with different color palette and few small modifications. But with new "module" support, I will try to sync them...

PR #1253, opened 2022-04-24
fix width of the route for style winter-sports

janbar:


stylesheets/winter-sports.oss
```diff
@@ -499,7 +499,7 @@ STYLE
     ICON { name: target;} */
 
   [MAG world-] {
-    [TYPE _route] WAY {color: @routeColor; displayWidth: 1.5mm; priority: 100; }
+    [TYPE _route] WAY {color: @routeColor; displayWidth: 1.5mm; width: 6m; priority: 100;}
```
Karry, line 502: Would not be better to use `MODULE "route"` ? I know that it is one-liner, but still, we may de-duplicate stylesheets a bit.

Karry (approved):
LGTM, just small suggestion

PR #1254, opened 2022-04-24
add a new style for cycle routing

janbar:
It is based from the style winter-sports. Lot of colors have been customized and finally the cycle routes are highlighted with the ref label in sidecar.


janbar:
A sample.
image

janbar:
After diner.
image

janbar:
Hidden flags (starting with '_') are handled.
image

Karry (approved):
I like it.

PR #1261, opened 2022-05-10
Stylesheet modules

Karry:
Standard, winter-sports and cycle stylesheets are pretty similar, just different colors are used and important types are rendered on lower zoom levels. So, it makes sense to deduplicate common parts. Some details may be different after this refactoring, but the main "feeling" from stylesheets should be the same.


janbar:
+1

Issue #1276, opened 2022-06-17
CMake imported target


vmartinlac:
Hello,

If I try to link against libosmscout, I get

```
-- Configuring done
-- Generating done
-- Build files have been written to: /home/victor/depotoir/aa
ninja: error: 'OSMScout::OSMScout-NOTFOUND', needed by 'some_executable', missing and no known rule to make it
```

My dummy CMakeLists.txt is 

```
cmake_minimum_required(VERSION 3.0)
project(some_project)
find_package(libosmscout REQUIRED)
add_executable(some_executable main.cpp)
target_link_libraries(some_executable PUBLIC  OSMScout::OSMScout)
```

The problem disappears if I add the following line to libosmscoutConfig.cmake:

```
set_property(TARGET OSMScout::OSMScout APPEND PROPERTY IMPORTED_LOCATION  ${LIBOSMSCOUT_OSMScout_LIB_RELEASE})
```
So it seems that the error was coming from the property IMPORTED_LOCATION not being filled.

My configuration is:
- Ubuntu 20.04
- ninja build system

Framstag:
@vmartinlac I have not written some defauls of the cmake support, so some questions:
 I assume that you have build and installed libosmscout before? To which location? Which applications, libraries did get build - ans which not?

find_package should resolve to a postprocessed version of cmake/libosmscoutConfig.cmake.in.  

Can you attach the resulting *.cmake file?

@Karry: Can you help? It would take me some time to get into these scripting parts of cmake. Perhaps we should have a Dockerfile for such test scenario at hand?

vmartinlac:
Here is a Dockerfile to reproduce the bug.

Dockerfile
```
FROM ubuntu:20.04

COPY setup.sh /root/setup.sh
COPY CMakeLists.txt /root/test/
COPY main.cpp /root/test/

RUN bash /root/setup.sh
```

setup.sh
```
set -e
export TZ=Europe/Paris
ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
apt update
apt upgrade --yes
apt install --yes git build-essential cmake libcairo2-dev
cd /root/
git clone https://github.com/Framstag/libosmscout.git
cd libosmscout
mkdir build
cd build
cmake .. -DCMAKE_INSTALL_PREFIX=../install -DCMAKE_BUILD_TYPE=Release
make -j8 install
cd /root/test/
[...]

PR #1279, opened 2022-07-04
setup IMPORTED_LOCATION_<CONFIG> cmake properties

Karry:
it should fix https://github.com/Framstag/libosmscout/issues/1276, tested with simple demo https://github.com/Karry/osmscout-cmake-demo


Framstag:
@Karry: Should I still merge this pull request?

Karry:
yes please. it will improve situation for library clients that are using cmake

PR #1285, opened 2022-08-14
Qt6 support:

Framstag:
* A number of fixes/workarounds to get it build under Qt6
* This does not fix all problems


Karry (approved):
LGTM, but not tested it

Framstag:
The goal would be to get Qt6 based build to compile without creating bugs in the Qt5 version.

PR #1318, opened 2022-09-19
Qt6 fixes

Karry:


Framstag:
This is still draft. Shouldn't we merge it, even if it will not fix all problems?

Karry:
Yes, you can merge. It solves at least something. For faster iteration it would be great to prepare docker image for Qt6 build... I will look at it later.

Framstag:
@Karry Regarding Docker, see https://github.com/Framstag/libosmscout/tree/master/ci/docker/ubuntu_22.04_gcc_meson_qt6 :-)

PR #1319, opened 2022-09-20
Many changes in different modules that affect optimization

GermanAizek:
@Framstag hi, review my PR and feedback me.
Optimization insertion method into containers, working with objects, improvement code generation by compiler, minor C++17 changes, code readability.

### DIFF
--- a/libosmscout-import/src/osmscoutimport/GenRelAreaDat.cpp
+++ b/libosmscout-import/src/osmscoutimport/GenRelAreaDat.cpp
@@ -121,7 +121,7 @@ namespace osmscout {
    */
   void RelAreaDataGenerator::ConsumeSubs(Progress& progress,
                                          const std::list<MultipolygonPart>& rings,
-                                         std::list<MultipolygonPart>& groups,
+                                         std::vector<MultipolygonPart>& groups,
                                          GroupingState& state,
                                          size_t topIndex,
                                          uint8_t id)
@@ -159,7 +159,7 @@ namespace osmscout {
                                         const TypeInfoRef& type,
                                         std::list<MultipolygonPart>& parts)
   {
-    std::list<MultipolygonPart>                 rings;
+    std::vector<MultipolygonPart>               rings;
     bool                                        allArea=true;
 
     std::map<Id, std::list<MultipolygonPart*> > partsByEnd;
@@ -322,7 +322,7 @@ namespace osmscout {
       }
     }
 
-    parts=rings;
+    std::move(rings.begin(), rings.end(), std::back_inserter(parts));
 
     return true;
   }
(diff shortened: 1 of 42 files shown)


libosmscout-map/src/osmscoutmap/oss/Parser.cpp
```diff
@@ -1662,7 +1662,6 @@ void Parser::AREABORDERSYMBOLSTYLE(StyleFilter filter, bool state) {
 
 void Parser::ROUTESTYLE(StyleFilter filter, bool state) {
 		LinePartialStyle style;
-		std::string      slot;
```
Karry, line 1665: I am worry that this file is generated by Coco generator. This change will be overwritten by next update style upate. It seems that support for "slot" of routes is not implemented yet...

libosmscout-import/src/osmscoutimport/GenRelAreaDat.cpp
```diff
@@ -322,7 +322,7 @@ namespace osmscout {
       }
     }
 
-    parts=rings;
+    std::move(rings.begin(), rings.end(), std::back_inserter(parts));
```
Karry, line 325: this changes the semantics, or not? Should not we call `parts.clear()` before adding entries? or possibly `assert(parts.empty())`

libosmscout-import/src/osmscoutimport/GenRelAreaDat.cpp
```diff
@@ -433,7 +433,7 @@ namespace osmscout {
     // Copy back data
     //
 
-    parts=groups;
+    std::move(groups.begin(), groups.end(), std::back_inserter(parts));
```
Karry, line 436: same here. it will not replace current content, just add to the end...

libosmscout/src/osmscout/ost/Parser.cpp
```diff
@@ -412,7 +412,7 @@ void Parser::TYPEKINDS(unsigned char& types) {
 }
 
 void Parser::TAGCONDITION(TagConditionRef& condition) {
-		std::list<TagConditionRef> conditions;
+		std::vector<TagConditionRef> conditions;
```
Karry, line 415: This change is fine in general. And as import definition is parsed just once, usage of vector is better IMHO (but would be great to confirm it by some bechmark). But problem is that this file is generated by this script: https://github.com/Framstag/libosmscout/blob/master/libosmscout-map/parser/generate.sh (manually, after definition change)

Karry (commented):
Thank you! LGTM in general, just few minor comments. Final decision is up to @Framstag

GermanAizek:
@Karry fix it https://github.com/Framstag/libosmscout/pull/1319/commits/00d5a708f8a0b6b6d95bc88e18342a58bf99d16d

Framstag:
Changes look fine to me after feedback and fixes. I'll merge.

Thanks for you patch!

libosmscout-import/src/osmscoutimport/GenRelAreaDat.cpp
Karry: I believe that it should look like this: https://github.com/Framstag/libosmscout/pull/1322

PR #1322, opened 2022-09-25
fix Area generator

Karry:
cc @GermanAizek


Framstag:
I already have a similar fix in #1320, since it broke the import I had to do to check my changes.

Issue #1220, opened 2022-02-24
Cluttered ways


vyskocil:
Way rendering maybe cluttered, especially the one way as you could see in the attached screenshot the way labels and one way arrows often overlaps. Is there a solution to avoid this issue ? 
Maybe the way rendering code should treat the labels and the images in the same process allowing to alternate both along the path ? 
Another simpler solution could be to provide the background color of the way to the label renderer, thus it could use it to draw the text with opaque background and overdraw the one way arrows.

![Cluttered ways](https://user-images.githubusercontent.com/452704/155505079-9a2dc839-c710-4c1d-b572-b73ce97c0284.png)

vyskocil:
Here is the standard OpenStreetMap rendering of the same area : 
 
<img width="752" alt="Screenshot 2022-02-24 at 11 23 46" src="https://user-images.githubusercontent.com/452704/155506107-b3618353-f771-4511-a7e1-2ea836122bf5.png">

Framstag:
This (coordinated rendering of symbols ad labels) should in principle possible, however would need some changes.

Currently labels are handled by DrawWayContourLabel, with positioning done in the LabelLayouter.RegisterContourLabel method, based on parameter passed in MapPainter::DrawWayContourLabel.

For symbols this is triggered in MapPainter::DrawWayDecoration with the actual positioning in MapPainterXXX.DrawContourLabel()

In both cases, start and offset to the next instance just would to be input from the outside (and calculated for both coordinated), instead being calculated separately.

Framstag:
@vyskocil 

The placement for the contour symbols has now been moved (in principle, did not check all backends - only cairo and Qt) from the individual backend code to the MapPainter.
Similar refactoring would have to be done now for the label placement of the way labels.
Finally, this would allow coordinating the placement of labels and contour symbols.

Note, that contour symbols now can be scaled in height to the width of the way.

If some of my refactoring broke the IOSX backend, please contact me. This should be easy to fix, just make sure to evaluate alle the information in SymbolContourData.

vyskocil:
Hi @Framstag ! Thank you very much taking this issue into account ! 
I just tried the up to date code just before the scalable contour symbol which is not merged yet, there is indeed a lot less clutter as in my case there is no more way labels at all... ;-) Maybe there are some parameters to tweak in the stylesheet file (I don't use the one provided with the code) ?

Here is a screenshot of the same place : 

[...]

PR #1324, opened 2022-09-27
Try to fix the contour label visibility problem as described in issue…

Framstag:
… #1220

* Improved spacing values and fix projection calculation.

### DIFF
--- a/libosmscout-map/include/osmscoutmap/LabelLayouter.h
+++ b/libosmscout-map/include/osmscoutmap/LabelLayouter.h
@@ -788,11 +788,10 @@ namespace osmscout {
 
       // We do not to see new label appear and disappear during zoom too often.
       // We want new labels to be "between" existing labels
-      // We thus, only draw labels in an amount of a multiple of 2
+      // We thus only draw labels in an amount of a multiple of 2
       countLabels=(size_t)pow(2.0,floor(log2(double(countLabels))));
 
       double labelSpace=(pathLength-countLabels*label->width)/(countLabels+1);
-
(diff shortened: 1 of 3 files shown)


Karry:
Nice. It is better. But still, some street names visible previously are hidden. Seems that short streets, like https://www.openstreetmap.org/way/22561047#map=19/43.62046/6.97074 are problematic.

Framstag:
@Karry The problem is, that the current (new) implementation only evaluates labelSpaces which is now a correct but still rather high value. This result in the requirement, that the way has to be rather long to get a label shown. It should take labelOffset also into account, which should then be a smaller value.
I'm also not yet satisfied with the overall result, I would like the label count to be a sequence of 1,3,7...labels to have a smoother effect on zooming in and out (the overall number of labels would be 1+2+4+8+16....). However that is currently not yet the implementation.

I'll merge this one in the evening and will take a look for an overall improvement in a later separate pull request.

Karry:
For shorter way, I would rather ignore spacing (break the rules), instead of not showing label at all.

Karry:
I tried to use this modification:
```
countLabels=std::max(size_t(1), size_t(pow(2.0,floor(log2(double(countLabels))))));
```
on line 792, as exception for short ways, but the ["Rue Pasteur" label](https://www.openstreetmap.org/way/22561047#map=19/43.62046/6.97074) was filtered out because of variance test (line 801). It will be hard to find some heuristics for label placement to fill requirements:

 - show at least some label even for shorter ways
 - make it stable to avoid label movements during zooming

PR #1326, opened 2022-09-28
revert fast food and restaurant symbols

Karry:
When some object has both, symbol and icon, icon has precedence in renderers. But some renderers do not support icons, or icons may be missing in installation / configuration. For that reason, it is useful to define both.


Karry:
Issue reported here https://github.com/Framstag/libosmscout/issues/1220 originally

PR #1328, opened 2022-09-28
scale symbol space together with contour symbols

Karry:
When symbol is upscaled for high zoom level, space between symbols is upscaled accordingly.

before: 
Screenshot_20220928_153435

after:
Screenshot_20220928_153556


Karry:
Issue reported here https://github.com/Framstag/libosmscout/issues/1220 originally

Issue #1232, opened 2022-03-28
Use constexpr conditions instead of #ifdef


Karry:
Hi, I believe that using `if constexpr (debugFlag)` has many benefits in comparison of `#ifdef`.

 - it is more readable IMHO
 - static analyzers see the code in condition block, so refactoring is simpler, compilation fails when code is not correct even when condition is false...

![Screenshot_20220328_122131](https://user-images.githubusercontent.com/309458/160378310-9eb404fa-848b-4784-a1fc-d05d22f1d72d.png)

![Screenshot_20220328_122900](https://user-images.githubusercontent.com/309458/160379515-72291740-08a8-49bc-9917-10d9ee8cb441.png)

Sad is that it cannot be used everywhere, for example for code that cannot be compiled because of missing dependency, like `#include` ...

What do you think?

Karry:
It actually is possible to switch some of these flags by cmake: https://github.com/Framstag/libosmscout/blob/master/CMakeLists.txt#L68 But not this one `DRAW_DEBUG`...

PR #1331, opened 2022-09-28
use constexpr flags for extra debugging instead of macros

Karry:
see https://github.com/Framstag/libosmscout/issues/1232


Framstag:
The concept looks good to me. Haven't checked each change, though.

PR #1335, opened 2022-10-01
try to place at least one label on short ways

Karry:
even when we do not keep less space than requested (contourLabelSpace). It is better to show at least some label than use strict spacing.

### DIFF
--- a/libosmscout-map/include/osmscoutmap/LabelLayouter.h
+++ b/libosmscout-map/include/osmscoutmap/LabelLayouter.h
@@ -791,7 +791,7 @@ constexpr bool debugLabelLayouter = false;
       // We do not to see new label appear and disappear during zoom too often.
       // We want new labels to be "between" existing labels
       // We thus only draw labels in an amount of a multiple of 2
-      countLabels=(size_t)pow(2.0,floor(log2(double(countLabels))));
+      countLabels=std::max(size_t(1), size_t(pow(2.0,floor(log2(double(countLabels))))));
 
       double labelSpace=(pathLength-countLabels*label->width)/(countLabels+1);
       double offset=labelSpace;


vyskocil:
Hi @Karry &  @Framstag !

This show some way labels but the overlapping between labels and symbols is back... Isn't  it possible that way labels "steal" slots from the line symbols in this case ?
IMG_2B1F2D300B0A-1

Framstag:
Hi @vyskocil, hi @Karry The overlapping of textual labels and contour symbols has not been fixed, yet. For this, positioning of individual labels and symbols has to be coordinates at one place (or at least, both decision point shave to know of each other).

Before my recent work, textual label positioning was done in the LabelLayouter (where it still is) and contour label positioning was done in the concrete MapPainter backend (MapPainterQt,...) based on general parameter.

With my recent changes, this was moved (for backends that already support it) to the MapPainter base class. So at least closer to the LabelLayouter. But not yet close enough.

On the same time, I changed the positioning of the textual labels in the LabelLayouter to improve visual appearance. 

I'll fix the LabelLayouter algorithm first (I might have some time for this next week, but I'm not yet sure).

Afterwards, we have some options:
* Change the positioning algorithm for symbols to implicitly know the algorithm in the LabelLayouter (possible, but somehow fragile).
* Extract the positioning algorithm from LabelLayouter and move it to MapPainter, too. After that, join both algorithms. I'm not sure, if this would work out.
* Move the symbol placement code from MapPainter to LabelLayouter, too. Possible, but will result in even more complex code in the LabelLayouter (=> possibly bad architecture). LabelLayouter AFAIK is already detected as "complex code" by static code analysis :-/ (though other parts of MapPainter, too, because of that I already did some refactoring in the past, to keep the code managable).
* Do some other refactoring that separates layouting and rendering even further (LabelLayouter does layouting and parts of text rendering) to have layouting at one place and rendering separate without getting a big ball of mud.

@Karry : Do you see other options?

@vyskocil : Sorry for the slow progress (and my bugs), but it shows that the code is already at its limit at some places, and we IMHO need to do cleanup first and should not hack in more features blindly.

I'm interested in improving the visual appearance of the map, so I'm interested in this improvement - but lack time...The recent boost of changes was mainly based on my participation at a recent Hacking Weekend (see https://www.fossgis.de/wiki/FOSSGIS_OSM_Communitytreffen_2022_Nummer_18) and some later work. I'm known at the meetings for "the one highly concentrated with the headphones on all day" ;-)

Note that I did show libosmscout to some people, though - and they are still very impressed with what I show (especially the speed on the mobile).

We need more marketing and more apps in app stores - but I'm not the marketing (or the app) guy. Help!

vyskocil:
@Framstag You don't have to be sorry, It's always great when the lib evolve even if it's slow and sometimes with bugs !

PR #1336, opened 2022-10-02
First version of new placement code

Framstag:
Needs further refactoring and cleanup.
Should already show the final effect.
It seems like it triggers effects in the collision detection (needs further investigation)


libosmscout-map/include/osmscoutmap/LabelLayouter.h
```diff
@@ -786,18 +788,54 @@ constexpr bool debugLabelLayouter = false;
       std::vector<Glyph<NativeGlyph>> glyphs = label->ToGlyphs();
 
       double pathLength=labelPath.GetLength();
-      size_t countLabels=(pathLength-labelData.contourLabelSpace)/(label->width+labelData.contourLabelSpace);
+      double minimalSpace=2*labelData.contourLabelOffset;
+      size_t workingLabelCount  =0;
```
Karry, line 792: How many iterations do we need for long ways, like ferry lines? It is good idea to start just with one label?

Karry (approved):
LGTM overall, I am just worry about performance with long ways.

libosmscout-map/include/osmscoutmap/LabelLayouter.h
Framstag: Good question, since I haven't tried yet. The *increment* increases with *2, so the iteration get to huge values fast! Do you have an example way that worries you?

Framstag:
I pushed a refactored and bugfixed version. There is still a problem though. While the offset of the labels as calculates are stable, rendered labels itself visually still move during zoom in/out even if the number of labels per way does not change.

It seems like there is a mismatch in transformation. The label positioning code assume that labels are rendered left aligned, while it looks like the rendering engine (at least for Qt) does something different. 

@Karry  any idea?

Karry:
I would suspect that `LabelPath` class may be the problem. It do not use every coordinate from original graphical path, but replace them with sparse network for better performance (it is dense enough for labels). You may try to change segment legnth in `explicit LabelPath(double minSegmentLength=5);` constructor.

Do you have some simple reproducer?

Karry:
When segment lenght is the problem, we may tune it depending on zoom level, to make label path computations more precise on lower zooms.

Framstag:
Reproducer

image

Zooming in...

image

(the "moving labels" effect is better seen, when slightly zooming via mouse wheel or touchpad)

Location:
51.572529019922804 7.464607318381902 (map center 51.572188169100485 7.4646401742454165)

Export: https://download.geofabrik.de/europe/germany/nordrhein-westfalen/arnsberg-regbez.html

I'm expecting the labels to be fixed in relation to the way on the ground (besides horizontal alignment of the label itself, which likely has to be fixed, too). If you zoom in/out without creating new label instances (number of labels stays the same), I still see the labels moving though.

My next step likely would be to draw lines or rectangles instead of text to make sure my expectation and the actual algorithm match and to rule out text rendering effects. Either my expectations are wrong or something broke...

The overall goal of the placement algorithm is, to do not make labels jump because of new labels appearing/disappearing on zoom in/out (another effect).

Framstag:
IMHO the problem with the first (and last) label is, that they have a hard pixel offset (via contourLabelOffset), so in this case the pixel offset to the start/end of the way should always be the same. This of course will have visual effects on zooming, since the distance on the ground changes in turn, so the labels moves relative to objects on the ground. Perhaps this is already the problem? Left over space (after handling contourLabelOffset and contourLabelSpace) must be delegated to all spaces?

Karry:
> MHO the problem with the first (and last) label is, that they have a hard pixel offset (via contourLabelOffset)...

It seems like that

Screenshot_20221009_100751
Screenshot_20221009_100810

Karry:
What about to adjust contourLabelOffset to nearest zoom level? Lets say that offset will be 100 pixels, when zoom level will be 16.5, it will be adjusted to ~ 150...

Framstag:
@Karry 
I now found a solution that works in spirit of the original idea. The question is, though, is this a "good" effect in the end?

libosmscout-map/include/osmscoutmap/LabelLayouter.h
```diff
@@ -210,11 +210,13 @@ constexpr bool debugLabelLayouter = false;
...
-    std::string text;
+    std::string                     text;     //!< The original label (if debug))
+    double                          offset;   //!< The offset of the label in relation the the way start (if debug))
+    Vertex2D                        start;    //!< Screen coordinates of the start of the path
 #endif
-    size_t priority;
-    std::vector<Glyph<NativeGlyph>> glyphs;
-    osmscout::PathTextStyleRef style;    //!< Style for drawing
+    size_t                          priority; //!< Priority of the label
+    std::vector<Glyph<NativeGlyph>> glyphs;   //!< Vector of glyphd
```
Karry, line 218: typo `s/glyphd/glyps/`

libosmscout-map/include/osmscoutmap/LabelLayouter.h
Karry: you are right, it should be fast and by manual testing, I do not see any impact.

Karry:
It looks really better ;-)

Issue #1341, opened 2022-10-21
Bug in WaterIndexGenerator (import step 19)


Karry:
Some recent Geofabrik extracts are causing crash in water index generator. For example [Albania OSM extract](http://download.geofabrik.de/europe/albania.html) from 19.10.2022. It seems that `WaterIndexProcessor::MergeCoastlines` generates empty coast lines in some conditions. With the fix, we should also add some unittests to `WaterIndexProcessor`, these methods are pretty complex.

TODO: 

- [x] fix actual OSM imports
- [x] document WaterIndexProcessor better
- [x] add unittests to WaterIndexProcessor

Karry:
Interesting, import build from 0a0abd7ab9fe3c483a2e7a231511326a7668516b is successful. So this bug is regression.

Karry:
Git bisect show me that 154ce1c8ddc02db3fae2cbe47e8f7daea17f892a is the first bad commit
```
git bisect log 
# bad: [0ae9a1a5140815c52d04c80657fba0c3772ddad5] Merge pull request #1337 from Karry/scaled-symbols
# good: [0a0abd7ab9fe3c483a2e7a231511326a7668516b] Merge pull request #1265 from Karry/waterway-dam-style
git bisect start 'master' '0a0abd7ab'
# good: [019ef0626943c60a52182930feead439e8d66a3c] Merge pull request #1300 from Framstag/minimal_qt6_build
git bisect good 019ef0626943c60a52182930feead439e8d66a3c
# bad: [22227511cea70fcc0550ab2ce610c31989384340] Merge pull request #1319 from GermanAizek/optimize
git bisect bad 22227511cea70fcc0550ab2ce610c31989384340
# good: [d524ded04c4e8075a2a27282a48d348333240727] Merge pull request #1310 from Framstag/further_symbol_cleanup
git bisect good d524ded04c4e8075a2a27282a48d348333240727
# good: [f4fb611621e863391aed479cf523afe93bf47eda] Merge pull request #1316 from Framstag/fix_warning
git bisect good f4fb611621e863391aed479cf523afe93bf47eda
# good: [b934071a77baa39866382bdd5cc1767a9a8c8015] Merge pull request #1317 from Karry/qt-basemap-fix
git bisect good b934071a77baa39866382bdd5cc1767a9a8c8015
# good: [4f4257cd6b4ac1157ca649c15ff2f8f87cadb10b] Using 'std::move' if object is no longer in use
[...]

PR #1342, opened 2022-10-21
fix WaterIndexProcessor

Karry:


Karry:
fix for: https://github.com/Framstag/libosmscout/issues/1341

PR #1350, opened 2022-11-06
fix placement of contour symbols for short ways

Karry:
When way length is shorter than required symbolSpace, computed label space is negative. It leads to infinite loop during rendering. We should skip contour symbol rendering in such cases.

It is strange that infinite loop happens just with clang, not gcc. But I was not looking for the reason. Negative symbol space should not happen at all.

cc @janbar , it should fix your issue https://github.com/Framstag/libosmscout/commit/ec7528d254393631446e735f5aacaca058fb5909#r88899667


Karry:
Well, this problem is affecting even gcc build. Just zoom enough to Prague location N 50°05.726' E 14°25.587'

janbar:
@Karry, The fix works. Thanks.

PR #1373, opened 2023-01-13
limit number of emited signals on map view change

Karry:
Every view property (lat, lon, angle, zoom level, pixel size)
has its own signal now and it is emited just when this property
is changed more than threshold.

It improves performance of QML UI during map animations.

For example it is not necessary to evaluate rotation of compass 
icon when map is moved without changing angle.


Framstag:
@Karry : Have not seen this...please trigger me, if I forget to merge :-/

Karry:
No problem :-) This one was just minor optimization.

PR #1379, opened 2023-02-25
Increased minimum meson version to check against all build platforms …

Framstag:
…for compatibility

### DIFF
--- a/.github/workflows/build_and test_on_ubuntu_20_04.yml
+++ b/.github/workflows/build_and test_on_ubuntu_20_04.yml
@@ -17,8 +17,10 @@ jobs:
         uses: actions/checkout@v3
       - name: Install software-properties-common
         run:  sudo apt-get install software-properties-common
-      - name: Install gcc compiler, meson and ninja
-        run:  sudo apt-get -y install gcc-10 g++-10 libtbb-dev ccache libtool pkg-config meson ninja-build
+      - name: Install gcc compiler and Co
+        run:  sudo apt-get -y install gcc-10 g++-10 libtbb-dev ccache libtool pkg-config
+      - name: Install meson and ninja via pip
+        run:  "pip install meson ninja"
       - name: Install libosmscout dependencies
         run:  "sudo apt-get update && sudo apt-get install -y
                libxml2-dev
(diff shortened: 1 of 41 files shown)


Framstag:
@Karry 

(pull request not planned to merged this way)

This pull request I did to just check, where we are regarding meson versions. IMHO this signals the following strategies:
* Always install current meson version via pip - not using distribution packages (at least where distribution package is too old)
* Take a look for possible third-party repositories to get a current meson package
* Try to keep compatibility to older (very old) meson versions - as you did.

Your opinion?

Karry:
Hi. I would go with installing Meson via pip on older distros. Newer Linux distribution (for example Ubuntu 22.10) contains new Meson anyway...

eli-schwartz:
Which functionality from newer versions of Meson is needed?

Framstag:
> Which functionality from newer versions of Meson is needed?

I have to take a look, but we recently made use of the subdir attribute for install_headers, which bumped us to 0.63.0 - which we workarounded by using conditional in the script.

Why do you ask? This pull request is just used, to see what version are available under the given build environments and which would fail if we go to 0.63.

I did though use now pip install meson in the sonar build and had no problems. I'm still thinking of requesting a certain version (=> stable builds but needs manual changes for newer meson versions)  vs. current version (which will change which each release and may make break builds, though we did not change anything).

eli-schwartz:
> I have to take a look, but we recently made use of the subdir attribute for install_headers, which bumped us to 0.63.0 - which we workarounded by using conditional in the script.

The conditional seems suboptimal. Given a suitable version of meson (has preserve_path) you install `<includedir>/osmscoutmap/*.h`, but the fallback code installs `<includedir>/*.h`.

This implies that bumping the minimum version and dropping the backwards-compat code is a good idea, if the backwards-compat code isn't being thoroughly tested?

Framstag:
> > I have to take a look, but we recently made use of the subdir attribute for install_headers, which bumped us to 0.63.0 - which we workarounded by using conditional in the script.
> 
> The conditional seems suboptimal. Given a suitable version of meson (has preserve_path) you install `<includedir>/osmscoutmap/*.h`, but the fallback code installs `<includedir>/*.h`.
> 
> This implies that bumping the minimum version and dropping the backwards-compat code is a good idea, if the backwards-compat code isn't being thoroughly tested?

(I have to correct: We used preserve_path, not subdir and preserve_path is 0.630+). Yes, the conditional would not do the same with older meson versions, that is why I went forward to make sure that we can drop the conditional and enforce a version that will work.

We do not have a test for install. That would be something interesting to have in our automatic builds for cmake and meson. I'll make an issue.

PR #1386, opened 2023-03-04
improve GeoCoord string parsing

Karry:
We should be benevolent in format that we are able to parse. As `GeoCoord::Parse` is used in Qt's search model, it is handy to allow even coordinate format used on booking.com, like `N 046° 40.425, E 05° 31.362`.


Framstag:
See https://sonarcloud.io/project/issues?id=Framstag_libosmscout&pullRequest=1386&resolved=false&types=CODE_SMELL for some Sonar findings. I'll disable warnings regarding [[nodiscard]] and use of parenthesis but the warning regarding missing constants makes sense.

PR #1423, opened 2023-05-01
Splitted TypeFeatures.cpp/.h into multiple files

Framstag:
Split TypeFeatures.h into multiple files. This might increase overall compile time but I hope that it reduces compile time for strategic files like MapPainter. It also reduces stress for IDE which do not handle big files.


Karry (approved):
LGTM, thank you for split of features, that source file was too huge

PR #1444, opened 2023-06-10
Asynchronous work

Karry:
This is initial preparation for moving some logic from libosmscout-client-qt to libosmscout-client. As `std::future` doesn't have callbacks, I need some custom future to do asynchronous work without blocking UI thread.


Framstag:
@Karry :I would have merged but GitHub says there are merge conflicts. Can you take a look?

Karry:
Resolved. There was just small issue in CMake and Meson files.

Karry:
Can we merge it @Framstag ?

PR #1458, opened 2023-06-23
supports qt5 and qt6 in cmake (draft)

playgithub:
As a draft, just add some files without changing any files.
If it's ok, remove qt5 related code in cmake and `include(feature-qt)` instead.


playgithub:
be a draft

PR #1459, opened 2023-06-23
supports qt5 and qt6 in cmake (draft)

playgithub:
As a draft, just add some files without changing any files.
If it's ok, remove qt5 related code in cmake and `include(feature-qt)` instead.


Framstag:
@Karry Makes sense that you have a look at it. Should suite you, too

playgithub:
cmake configure passed
build not tested, because I'm totally new to the lib
Any help?

Framstag:
I have approved execution of the builds...let's see what happens :-)

Framstag:
The Ubuntu 20.04 gcc cmake builds does not seem to build the qt backend anymore though qt5 was installed. Same for the OS X build.

playgithub:
The options below are `OFF` by default now, so they are not in test by default.

```cmake
option(OSMSCOUT_BUILD_MAP_QT "Enable build of Qt map drawing backend" OFF)
option(OSMSCOUT_BUILD_CLIENT_QT "Find newest Qt in respect to QT_VERSION_MAJOR and QT_VERSION_MINOR" OFF)
option(OSMSCOUT_BUILD_TOOL_OSMSCOUT2 "Enable build of OSMSCout2 demo" OFF)
option(OSMSCOUT_BUILD_TOOL_STYLEEDITOR "Enable build of StyleEditor application" OFF)
```

Framstag:
I would expect the following behavior (similar to the meson build):
* If Qt is not installed, anything depending on Qt is dropped from the build
* If only Qt5 is installed, it is automatically detected and all Qt dependencies are build against Qt5
* The same for Qt6
* If Qt5 and Qt6 are installed, the current default is to make use of Qt5 (because its resulting in working binaries). This though should be easily changeable in the future.
* If Qt 5 and Qt6 are installed, the user should be able to explicitly decide which is used.
* If Qt support is explicitly switched off no Qt dependent binary should be build (question is, if Qt should still be searched)

As far as I remember, the meson build does make shortcuts. If Qt5 is the default, it will not look for Qt6. That ca likely be changed to align both build system behaviors (the Qt6 meson support was a quick and dirty hack, to see how compatible Qt and Qt5 are).

playgithub:
> I would expect the following behavior (similar to the meson build):
> 
> * If Qt is not installed, anything depending on Qt is dropped from the build
> * If only Qt5 is installed, it is automatically detected and all Qt dependencies are build against Qt5
> * The same for Qt6
> * If Qt5 and Qt6 are installed, the current default is to make use of Qt5 (because its resulting in working binaries). This though should be easily changeable in the future.
> * If Qt 5 and Qt6 are installed, the user should be able to explicitly decide which is used.
> * If Qt support is explicitly switched off no Qt dependent binary should be build (question is, if Qt should still be searched)

OK, a function will set the flags as expected, I'll do it soon.

> Hmm, something is broken now.

I'll debug it.

playgithub:
> I would expect the following behavior (similar to the meson build):
> 
> * If Qt is not installed, anything depending on Qt is dropped from the build
> * If only Qt5 is installed, it is automatically detected and all Qt dependencies are build against Qt5
> * The same for Qt6
> * If Qt5 and Qt6 are installed, the current default is to make use of Qt5 (because its resulting in working binaries). This though should be easily changeable in the future.
> * If Qt 5 and Qt6 are installed, the user should be able to explicitly decide which is used.
> * If Qt support is explicitly switched off no Qt dependent binary should be build (question is, if Qt should still be searched)

The rules are convenient, but if it is complicated as a policy.
According to unix style, what about to keep it simple by removing policies and leaving only mechanisms?

playgithub:
Trying to support multiple ways to enable qt:
* Policy as above
* Hard coded

e.g.
```
if(ENABLE_QT_WITH_BUILT_IN_RULES)
    enable_qt_by_rules()
else()
    if (OSMSCOUT_BUILD_MAP_QT OR OSMSCOUT_BUILD_CLIENT_QT)
        if(NOT ${QT_MAJOR_VERSION} OR ${QT_MAJOR_VERSION} STREQUAL "")
            set(QT_MAJOR_VERSION "6" CACHE STRING "" FORCE)
            set(QT_MINOR_VERSION "" CACHE STRING "" FORCE)
        endif()

        if(${QT_MAJOR_VERSION} STREQUAL "6")
            include(feature-qt6)
        elseif(${QT_MAJOR_VERSION} STREQUAL "5")
            include(feature-qt5)
        endif()
    endif()
endif()
```

Framstag:
Just a short note: From history: There are people intereste din libosmsocut, but the first major hurdle ist "getting it to build". Especially Windows people struggle a lot. That is the reason, why we have few dependencies, most dependencies are optional and we have a build for everyone. That is also the reason, that the build should make the most and the best out "not passing any arguments to the build". It is also one of the reasons, why we have so many GitHub builds. I can also point to: It works, see there, make it the same way...

Possibly we should offer Binaries for Windows to further support people.

Framstag:
And yes, I appreciate any new helpful sould interested in making libosmscout better :-). Thank you!

Framstag:
> The rules are convenient, but if it is complicated as a policy. According to unix style, what about to keep it simple by removing policies and leaving only mechanisms?

That I do not understand. What is the difference in this case between policy and mechanismus? If there is a technical problem with handling Qt5 and Qt6 just tell. It is possible, that we might have similar problems in meson.

playgithub:
> That I do not understand. What is the difference in this case between policy and mechanismus?

In this case, using qt5 is a mechanism, using qt6 is another mechanism.
Weather to use qt is a policy,  how to automatically choosing one version of qt to use is another policy.

playgithub:
It failed to find `Qt6` in `OSMScout2/CMakeLists.txt`:
```cmake
find_package(Qt6 REQUIRED COMPONENTS LinguistTools)
```
which is the offical way to find Qt6: https://doc.qt.io/qt-6/qtlinguist-cmake-qt-create-translation.html

BTW `qt_create_translation` has been deprecated, will be replaced by `qt_add_translations` later.

playgithub:
Debug heavily depends on build & run, what about making all the workflows be automatically triggered by push?

Framstag:
@playgithub : Since you are touching vcpkg: It makes sense to add Qt6 to vcpkg (or offer alternative files), not however that - at least in the current setup - vcpkg may build all dependencies within the build. AFAIK OS GitHub Actions are time-limited. This results in canlceled builds, if Qt5/6 is rebuilt within the libosmscout build (creating a kind of deadlock situation). vcpkg builds are not rebuilt on every build, because we use GitHub Actions caching (which is only updated on successful builds (by default)).

All that made me drop Qt5 from the vcpkg-based builds.

Framstag:
> Debugging heavily depends on build & run, what about making all the workflows be automatically triggered by push?

That is the default...for developers of the project. Which you are currently not. Disabling automatic builds for non-project members is a security measure. This allows members to check the build for possibly malicious content. I have no problem adding you...if you plan to add further code :-)

playgithub:
> I have no problem adding you...if you plan to add further code :-)

I like osm, I think I will.
Thanks

playgithub:
How can I make the git actions automatically triggered by push in my fork?

playgithub:
> Since you are touching vcpkg: It makes sense to add Qt6 to vcpkg (or offer alternative files)

In my experience, `vcpkg` is very active for this lib, as I post issue to supoort qt, they support it in one day.
So what about to remove vcpkg support here?
Anyway I'm going to let the build pass first, then I'll refine it until it's fine enough.

playgithub:
I've push 2 more commits, but it doesn't appear here.

playgithub:
"Unknown CMake command "qt5_create_translation" for "iOS/cmake",
lack of `qt5-tools` according to https://github.com/Cockatrice/Cockatrice/issues/1548

playgithub:
> `Unknown CMake command "qt5_create_translation"` for "iOS/cmake" and "OS X / cmake", lack of `qt5-tools` according to [Cockatrice/Cockatrice#1548](https://github.com/Cockatrice/Cockatrice/issues/1548)

@Framstag Could you make sure that `qt5-tools` for iOS and OS X has been installed?

Framstag:
> > `Unknown CMake command "qt5_create_translation"` for "iOS/cmake" and "OS X / cmake", lack of `qt5-tools` according to [Cockatrice/Cockatrice#1548](https://github.com/Cockatrice/Cockatrice/issues/1548)
> 
> @Framstag Could you make sure that `qt5-tools` for iOS and OS X has been installed?

Under OS X/ Brew there is only one installation pckage for Qt5. Attsched the complete content of the package as returned by brew ls --verbose qt5

[qt5.txt](https://github.com/Framstag/libosmscout/files/11994521/qt5.txt)


What exactly are you looking for?

playgithub:
> What exactly are you looking for?

Looking for something for `qt5_create_translation` to run,
after checking the output of `ls --verbose qt5`, `Qt5LinguistTools` might make `qt5_create_translation` available, could you install it for testing.

Framstag:
> > What exactly are you looking for?
> 
> Looking for something to make `qt5_create_translation` available. After checking the output of `ls --verbose qt5`, `Qt5LinguistTools` might make `qt5_create_translation` available. Could you try it?

The Qt5 ist installed via https://formulae.brew.sh/formula/qt@5#default. I do not see a separate package for the Linguist Tools. Based of the package is the source here: https://formulae.brew.sh/formula/qt@5#default . As ou can see, it used the open source release package

```
 url "https://download.qt.io/official_releases/qt/5.15/5.15.10/single/qt-everywhere-opensource-src-5.15.10.tar.xz"
```

That is, what is there (and no more).

playgithub:
For `OS X`

```cmake
find_package(Qt5 COMPONENTS Core Gui Widgets Qml Quick Svg Positioning Multimedia LinguistTools QUIET)
```
in `cmake/feature-qt5.cmake` succeeds,
PS: Though no keyboard `REQUIRED`, the debug message shows that all components are found.



```cmake
find_package(Qt5 REQUIRED
		         COMPONENTS Core LinguistTools)
```
in `OSMScout2/CMakeLists.txt` fails, log:
```
CMake Error at OSMScout2/CMakeLists.txt:88 (find_package):
  By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Qt5", but
  CMake did not find one.

  Could not find a package configuration file provided by "Qt5" with any of
  the following names:

    Qt5Config.cmake
    qt5-config.cmake

  Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
  to a directory containing one of the above files.  If "Qt5" provides a
  separate development package or SDK, be sure it has been installed.
```

If the former succeeds, that latter should also succeeds, why not?

playgithub:
After some tests on OSX, the code below finds Qt5.

```cmake
find_package(Qt5 REQUIRED
                         COMPONENTS Core LinguistTools
                         PATHS "/usr/local/Cellar/qt@5/5.15.10/")
```

* `PATHS "/usr/local/Cellar/qt@5/5.15.10/"` must be there
* `PATHS "/usr/local/Cellar/qt@5/"` not work

For Qt5 version independent, searching path for Qt5 should be set before cmake configure, i.e. it should be set by the tester.

playgithub:
@Framstag

### Problem 1

`CMAKE_PREFIX_PATH` should be set for searching Qt5

e.g. the code below succeeds on OS X, without `PATHS` it would fail, but it's Qt5 version dependent, `CMAKE_PREFIX_PATH` should be set properly instead.
```cmake
find_package(Qt5 REQUIRED
             COMPONENTS Core LinguistTools
             PATHS "/usr/local/Cellar/qt@5/5.15.10/")
```
#### logs
* Visual Studio 2019 / cmake
* OS X / cmake
* Sanitize/clang
* Sanitize/gcc
    ```
    Error:   By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
    -- Configuring incomplete, errors occurred!
    asked CMake to find a package configuration file provided by "Qt5", but
    CMake did not find one.

    Could not find a package configuration file provided by "Qt5" with any of
    the following names:

        Qt5Config.cmake
        qt5-config.cmake

    Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
    to a directory containing one of the above files.  If "Qt5" provides a
    separate development package or SDK, be sure it has been installed.
    Call Stack (most recent call first):
    cmake/feature-qt5.cmake:14 (find_package)
    cmake/feature-qt.cmake:38 (include)
    CMakeLists.txt:327 (include)
    ```

### Problem 2

* iOS/cmake
    ```
    CMake Error at /usr/local/Cellar/qt@5/5.15.10/lib/cmake/Qt5Gui/Qt5GuiConfigExtras.cmake:9 (message):
    Failed to find "gl.h" in
    "/System/Library/Frameworks/OpenGL.framework/Headers;/System/Library/Frameworks/AGL.framework/Headers".
    Call Stack (most recent call first):
    /usr/local/Cellar/qt@5/5.15.10/lib/cmake/Qt5Gui/Qt5GuiConfig.cmake:227 (include)
    /usr/local/Cellar/qt@5/5.15.10/lib/cmake/Qt5/Qt5Config.cmake:28 (find_package)
    cmake/feature-qt5.cmake:8 (find_package)
    cmake/feature-qt.cmake:38 (include)
    CMakeLists.txt:327 (include)
    ```

I have no experience on apple, hints found: https://itecnote.com/tecnote/macos-cmake-not-finding-gl-h-on-os-x/

playgithub:
need help for paths to find qt

@Framstag

Framstag:
On my Macbook Pro I get:

```
-- libosmscout 1.1.1 build configuration:
-- Shared libraries:                ON
-- Build frameworks:
-- Qt libraries:
--  - Core:                         1
--  - Gui:                          1
--  - Svg:                          1
--  - Widgets:                      1
--  - Qml:                          1
--  - Quick:                        1
--  - Multimedia:                   1
-- core library:                    ON
--  - marisa support:               TRUE
-- import library:                  ON
--  - libxml2 support:              TRUE
--  - protobuf support:             TRUE
--  - C++17 parallel execution:     0
-- gpx library:                     ON
--  - libxml2 support:              TRUE
-- map drawing interface:           ON
--  - Agg map drawing backend:      OFF
--  - Cairo map drawing backend:    ON
--  - OpenGL map drawing backend:   ON
--  - DirectX map drawing backend:  OFF
--  - Qt map drawing backend:       ON
--  - SVG map drawing backend:      ON
--  - OS X/iOS map drawing backend: ON
--  - GDI+ map drawing backend:     OFF
-- client libraries:
--  - Qt client library:            ON
-- tools:
--  - Import/Import:                ON
--  - Import/BasemapImport:         ON
--  - DumpData:                     ON
--  - OSMScout2:                    ON
--  - OSMScoutOpenGL:               ON
--  - StyleEditor:                  ON
-- tests:                           ON
-- demos:                           ON
--  - heap profiler (Gperftools)    OFF
-- bindings:
--  - Java binding:
--  - C# binding:
-- external interfaces:
--  - MATLAB mex function:          OFF
-- doxygen:                         NO
-- documentation:                   OFF
-- webpage:                         ON
-- Using ccache:                    /usr/local/bin/ccache
-- Configuring done (27.6s)
```

This likely works becaus of one of my folling environement settings:

PATH contains `/usr/local/opt/qt@5/bin`

```
PKG_CONFIG_PATH=/usr/local/opt/qt/lib/pkgconfig
LDFLAGS=-L/usr/local/opt/qt@5/lib
CPPFLAGS=-I/usr/local/opt/qt@5/include
```

This are the recommendation on command line after Qt5 installation via brew. I assume that the PATH one is the relevant one (pkg-config --list-all does not return a qt dependency) , calling qtpaths or similar.

Looks like the pkg-config path on my system is wrong and should be `export PKG_CONFIG_PATH=/usr/local/Cellar/qt@5/5.15.10/lib/pkgconfig`

playgithub:
> Problem 1:
> This is strange, since I assume that vcpkg should handle this all for you. I'm sure I did get a local build of cmake+vcpkg including Qt on my local computer - with a Windows cmake installation (which the VisualStudio19 image from GitHub should already have installed).

Does `libosmscout` depend on `vcpkg`?
If yes, is it easy to make `libosmscout` not depend on `vcpkg`? And if `vcpkg` wants `libosmscout`, it should be "`vcpkg` depends on `libosmscout`", right?

Framstag:
> Does `libosmscout` depend on `vcpkg`? If yes, is it easy to make `libosmscout` not depend on `vcpkg`? And if `vcpkg` wants `libosmscout`, it should be "`vcpkg` depends on `libosmscout`", right?

No. libosmscout support the use of vcpkg via cmake as a supplier for the requested dependencies.
 
> BTW if `libosmscout` depends on some libs that's in `vcpkg`, it can integrate `vcpkg` in cmake with
> 
> ```
> cmake -DCMAKE_TOOLCHAIN_FILE=/path/to/vcpkg/scripts/buildsystems/vcpkg.cmake ...
> ```

That is exactly what is happening in the Visual Studio and cmake based build, see:

https://github.com/Framstag/libosmscout/blob/master/.github/workflows/build_and%20test_on_vs2019.yml

So we request vcpkg via cmake and via vcpkg.json to offer use the dependencies and configure cmake to find them.

playgithub:
> So we request vcpkg via cmake and via vcpkg.json to offer use the dependencies and configure cmake to find them.

`vcpkg` has its own `vcpkg.json` for `libosmscout` as `/path/to/vcpkg/ports/libosmium/vcpkg.json`.
Why don't use `vcpkg install zlib libiconv libxml2 ...` to install dependencies, it'll be more clear, right?

playgithub:
> libosmium is another library than libosmscout. See https://github.com/osmcode/libosmium

Sorry for typo, `/path/to/vcpkg/ports/libosmscout/vcpkg.json` also there.

playgithub:
I've found the libs below also meet my demands, but less coupled:
* libosmium
* osrm
* mapnik

I felt that `cmake` is not very hot here, anyway I'll try to finish this PR if it helps.

Framstag:
We have found another solution. Nevertheless thanks for your effort. I'll close this pull request now.

PR #1466, opened 2023-07-01
Make client setting Qt independent

Karry:
This is first merge requests that moves some classes from (libosmscout) client-qt to client library, to provide high-level asynchronous apis to users that don't want, or just cannot use Qt. 

This request is huge and the settings api is now a bit more complicated, because there have to be conversion between osmscout signals and std data types to Qt ones - sorry for that. But I believe that result will be usable for more users...

There is one issue that is not solved yet - standard C++ doesn't support reading Qt resources (compiled in to binary), so all resources (of OSMScout2, other demos and StyleEditor) have to be installed as files... 

@janbar , as you are another user of Qt / QML apis, may I ask you for testing?


Framstag:
Looks OK for me. Wait for @janbar or should I merge?

Framstag:
> Looks OK for me. Wait for @janbar or should I merge?

@Karry ?

Karry:
Yes, you may merge it. Thank you for review.

PR #1467, opened 2023-07-04
Buildsystem improvements

syyyr:
Hi, I'm trying to use libosmscout in my CMake Qt6 project, but libosmscout's CMake doesn't support Qt6 yet.  Meson does support Qt6, but there isn't any meson->cmake integration, so I need libosmscout to generate pkgconfig files.

Some of the generated headers also needed to be installed, so I've changed that also.


Karry:
Hi. Don't look to changes yet. But in general, there should not be need for pkgconfig, there is `libosmscoutConfig.cmake` that should be enough to include library...

Karry (approved):

syyyr:
Thanks for merging.

PR #1468, opened 2023-07-11
thread finalizer utility

Karry:
Thread finalizer provides signal on thread exit.
It is not guaranteed that signal will be emitted for all threads, some threads may be still running on program exit. For example detached thread, global scheduler threads, when std::exit is called...


Karry:
replaced by https://github.com/Framstag/libosmscout/pull/1471

PR #1469, opened 2023-07-11
add nlohmann/json to libosmscout-client library

Karry:
- it is header only json library
 - licensed under terms of MIT license (compatible with LGPL)
 - require C++11

This library may be used as replacement for Qt json api. It will be possible to make some parts of client-qt library independent on Qt and provide it to other platforms...


Framstag:
@Karry: Can we mix the licenses?

Framstag:
OK, I should read what you have written :-(

PR #1488, opened 2023-08-21
use jvm 17 for Sonar scanner

Karry:
Should solve this warning: 

Screenshot_20230821_234421


Karry:
it seems that it is not working :-/

PR #1489, opened 2023-08-22
jvm version test

Karry:


Karry:
replaced by: https://github.com/Framstag/libosmscout/pull/1490

PR #1491, opened 2023-08-26
Add support for Qt 6 or Qt 5 builds using cmake

Framstag:


Framstag:
@Karry: This seems to work in principle, however I'm using the versionless support starting with 5.15. Currently, we support >=5.6 ad Ubuntu seem to come with 5.12.

If we want to support older version, we have to explicitly duplicate all Qt code to use Qt5 or at6 functions in cmake. Or we have to write wrapper functions ourself.

Karry (approved):
LGTM

Karry:
Your solution looks straightforward, I would avoid cmake duplication for Qt5 and Qt6. I am using slightly modified CMake build in my "Sailfish OS" branch, let me check how it will work when I merge your branch there...

Karry:
So, I am including individual libraries to my build, not "root" CMakeList.txt, patch for Qt 5.6 compatibility is pretty tiny. You may merge it in current form, I will deal with that for my project. 

Have simple Qt6 support in CMake is more beneficial than supporting Qt 5.6 in upstream repository...

PR #1494, opened 2023-08-30
cmake: Build improvements

syyyr:


syyyr:
Okay, I think I finally got it, sorry for spamming the CI with so many builds.

Framstag (approved):

Karry (approved):
doesn't test it, but looks correct

PR #1500, opened 2023-09-14
cmake-config: Fall back to other build type

syyyr:
It is impossible to link with a system-installed libosmcout if your application uses a different type than the system. This change makes it so that the config module falls back to the other build types.


syyyr:
I have updated the PR so that even the components fall back to the other build types. It's not pretty, but I don't how to make it better.

Also, a question: is it necessary to include the "d" suffix to the libraries? I know that on Windows, you can tell CMake to build both version, but on Linux you can't, and so I have to make patches like these. Now, that I think of it, the PkgConfig file generation code is also broken, because it won't work with the suffixed library names (because of the -l argument).

Framstag:
> Also, a question: is it necessary to include the "d" suffix to the libraries? I know that on Windows, you can tell CMake to build both version, but on Linux you can't, and so I have to make patches like these. Now, that I think of it, the PkgConfig file generation code is also broken, because it won't work with the suffixed library names (because of the -l argument).

AFAIR some platforms generate libraries for debug and release that cannot be mixed. In this case, a prefix or postfix is required. As far as I remember, this is/was the case for Windows. It should not be the case for e.g. Linux.

Using CMake and meson it should always be possible to create different build directories to build different versions of the library. However, if you *install* them under Windows it should be possible to explicitly select the release or the debug version (and there are other compiler options that could make libraries incompatible), since you cannot mix. A OSMScout2 debug binary must use debug libraries (under Linux you just loose debugging information for parts).

Issue #1503, opened 2023-09-21
Way missing in MapService query


frank-aurich:
Hi,

while experimenting with libosmscout I discovered that some OSM ways are missing from the MapData.

I used this OSM dataset: http://download.geofabrik.de/europe/germany/sachsen-latest.osm.pbf
and converted it with the `Import` tool using the default settings (i.e. the `maps.ost` from the repository).

The way missing from the result is https://www.openstreetmap.org/way/25325001.

Simplified code example:
```
const osmscout::Magnification magnification(osmscout::MagnificationLevel(14));
const osmscout::TileKey tileId(magnification, osmscout::TileId(8814, 12839));
osmscout::MapService::TypeDefinition types;
m_styleConfig->GetWayTypesWithMaxMag(magnification, types.wayTypes);

auto data = std::make_shared<osmscout::MapData>();
auto tile = m_mapService->LookupTile(tileId);
m_mapService->LoadMissingTileData(searchParameter, magnification, types, tiles);
m_mapService->AddTileDataToMapData(tiles, *data);
[...]
```
Another example:
https://www.openstreetmap.org/way/4851357 (`osmscout::TileId(8819, 12838)`)
This way also is excluded from the `MapService` results.

Is this by intention? I would expect a `Way` there, and both ways do exist in the resulting `osmscout` database (I checked with the `DumpData` tool).

Karry:
Looking to the OSM data, these ways are closed, but not marked as a roundabout. So, I would guess that osmscout evaluate them as a area, because `highway_residential` type have no `PIN_WAY` flag...

Framstag:
Patch merged. We should be carefully with setting PIN_WAY, since some types can be represented by area or way, But not highway residential:

![image](https://github.com/Framstag/libosmscout/assets/14061582/58540547-10d5-4ef1-ab77-44a5da6fe69f)

Framstag:
@frank-aurich : Can we close the issue?

frank-aurich:
Yes, I can confirm the fix works as expected

PR #1505, opened 2023-09-21
add PIN_WAY flag to highway_residential type

Karry:
it should solve https://github.com/Framstag/libosmscout/issues/1503

Can you try that @frank-aurich  please?


frank-aurich (approved):
I can confirm this change fixes #1503.

PR #1517, opened 2023-10-10
ScreenBox resize fix

Karry:
As reported by @ToFuCH at https://github.com/Karry/osmscout-sailfish/issues/325 , rendering Bern (Switzerland) map on zoom level 13 is causing crash. 

Problem is that some area (not important which one) has zero height during rendering and renderer tries to resize it with zero offset. This operation should be correct, but it hits assert
```
assert(offset>-GetHeight());
```

Correct assert should be:
```
assert(offset >= GetHeight()/-2);
```

### DIFF
--- a/Tests/src/ScreenBox.cpp
+++ b/Tests/src/ScreenBox.cpp
@@ -101,3 +101,15 @@ TEST_CASE("Resize empty rectangle by zero")
 
   REQUIRE(expectedResizedBox==resizedBox);
 }
+
+TEST_CASE("Resize square to zero")
+{
+  osmscout::ScreenBox box(osmscout::Vertex2D(-1, -1),
+                          osmscout::Vertex2D(1, 1));
+  osmscout::ScreenBox resizedBox=box.Resize(-1);
+
+  osmscout::ScreenBox expectedResizedBox(osmscout::Vertex2D(0, 0),
+                                         osmscout::Vertex2D(0, 0));
+
+  REQUIRE(expectedResizedBox==resizedBox);
+}
--- a/Tests/src/ScreenBox.cpp
+++ b/Tests/src/ScreenBox.cpp
@@ -77,3 +77,27 @@ TEST_CASE("Resize >0")
 
   REQUIRE(expectedResizedBox==resizedBox);
 }
+
+TEST_CASE("Resize ==0")
+{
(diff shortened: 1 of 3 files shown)


Karry:
When I thinking about ScreenBox api again, it would make me sense to allow higher negative offset without crashing the app. Square size should just reduced to zero... I'm worry that we will hit similar bug again if negative offset is used somewhere...

Karry:
Ok, there is no case in the library when offset may be negative... So my concern is not valid. But still, I would like the api that is accepting all numbers...

Framstag:
Hi @Karry :This is a "good API" question.

I try to summarize my opinion:
- Objects and their APIs define state constraints (APIs are contracts). Defined constrains allow writing valid, correct, robust and simple code. Calling or using code can rely on the constraints and thus can remote checks and tests, which would be required if no constraints would be defined.
- Constraints should be defined as soon as possible to make as much code as possible "simpler"
- This is the reason why I introduced a number of objects like ScreenBox, GeoBox...
- The late introduction is problematic in libosmscout because we have to change code with unclear constraints
- Objects/classes thus should only and always have state that is valid. Trying to set an object into an invalid state should be rejected (boolean return, exception or whatever to signal rejection).
- A resize method thus should only resize to a valid new state
- An object with area <0 is not valid
- An object with area 0 is valid, but there should be some getter like "IsVisible()" or similar to check this special state. This is of course debatable and depending on the requirements one could decide differently. For our "universe" it makes sense to handle it the way that size 0 is still valid.
- the render stack should have the constraint to only render areas which have area >0. Thus we can resize areas to ß, but should not pass them to the render stack (or the stack has to filter them out) since the visual outcome is rather undefined.
- A resize() should either resize to a valid state or reject the resize
- We can however define an additional method like "resizeUpToZeroArea()" which does, what you like to have. Resize to valid state even if the offset is "wrong". In this case, the method and its action are consistent. We should of course try to avoid usage of this method as much as possible ;-)
- We can also add a method "canBeResizeTo():bool" so that one can pre-check if a resize would be possible (and remove the object from the list in the other case). Possibly this already avoids implementation of resizeUpToZeroArea()?

Hope this helps?

Karry:
Thank you for your comments. I just make it simplest for the method user - accept any value and keep object valid...

When negative offset is bigger than size / 2, it is reduced to empty box...

PR #1522, opened 2023-11-09
try to clean up partial download on resume

janbar:
Trying to resume an aborted partial download. The flow is simple as to delete the invalid map directory, but not recursively for security reason. Therefore now the function deleteDatabase handle partial downloads too.

The original issue has been reported by https://github.com/janbar/osmin/issues/63.


Karry (approved):
thank you!

PR #1527, opened 2023-11-15
Check if lookupDir exists before iterating

syyyr:
std::filesystem::recursive_directory_iterator throws if the directory doesn't exist.

### DIFF
--- a/libosmscout-client/src/osmscoutclient/MapManager.cpp
+++ b/libosmscout-client/src/osmscoutclient/MapManager.cpp
@@ -43,6 +43,15 @@ CancelableFuture<bool> MapManager::LookupDatabases()
     std::vector<std::filesystem::path> databaseFsDirectories;
 
     for (const auto &lookupDir:databaseLookupDirs){
+      // Symlinks are automatically resolved when using exists() and is_directory().
+      // https://en.cppreference.com/w/cpp/filesystem/exists
+      // https://en.cppreference.com/w/cpp/filesystem/is_directory
+      // https://en.cppreference.com/w/cpp/filesystem/status
+      if (!std::filesystem::exists(lookupDir) || !std::filesystem::is_directory(lookupDir)) {
+        osmscout::log.Warn() << "Lookup dir" << lookupDir.string() << "doesn't exist or isn't a directory";
+        continue;
+      }
+
       for (const auto & fInfo : std::filesystem::recursive_directory_iterator(lookupDir)) {
         auto entryPath = fInfo.path();
         if (fInfo.is_regular_file() && entryPath.has_filename() && entryPath.has_parent_path() && entryPath.filename() == TypeConfig::FILE_TYPES_DAT){


libosmscout-client/src/osmscoutclient/MapManager.cpp
```diff
@@ -43,6 +43,11 @@ CancelableFuture<bool> MapManager::LookupDatabases()
     std::vector<std::filesystem::path> databaseFsDirectories;
 
     for (const auto &lookupDir:databaseLookupDirs){
+      if (!std::filesystem::exists(lookupDir) || !std::filesystem::is_directory(lookupDir)) {
```
Karry, line 46: and what about symlinks?

libosmscout-client/src/osmscoutclient/MapManager.cpp
```diff
@@ -43,6 +43,11 @@ CancelableFuture<bool> MapManager::LookupDatabases()
...
+        osmscout::log.Warn() << "Lookup dir" << lookupDir << "doesn't exist";
```
Karry, line 47: You need to convert lookupDir to string by `lookupDir.string()` explicitly, because there is just implicit conversion to "native path representation". And it is wstring on MSVC.

Framstag:
@syyyr : Can you fix the code in regard to @Karry comments?

libosmscout-client/src/osmscoutclient/MapManager.cpp
syyyr: Done

libosmscout-client/src/osmscoutclient/MapManager.cpp
syyyr: I have edited the code to resolve symlinks.

syyyr: Although, I have just checked and `fs::exists` does resolve symlinks and a quick test of `fs::is_directory` shows that it does also. Should I remove the symlink-resolving code?

Framstag: If fs:exists officially checks, then yes. It makes sense to reply on official documentation. In this case, see https://en.cppreference.com/w/cpp/filesystem/exists, corresponding behaviour is documented. For is_directory it is not as obvious: https://en.cppreference.com/w/cpp/filesystem/is_directory

syyyr: cppreference is a great resource, and I have checked that the C++ standard agrees just to be sure: `is_directory` should be implemented in terms of `fs::status` and that should resolve symlinks.

Should I remove symlink resolving code then? Maybe put a comment with a link to cppreference, to make things clear?

Framstag: If your code correctly handles symlinks without explicit check, then yes, you can remove the symlink specific code. And yes, a comment in this case IMHO is helpful, to make sure, that we do not get an "improvement" pull request wtihout need.

syyyr: We

syyyr: >If your code correctly handles symlinks without explicit check, then yes, you can remove the symlink specific code

According to `[fs.op.exists]` in the standard, `std::fs::exists` is implemented in terms of `std::fs::status`:
```
bool exists(const path& p, error_code& ec) noexcept;
Let s be a file_status, determined as if by status(p) or status(p, ec), respectively.
Effects: The signature with argument ec calls ec.clear() if status_known(s).
Returns: exists(s).
```
The same for `std::fs::is_directory` in `[fs.op.exists]`:
```
bool is_directory(const path& p);
bool is_directory(const path& p, error_code& ec) noexcept;
Returns: is_directory(status(p)) or is_directory(status(p, ec)), respectively. The signature
with argument ec returns false if an error occurs.
```
And according to `[fs.op.status]`, `std::fs::status` resolves symlinks:
```
file_status status(const path& p, error_code& ec) noexcept;

...

Remarks: If a symbolic link is encountered during pathname resolution, pathname resolution continues
using the contents of the symbolic link.
```
So, as long as compiler implementers adhere to the standard, I'm convinced, that both those functions resolve symlinks. I will remove the symlink-resolving code. It was completely wrong anyway - I have created the `resolvedSymlink` variable, which I then failed to use. It seems that it went through the CI linters just fine though...

> And yes, a comment in this case IMHO is helpful, to make sure, that we do not get an "improvement" pull request wtihout need.

Alright, I'll add a comment with a short explanation and a link to cppreference.

syyyr: I have also changed the error message to indicate that `lookupDir` might not be a directory.

syyyr:
By the way, are you okay with me just sending fixes like these? It seemed to me like a quick and easy fix and for those, I tend to make PRs straight away without first making an issue with an exact explanation of the problem (and a reproducer). I wouldn't want to spam the repo with fixes that can't be reproduced by anyone else, so feel free to tell me if I should change my approach in any way. Thanks.

Framstag:
> By the way, are you okay with me just sending fixes like these? It seemed to me like a quick and easy fix and for those, I tend to make PRs straight away without first making an issue with an exact explanation of the problem (and a reproducer). I wouldn't want to spam the repo with fixes that can't be reproduced by anyone else, so feel free to tell me if I should change my approach in any way. Thanks.

Libosmscout is a library for its users. As such I definitely accept such patches, if they improve the overall codebase and its stability or resilliance. Or add features that make sense in the code base. 
I'm more. Strict if it comes to portability, dependencies or similar high impact stuff

libosmscout-client/src/osmscoutclient/MapManager.cpp
Karry: Sorry, I was no looking to documentation before this comment. Thank you to do so :-)

Karry (approved):
LGTM, thank you!

Karry:
> By the way, are you okay with me just sending fixes like these? It seemed to me like a quick and easy fix and for those, I tend to make PRs straight away without first making an issue with an exact explanation of the problem (and a reproducer). I wouldn't want to spam the repo with fixes that can't be reproduced by anyone else, so feel free to tell me if I should change my approach in any way. Thanks.

I am more than happy with such merge requests ;-) I am creating issues just in case when I have no time for fix it right away. When it is simple like this one, there is no reason to waste time with describing problem in issue. Just my opinion.

PR #1531, opened 2023-12-01
fix cmake install stage for stylesheets directory

janbar:
As sub-project the install stage fails to find stylesheets directory from CMAKE_SOURCE_DIR as it is the top source path. Use instead PROJECT_SOURCE_DIR, which is the top source path of the project libosmscout.


Karry:
LGTM. But I don't get how CMAKE_SOURCE_DIR and PROJECT_SOURCE_DIR may lead to different directories in this cmake file?

janbar:
by this way:
```
project(dummy) -> Top of source
... CMAKE_SOURCE_DIR -> $DUMMY_SOURCE
add_subdirectory(libosmscout)
```

PR #1555, opened 2024-01-30
Overriding read write lock

janbar:


Karry (approved):
LGTM, just minor suggestions... I like that this implementation is working properly (with write preference) on all platforms, not just with pthread. But I am unable to say if it is 100% correct ;-)

Seems that Sonar and Android builds are broken in general, not just for this branch. They are failing even on master branch: https://github.com/Framstag/libosmscout/actions/runs/7430737711/job/21037909684 , https://github.com/Framstag/libosmscout/actions/runs/7430737698/job/21039440667

libosmscout/include/osmscout/async/ReadWriteLock.h
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  void swap(ReadLock& rl) noexcept {
+    Latch *_p = p;
+    bool _owns = owns;
+    p = rl.p;
+    owns = rl.owns;
+    rl.p = _p;
+    rl.owns = _owns;
+  }
```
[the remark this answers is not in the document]
janbar: Yes it could be for our usage.

janbar:
Yes it is a tricky piece of code. And yesterday by running a benchmark versus pthread , I discovered 2 bug in pthread_rwlock on Linux x64 , and OSX. Those bug will crash the app for sure. In fact pthread is a nice lib, but never use the pthread_rwlock in critical project.

janbar:
@Framstag , please wait. I have an other commit to improve the locking strategy used by the rendering jobs.

Karry:
> Completed : counter=204960

@janbar it looks like a pretty serious bug. Did you were able to reproduce it on Linux, or just on MacOS? It would be great to report it to upstream...

janbar:
> > Completed : counter=204960
> 
> @janbar it looks like a pretty serious bug. Did you were able to reproduce it on Linux, or just on MacOS? It would be great to report it to upstream...

This bug is not on Linux, but I found an other with a different behavior. I will show you later.

janbar:
The last commit "delegates the lock state to the task" fixes the thread priority issue during navigation. @Karry could you have a look please ?
It delegates the management of the lock to the task. Each job on startup check if the shared lock can be acquired, if no then the job return.

Framstag:
@janbar, @Karry : Can I merge?

janbar:
In my opinion YES. Waiting @Karry.

libosmscout-client/src/osmscoutclient/DBThread.cpp
```diff
@@ -508,18 +508,20 @@ CancelableFuture<bool> DBThread::FlushCaches(const std::chrono::milliseconds &id
...
-  std::shared_lock locker(lock);
+  ReadLock locker(latch);
   if (!isInitializedInternal()){
     locker.unlock();
     osmscout::log.Warn() << "ignore request, dbs is not initialized";
     return;
   }
+  // delegates the lock state to task
+  locker.unlock();
   job(basemapDatabase,databases,std::move(locker));
```
Karry, line 519: This is not correct, when you are not holding read-lock, list of database may be changed!

libosmscout/include/osmscout/async/ReadWriteLock.h
```diff
@@ -0,0 +1,206 @@
...
+
+  void prefer_exclusive(bool yesno) { px = yesno; }
+
+  void lock();
+  void unlock();
+
+  void lock_shared();
+  void unlock_shared();
+
+  bool try_lock_shared();
```
Karry, line 82: please add documentation, it is not very clear when it will fail...

Karry (changes requested):
The last commit may break synchronization, it is necessary to hold read-lock during the time when task may access to list of databases.

As you wrote here: https://github.com/Karry/libosmscout/pull/2#issuecomment-1920244282 style change (requiring write lock) may takes 10 seconds. But it is expected when rendering (holding read lock) takes so long sometimes (on mobile device).

libosmscout-client-qt/include/osmscoutclientqt/MapRenderer.h
```diff
@@ -68,9 +68,9 @@ class OSMSCOUT_CLIENT_QT_API DBRenderJob : public QObject, public DBJob{
 
   void Run(const osmscout::BasemapDatabaseRef& basemapDatabase,
            const std::list<DBInstanceRef> &allDatabases,
-           std::shared_lock<std::shared_mutex> &&locker) override;
+           ReadLock &&locker) override;
 
-  inline bool IsSuccess() const{
+  inline JobStatus  IsSuccess() const{
```
Karry, line 73: Naming is confusing now. But more important is, that there are some places when this enum is assigned to boolean. And assign other value than 0/1 to boolean is undefined-behaviour...

libosmscout-client/src/osmscoutclient/DBThread.cpp
janbar: Here the call of job() is synchronous, so I can call try_lock_share() near and no longer delegate the lock to the task. To be tested.

libosmscout-client-qt/include/osmscoutclientqt/MapRenderer.h
janbar: All places have been updated to handle enum in the commit. May be I missed one ...

libosmscout/src/osmscout/async/ReadWriteLock.cpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+void Latch::lock_shared() {
+  spin_lock();
+  if (x_owner != std::this_thread::get_id()) {
+    /* if flag is 0 or 1 then it hold S with no wait,
+     * in other case it have to wait for X gate
+     */
+    for (;;) {
+      if (x_flag.load() < 2) {
```
Karry, line 112: when x_flag is one (`X is held, but waits for release of S`) should not be new shared lock waiting?

janbar: Not at all ! if you do that it is a deadlock

libosmscout/src/osmscout/async/ReadWriteLock.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include <cassert>
+
+namespace osmscout {
+
+/**
+ * X flag is defined as the following:
+ * 0 : X is released
+ * 1 : X is held, but waits for release of S
+ * 2 : X is held
+ * 3 : X was released and left available for one of request in wait
```
Karry, line 35: I would suggest to use named contants (constexpr) for these meanings...

janbar: mmm, using addl vs mov is better for performance. Also x_flag is atomic, and I use as much as possible "lock add/sub 1"

libosmscout/src/osmscout/async/ReadWriteLock.cpp
Karry: can you please elaborate it more? how can I get deadlock?

During my testing, I get to the state, when one thread is waiting for exclusive lock (`x_flag=1`), then shared lock was released and immediately acquired another shared lock... I would expect that second shared lock will wait until exclusive one is properly acquired and released...

libosmscout-client-qt/include/osmscoutclientqt/MapRenderer.h
Karry: `PlaneMapRenderer:403`

libosmscout/src/osmscout/async/ReadWriteLock.cpp
Karry: I would just replace numeric constants like: `if (x_flag == 0)` with:
```
constexpr int X_RELEASED = 0;
...
if (x_flag == X_RELEASED)
```
generated assembler will be the same

libosmscout/src/osmscout/async/ReadWriteLock.cpp
Karry: you are right, I get deadlock, but I am just blind - don't understand how :-/

libosmscout/include/osmscout/async/ReadWriteLock.h
janbar: It is the same behavior than standard mutex or shared_mutex.

libosmscout/src/osmscout/async/ReadWriteLock.cpp
Karry: I want to solve this case:
```
    osmscout::Latch latch;
    std::atomic_int i=0;
    osmscout::ReadLock rl(latch);

    std::thread t([&latch, &i](){
      osmscout::WriteLock wl(latch);
      i++;
    });

    std::this_thread::sleep_for(1s);

    rl.unlock();
    rl.lock();

    assert(i==1);

    t.join();
```
this code may hit the assert... but it should be guaranteed that second shared lock will to be locked before exclusive one...

Karry: Ah, I see... The deadlock will happen with re-entrant shared locking. Damn!

janbar: for more explanation: blocking a shared lock which is by nature re-entrant you will go straight into a dead end == deadlock.

Karry:
Please keep in mind that DBThread::RunJob is not used just by renderer. But also by data loading and it is part of public api. It is not nice that job may be canceled just in case that lock is hold already. It should wait IMHO. 

- I would rather rework lock to really prefer writers (it would require the knowledge about all threads holding the shared lock)
- or rewrite locking to awoid re-entry. But it would be hard.

janbar:
> Please keep in mind that DBThread::RunJob is not used just by renderer. But also by data loading and it is part of public api. It is not nice that job may be canceled just in case that lock is hold already. It should wait IMHO.

The data loading get writer lock before running dbthread job, so it wouldn't be cancelled.

janbar:
> * I would rather rework lock to really prefer writers (it would require the knowledge about all threads holding the shared lock)

Here , it is already done: by calling "try_lock_shared()".

janbar:
> * or rewrite locking to awoid re-entry. But it would be hard.

I agree ; but we have to refactor all around jobs and maybe without success or real benefit, and for sure adding complexity.

Karry:
> > Please keep in mind that DBThread::RunJob is not used just by renderer. But also by data loading and it is part of public api. It is not nice that job may be canceled just in case that lock is hold already. It should wait IMHO.
> 
> The data loading get writer lock before running dbthread job, so it cannot be cancelled.

Well, I mean `DBLoadJob`, this job may be canceled without caller notice! It is not good api.

> If a reader (that can be re-entrant) is in wait you are facing to deadlock.

I understand, but it is possible to store knowledge about threads holding shared locks and not blocking threads that are holding lock already...

> the case you pointed, is when you have 2 writers, but it is not possible because writer is exclusive.

No, take a look to this simple test: https://github.com/Framstag/libosmscout/commit/815f6e37e33f586aaba25c53fcfe491d86c357c1 it is failing now... Writer has no precedence now.

janbar:
@Karry, you want few things that are incompatible. I explain below:

- You want threads go to sleep and so not loop in the spin_lock because it consumes cpu.
- You want sleeping threads wake up as if they are working (looping the spin).

The problem is the writer thread is in SLEEP (managed by he conditional variable). So the system requires some time to wake up the thread in this situation.

In your code you have to add a loop to wait the thread wake up after the line 33:
```
  while ( latch.get_x_flag() > 0 ) {
    osmscout::log.Error() << "c=" << latch.get_shared_count()
       << " x=" << latch.get_x_flag();
    std::this_thread::yield();
  }
```
Also we have to hint the compiler that "i" is volatile.
Adding that, I get the following.

    ./RwLockWriterPrecedence
    ===============================================================================
    All tests passed (2 assertions in 1 test case)

Karry:
But as try_lock_shared() may fail, it is hard to use it. As you wrote, even spimlock dont quaratnee that exclusive lock will be acquired first. I want to block shared lock, when exclusive one is wayting and thread is not holding lock yet. I will try to send some proposal implementation during weekend...

janbar:
@Karry, you are lucky ;) , it was soo simple ... I know I am lazy

Karry:
Switch state to "X_STEP_2" when last shared lock is released was good idea ;-) I am going to test it on the phone.

janbar:
It was a battle to make testing to work properly. So now this is OK.

janbar:
I tried with the phone. The precedence is better than before but not perfect, because in navigation job constantly hit the counter, and it must be down to zero.

With this commit, Map style change gets 3 ~ 15 sec.
Using try_lock , I got less than 1 sec.

@Karry, I let you decide what it better or acceptable

Karry:
you are right, it is better - it solves my first test. but second parallel reader is not blocked, here is test for it:
```
TEST_CASE("Second shared lock should be blocked when exclusive is requested") {
  osmscout::Latch latch;
  volatile int i=0;
  volatile int j=0;
  osmscout::ReadLock rl(latch);

  std::thread t([&latch, &i](){
    osmscout::WriteLock wl(latch);
    i++;
  });

  // wait until writer lock is requested
  while (true) {
    if (!latch.try_lock_shared()) {
      // writer lock is requested already
      break;
    }
    latch.unlock_shared();
    std::this_thread::yield();
  }

  std::thread t2([&latch, &j](){
    osmscout::ReadLock rl(latch);
    j++;
  });

  osmscout::ScopeGuard finalizer([&t, &t2]() noexcept {
    if (t.joinable()) {
      t.join();
    }
    if (t2.joinable()) {
      t2.join();
    }
  });

  using namespace std::chrono_literals;
  std::this_thread::sleep_for(2s);

  REQUIRE(i == 0); // write lock is still waiting
  REQUIRE(j == 0); // second read lock is blocked because exclusive lock is requested
  rl.unlock();
}
```

janbar:
I have a solution: we can create 2 functions:

DBThread::RunJob() and DBThread::TryRunJob()

Rendering job will use the try version. Is a good option ?

Karry:
> It runs, but it is slow and requires much cpu. 

It is too strong claim IMHO, when looking to implementation, it seems to me pretty simple: https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/libsupc%2B%2B/hash_bytes.cc , thread id is 64 bit integer on x86 with pthread...
 
> edit: this type of operations should be banned on a shared lock (hash, modulo etc). Efficiency is the same as using a mutex that serializes readers.

When reader operation tooks several seconds, serializing readers would be worse by magnitude. I believe that this implementation is still faster than QReadWriteLock used before. But I didn't measure it.

> What about my last solution ?
Yes, it can be used for rendering, on style change, new rendering is triggered after...

> Also, I didn't anderstand why DBLoadJob cannot be cancelled, on loading a new Map or changing style ...

Because there is no mechanism (now) that would restart loading on all possible usages. And it would complicate the api.

> Looking in the code I anderstood the flow: run DBloadob -> trigger on loaded do run DBRender

Yes, but it is not the only place where DBLoadJob is used (IconLookup, LookupModule). And consider other users of this api (future or current that we don't know about...).

janbar:
I will test it, first to check that resolves the case (not sure), and finally to re-run a benchmark and possibly to improve it.

Karry:
Thanks. There is 1:64 possibility of thread id conflict, it that case share lock will not be blocking...

janbar:
It was ran x 2.6. After few adjustments it runs x 2 and new is better than pthread on Linux intel. Still to do the killer test with the phone.

janbar:
So, the killer test with the phone has passed successfully ;-). Changing style in move gets 1 sec, so GO.

janbar:
typo in comments again ... fixed.

janbar:
The check for writer precedence could failed if the hash of the thread id has the same bucket, and one has failed (Ubuntu 22.04/gcc meson). It would be difficult to run this test.

Karry:
> The check for writer precedence could failed if the hash of the thread id has the same bucket, and one has failed (Ubuntu 22.04/gcc meson). It would be difficult to run this test.

You are right, I realized that problem when wrote about the id hash conflict... One possibility is to check thread ids in the test and skip check in case of conflict and possibly repeat the test - but it not nice :-/

Karry (approved):
I am satisfied now :-) @janbar are you still working on some fix/improvement? Both tests may be merged to one file and using catch framework, but functionality is finished...?

janbar:
@Karry, I am fixing testing. Maybe the last push will succeed.

janbar:
@Karry, that's finished.

janbar:
Rerun testing again to validate.

Framstag:
@Karry, @janbar Now :-)?

Karry:
> Really testing is a pain, with the sanitizer.

What kind of sanitizer? GCC thread sanitizer reports many false positives, I always consider it as a hints where to be careful, but nothing that is really wrong. The only sanitizer that is working for me was address sanitizer. Clang is not much better in this regards.

janbar:
In the test, I used detached threads, and even they are finished and destroyed, the sanitizer complained that some local stack addresses was in use. So I changed to not use detached threads and join them before exiting the test program.
Finally the testing work without issue in the last checks (Sanitize / gcc).

To fix the LOTTO in test of precedence, I launch 4 threads as reader, and at least 1 should be blocked. If later it is not enough then we launch more (6, 8 etc).

So, now you can merge it.

PR #1562, opened 2024-02-13
Read write lock fully reentrant

janbar:
- It is fully re-entrant: read/write/read , and so it allows upgrading of reader to writer
- No more collisions when testing recursive S
- At least 5% faster than previous

### DIFF
--- a/Tests/src/Latch.cpp
+++ b/Tests/src/Latch.cpp
@@ -487,6 +487,43 @@ TEST_CASE("Multi Recursive Writer Worker") {
   REQUIRE(refCounter == pc);
 }
 
+TEST_CASE("Check Re-Entrant One Reader Writer") {
+  osmscout::StopClock stopClock;
+  volatile int i=0;
+  osmscout::ReadLock rl(latch);
+
+  std::thread t([&i](){
+    osmscout::ReadLock rl0(latch);
+    {
+      osmscout::WriteLock wl(latch);
+      {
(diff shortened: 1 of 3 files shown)


Karry (approved):
LGTM, just curious if usage of `std::list` would not be enough...?

janbar:
> LGTM, just curious if usage of `std::list` would not be enough...?

Hehe, using std::list you will get 5 times slower. You are right, std::list use a chain internally, but we haven't access directly to the linked nodes, previous or next: We have to use the iterator. Also it allocates a new pointer for each added element, unless we implement a dedicated allocator ... much work for a so simple thing, and not really efficient.

janbar:
I tuned the spin lock, using TTAS instead TAS. It seems to be little better for L1-dcache misses:

TAS:

       3 569 453 619      L1-dcache-loads
          79 832 116      L1-dcache-load-misses            #    2,24% of all L1-dcache accesses

TTAS:

       3 278 236 651      L1-dcache-loads
          67 117 885      L1-dcache-load-misses            #    2,05% of all L1-dcache accesses

I can not do better. It can therefore be merged as you wish.

janbar:
Please wait before merge it. I am running some tests.

janbar:
I updated the spin lock to implement TTAS instead TAS. The performance is better again, and I cannot do better.

@Framstag , you can merge the PR as you wish.

PR #1563, opened 2024-02-18
close previous db-load job, an other one

janbar:
See commit 5b62a91.

And not rely on job destructor, it seems that Qt may postpone object deletion after deleteLater() invocation for very long time. In such situation data are still loading and job is holding database read-lock.

@Karry, an other one is fixed here. I catched it on mobile device (Android). I guess the issue appear too on device slower than desktop used to develop or test.

### DIFF
--- a/libosmscout-client-qt/src/osmscoutclientqt/PlaneMapRenderer.cpp
+++ b/libosmscout-client-qt/src/osmscoutclientqt/PlaneMapRenderer.cpp
@@ -413,6 +413,7 @@ void PlaneMapRenderer::DrawMap()
 
     if (loadJob->IsFinished()){
       // this slot is may be called from DBLoadJob, we can't delete it now
+      loadJob->Close();
       loadJob->deleteLater();
       loadJob=nullptr;
     }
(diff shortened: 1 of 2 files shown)


libosmscout-client-qt/src/osmscoutclientqt/PlaneMapRenderer.cpp
```diff
@@ -413,6 +413,7 @@ void PlaneMapRenderer::DrawMap()
 
     if (loadJob->IsFinished()){
```
Karry, line 414: But when job is finished, Close method should be no-op. Or I am miss something here?

Karry: I see, rendrers are setting `closeOnFinish` to false explicitly, so job is holding read-lock even when it is finished. It makes sense as renderer requires to have corret link to loaded data and used stylesheet...

Karry (approved):

Karry:
Please close loadJob in even on `TiledMapRenderer.cpp:491`

janbar:
> Please close loadJob in even on `TiledMapRenderer.cpp:491`

Finally we managed to fix all the annoying reentrant locks. +1

PR #1565, opened 2024-03-06
Remove unnecessary include

### LINKED ISSUE #1564: Build fails because Config.h isn't installed
Hi, since 8c875b6940026bf70a427f1061c31bf94960987d, `osmscout/private/Config.h` is included transitively inside public headers. However, `Config.h` is not installed, which means that applications can't include these public headers. Since it's inside the `private` directory, it probably isn't supposed to be installed, so I'm not sure what's correct solution.

syyyr:
Turns out it's not really needed.

Fixes #1564


syyyr:
Thanks for merging.

PR #1568, opened 2024-05-04
Update LabelLayouter code

Framstag:
- Rewrote Mask and Canvas
- Split Int and DoubleRectangle
- Added FileObjectRef to labels
- Changed sorting and processing order of labels
- Moved helper code to separate header
- Added tests for masking code

### DIFF
--- a/libosmscout-map/include/osmscoutmap/LabelLayouterHelper.h
+++ b/libosmscout-map/include/osmscoutmap/LabelLayouterHelper.h
@@ -52,15 +52,13 @@ namespace osmscout {
     {
     }
 
-    ScreenPixelRectangle& Set(int nx, int ny, int nw, int nh)
-    {
-      x = nx;
-      y = ny;
-      width = nw;
-      height = nh;
-      return *this;
-    }
-
+    /**
+     * Returns true , if the area of the rectangles intersect. The area is defined by
+     * area including [x-x+width-1] and [y,y+height-1]. x+width or y+height is outside the rectangle.
+     *
+     * @param other rectangle
+     * @return true, if areas intersect, else false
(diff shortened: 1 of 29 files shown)


libosmscout-map/include/osmscoutmap/LabelLayouterHelper.h
```diff
@@ -0,0 +1,169 @@
...
+
+    int GetFirstCell() const {
+      return cellFrom;
+    }
+
+    int GetLastCell() const {
+      return cellTo;
+    }
+
+    uint64_t GetCell(size_t idx) const;
```
Karry, line 151: please add documentation. at least that cell bites are "little-endian" so, `0x01` is the first pixel (left) and `0x8000000000000000` is the last pixel of the the cell.

I do not remember such implementation details anymore :-)

libosmscout-map/include/osmscoutmap/LabelLayouterHelper.h
```diff
@@ -0,0 +1,169 @@
...
+    ScreenPixelRectangle& Set(int nx, int ny, int nw, int nh)
+    {
+      x = nx;
+      y = ny;
+      width = nw;
+      height = nh;
+      return *this;
+    }
+
+    bool Intersects(const ScreenPixelRectangle &other) const
```
Karry, line 64: documentation was lost ;-)
```
     * Test if this Rectangle intersects with another.
     * It is using open interval, so if two rectangles are just touching
     * each other, these don't intersects.
     *
     * @param other
     * @return true if intersects
```

Karry (approved):
seems to be fine

PR #1572, opened 2024-05-17
Lane suggestion rewrite

Karry:
Suggested route lanes are based on turn geometry now (`DirectionDescription`).

But based on my experience and extensive report https://github.com/Karry/osmscout-sailfish/issues/326 , it provides misleading results on many junctions. So, it should analyze lanes and angles of all ways connected to the analyzed junction node...

### DIFF
--- a/Tests/CMakeLists.txt
+++ b/Tests/CMakeLists.txt
@@ -147,6 +147,9 @@ osmscout_test_project(NAME GeoCoordParse SOURCES src/GeoCoordParse.cpp)
 osmscout_test_project(NAME NumberSet SOURCES src/NumberSet.cpp)
 
 #---- ScanConversion
+osmscout_test_project(NAME RoutePostprocessor SOURCES src/RoutePostprocessor.cpp)
+
+#---- ScanConversion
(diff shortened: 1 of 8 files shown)


Karry:
it works on simple junction, will need to test it with more complex one...

Karry:
there is still some room for improvements (see TODOs in the tests), but it behaves better than current solution IMHO...

janbar:
Hi @karry, the routing crash on the following assert:
```
osmin: /home/jlb/src/janbar/QML/osmin-qt6/backend/libosmscout/libosmscout/src/osmscout/routing/RoutePostprocessor.cpp :1978 :
virtual osmscout::NodeRef osmscout::RoutePostprocessor::GetNode(const osmscout::DBFileOffset&) const:  l'assertion « entry!=nodeMap.end() » a échoué.
```
Map: Europe/France/Ile-de-france version 26 build on 2024-08-05 from osm data
Departure from: N 48°33.798' E 2°14.502'
Arrived to: N 48°27.377' E 2°03.320'

Has I need to rebuild the map version 26, bold before this PR ?

Karry:
Thank you for the notice. Here is the fix: https://github.com/Framstag/libosmscout/pull/1599

PR #1579, opened 2024-06-07
Use std::execution functionality with MSVC

frank-aurich:
Visual Studio also supports `std::execution` functionality, when C++17 is enabled it works out of the box.
No additional library like with gcc/clang is required.

This MR enables `std::execution` for MSVC by default.
I also modified the checks in `libosmscout-client` and `libosmscout-import` to only include the TBB lib where needed.


Framstag:
Thanks for you pull request!

PR #1580, opened 2024-06-08
Remove compiler warnings

Framstag:


libosmscout-client-qt/src/osmscoutclientqt/RouteDescriptionBuilder.cpp
```diff
@@ -69,23 +69,23 @@ static QString TurnCommandType(const RouteDescription::DirectionDescriptionRef&
 static QString ShortTurnCommand(const RouteDescription::DirectionDescriptionRef& directionDescription)
 {
   if (!directionDescription){
-    return osmscout::RouteDescriptionBuilder::tr("Turn");
```
Karry, line 72: I am worry that qt translation is not working with full namespace :-/ Will need to check

Framstag: OK, there ws a warning regarding namespaces, so I tried...

I'm still struggeling with glm+meson+OS X. Looks like I needed to make some updates, to get better C++20 support.

Framstag: See the gcc and cmake builds (still contain such wearnings)

Karry: this small comment fix lupdate: https://github.com/Framstag/libosmscout/pull/1598

PR #1585, opened 2024-07-14
feat: New features "Brand" and "ChargingStation"

Framstag:


Karry (commented):
LGMT. Just keep in mind that adding new feature to database is not backward compatible, please increase database version and re-import test region database in `Tests/data/testregion`

PR #1589, opened 2024-08-17
workaround for screen touch/pointer event with qt6

janbar:
Make it work with qt-6.7.2. The goal is to workaround bugs and issues with major changes from qt6.


janbar:
It fixes the no-longer-work of gestures when compiling with qt6 on Linux, Android, and MacOS. I didn't tested it on other platforms, but it should fix too.

Karry (approved):

Karry:
not tested on Qt6, but looks fine

PR #1591, opened 2024-08-30
Fix crash during style parsing

janbar:
During the refactoring of the StyleEditor, I found these issues in the parsing stage.
This fixes a "bypass" of symbol errors during parsing. At least the first error shouldn't be bypassed, else later it will try to deref nullptr after some dynamic cast, and finally crash with segfault.

### DIFF
--- a/libosmscout-map/src/osmscoutmap/oss/Parser.cpp
+++ b/libosmscout-map/src/osmscoutmap/oss/Parser.cpp
@@ -61,7 +61,7 @@ void Parser::SynErr(int n)
 
 void Parser::SemErr(const char* msg)
 {
-  if (!state) {
+  if (!state && errors->hasErrors) {
     return;
   }
 
--- a/libosmscout-map/parser/OSS/Parser.frame
+++ b/libosmscout-map/parser/OSS/Parser.frame
@@ -141,7 +141,7 @@ void Parser::SynErr(int n)
 
 void Parser::SemErr(const char* msg)
 {
-  if (!state) {
+  if (!state && errors->hasErrors) {
     return;
   }
 


janbar:
Also, it is difficult to edit the file Parser.cpp using an IDE. Because the file contains different type of indentation, and many other lines of code are updated on save. I could add a distinct commit to fix the indentation, dedicated for that. (same with the header).

libosmscout-map/src/osmscoutmap/oss/Parser.cpp
```diff
@@ -61,7 +61,7 @@ void Parser::SynErr(int n)
 
 void Parser::SemErr(const char* msg)
 {
-  if (!state) {
```
Framstag, line 64: Note that Parser.cpp is generated from Parser.frame by coco, see generate.sh. You can change it here, but you must also apply the fix in he frame file, else Parser.cpp will be overwritten the next time, I change the parser...

Framstag (changes requested):

libosmscout-map/src/osmscoutmap/oss/Parser.cpp
janbar: Ah ! okay

janbar:
"OSS/Parser.frame" has been updated too.

janbar:
@Framstag ,

It breaks the parsing when FLAG Daylight change from true to false. But no crash. We have to fix something elsewhere:

https://github.com/Framstag/libosmscout/blob/47d95b6d721c55c5717a1697853867f2584f914f/libosmscout-map/src/osmscoutmap/StyleConfig.cpp#L1618

Seems the boolean "state" allowed to discard errors, avoiding a failure when loading some module.
An other solution is to test all deref of nullptr after a dynamic_cast in Parser.cpp.

janbar:
Later soon, I will try to fix the deref of nullptr, if you think is the good way.

PR #1592, opened 2024-08-31
parser oss: check deref nullptr, therefore revert commit cf8d1c746aa5…

janbar:
For instance, all errors should be discarded when state is false. So here the commit implements all check or flow to avoid deref nullptr.

### DIFF
--- a/libosmscout-map/parser/OSS/OSS.atg
+++ b/libosmscout-map/parser/OSS/OSS.atg
@@ -544,8 +544,7 @@ PRODUCTIONS
 
                       SemErr(e.c_str());
                     }
-
-                    if (!errors->hasErrors) {
+                    else {
                       config.AddConstant(name,std::make_shared<StyleConstantColor>(color));
                     }
                   }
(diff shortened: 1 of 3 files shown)


janbar:
@Framstag , this is the good fix. The style parsing works great now and there is no longer side effect.

Karry:
Hi @janbar , this change breaks rendering of `highway_path` with https://github.com/Karry/osmscout-sailfish/blob/master/stylesheets/outdoor.oss stylesheet. There is this warning in the log:
```
Line style for way XXX of type highway_path results in empty mainSlotWidth
```

Do you have idea what is broken? I don't review your changes yet. Just done some git archeology :-)

janbar:
@Karry, trying and get the same, but don't know the root cause.

Capture d’écran du 2024-09-07 11-23-27

In the example above, changing "DisplayWidth" to 0mm, the warning disappears. Any idea ?

janbar:
Same case here:
Capture d’écran du 2024-09-07 12-31-21

janbar:
@Karry, I found an issue with the color value parsing. I prepare a new PR.

janbar:
I pushed fixes with the PR #1597

Karry:
Thank you, it fixes the issue.

PR #1594, opened 2024-09-02
Cleanup and more optimization for styleeditor

janbar:
Here,

- I remove old source files no longer used (FileIO and the dot pro).
- The page move operations in the editor are now native.
- Use a preferred font depending of the platform: The old font "Courier" has no fixed height by weight, which breaks the alignment of the lines ruler.
- Refactor the StyleAnalyser as a private child of the syntax highlighter: The connections are managed by the highlighter.
- Let the thread analyser to finish before deletion (delete later)
- Delay parsing request to discard unnecessary request for analysis
- Minor fixes for deprecation etc...


janbar:
@Framstag , it is ready for reviewing

PR #1597, opened 2024-09-07
fix style parsing 3

janbar:


janbar:
@Karry, this will fix the issue.

I check if there is others cases, where only a constant is processed ...

janbar:
I found only those cases.

PR #1600, opened 2024-09-08
Navigation Voice for Qt6 and Qt5

janbar:
I propose a change to manage the major changes in the Qt framework multimedia, from qt6. And then enable Navigation Voice for the lib compiled with the latest major version (6), and the previous (5) too.
The idea is to show an agnostic player, that wrap a core player depending of the Qt version at compile time.

The core implementation is in a local header (VoiceCorePlayerImpl.h), and it should be hidden.

For instance, the voice player isn't registered as qml type, but it could be if we need in future. e.g to play an alert message or something else.

### DIFF
--- a/libosmscout-client-qt/CMakeLists.txt
+++ b/libosmscout-client-qt/CMakeLists.txt
@@ -54,6 +54,8 @@ set(HEADER_FILES
     include/osmscoutclientqt/InstalledMapsModel.h
     include/osmscoutclientqt/Voice.h
     include/osmscoutclientqt/VoiceManager.h
+    include/osmscoutclientqt/VoicePlayer.h
+    include/osmscoutclientqt/VoiceCorePlayerImpl.h
     include/osmscoutclientqt/AvailableVoicesModel.h
     include/osmscoutclientqt/InstalledVoicesModel.h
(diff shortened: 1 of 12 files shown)


janbar:
It seems there is an issue using **meson** and qt **moc**: QT_VERSION wasn't passed to moc. I could split the the header VoiceCorePlayer.h by 2 files, first for qt5 and second for qt6. Then the check for QT_VERSION is moved to the source VoicePlayer.cpp to include the right header.

janbar:
Seems meson has a limited support with qt moc. The moc stage fails about preprocessing ```#if QT_VERSION ... #endif```. The workaround is to split the header, and make the conditional step in meson.build.

janbar:
Now the meson build works !
Well, "meson" is not great with qt. It took me about thirty attempts to get the build to work. I don't use it here, and ultimately I won't use it.

janbar:
Documentation added.

@Karry , please could you do a review ?

janbar:
try to fix sonar issue (a false positive)

Karry (approved):
Looks good. Thank you! I tested just Qt5 implementation.

PR #1612, opened 2024-10-02
try to fix crash starting routing on highway

janbar:
laneTurns = an empty vector
allowedLaneFrom = 0
allowedLaneTo = 0

```
* thread #33, name = 'Router', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x0) frame #0: 0x00005555558eb30e osmin`osmscout::RoutePostprocessor::SuggestedLanesPostprocessor::EvaluateLaneSuggestion(osmscout::PostprocessorContext const&, osmscout::RouteDescription::Node const&, std::__cxx11::list<osmscout::RouteDescription::Node*, std::allocator<osmscout::RouteDescription::Node*>> const&) const at RoutePostprocessor.cpp:1842:44 1839
   1840	    // detect what all allowed turns have common
   1841	    // keep in mind that laneTurns may have removed directions used by exits, it is benefitial here
-> 1842	    uint32_t suggestedTurnBits = TurnToBits(laneTurns[allowedLaneFrom]);
   1843	    for (int i = allowedLaneFrom + 1; i <= allowedLaneTo; i++) {
   1844	      suggestedTurnBits &= TurnToBits(laneTurns[i]);
   1845	    }
```

### DIFF
--- a/libosmscout/src/osmscout/routing/RoutePostprocessor.cpp
+++ b/libosmscout/src/osmscout/routing/RoutePostprocessor.cpp
@@ -1776,6 +1776,9 @@ namespace osmscout {
     int allowedLaneFrom = 0;
     int allowedLaneTo = prevLanes->GetLaneCount()-1; // inclusive
     std::vector<LaneTurn> laneTurns = prevLanes->GetLaneTurns();
+    if (laneTurns.empty()) {
+      return;
+    }
 
     // remove allowed lanes used for left exits
     for (const auto &exit: junctionLeftExits) {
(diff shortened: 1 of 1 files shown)


janbar:
@Karry , it is a suggested fix. For sure there is a route cause ...

janbar:
[test.gpx.gz](https://github.com/user-attachments/files/17232789/test.gpx.gz)

You could use my simulator to check the issue on Linux:
Start the osmin-simulator in a bash, then launch osmin, and finally type the following commands in the simulator: 
```
load test.gpx
run 1 1 280
```
The simulation runs starting from pts 280 => You are on the highway. So in osmin, you have to click the button burger -> Routing, then choose a destination, and click the icon Car.

Karry:
Thank you for pointing it out! But the fix you are proposing is not enough. List of turns may be non-empty, but its size may be smaller than count of lanes. It should be checked on every access. Moreover, it is beneficial to continue in lane evaluation even when lane turns are not defined...

Proposing this fix: https://github.com/Framstag/libosmscout/pull/1613

janbar:
@karry Thank for the fix. I close this.

PR #1614, opened 2024-10-04
Adjust the pixel ratio of tiles for a better visual rendering

janbar:
- To make it as similar as possible to plane rendering
- To fix the huge aliasing defect when a magnification is applied

I attached a set of screenshots to show the enhancement for ours eyes, and it doesn't require more cpu time !

Offline tiles: The left is without this commit, the center has the commit, and the right is plane rendering.

Capture d’écran du 2024-10-04 14-54-26

Capture d’écran du 2024-10-04 14-45-51

Note that the antiliasing isn't enable on tiled views, because the shot has done during animation. By forcing the hints to enable antiliasing, the unpatched view (left) shows blurred contours (lines, areas, icons etc...).

---
Online tiles: Finally, I tested too with online tiles: the same benefits, the rendering is much better as before, even zooming a level:

Capture d’écran du 2024-10-04 16-15-06
A view when I zoom the level:
Capture d’écran du 2024-10-04 16-31-03

### DIFF
--- a/libosmscout-client-qt/src/osmscoutclientqt/TiledRenderingHelper.cpp
+++ b/libosmscout-client-qt/src/osmscoutclientqt/TiledRenderingHelper.cpp
@@ -65,10 +65,16 @@ bool TiledRenderingHelper::RenderTiles(QPainter &painter,
     }
   }
 
+  // enable subpixel rendering so that magnification up to 2x remains
+  // as smooth as possible
+  static constexpr double pitchFactor = 1.5;
+  // the magnification level is adjusted taking into account the pixel
+  // ratio of the request and the fixed dpi of the OSM tile
   projection.Set(request.coord,
                  0,
-                 request.magnification,
-                 request.dpi,
+                 Magnification(request.magnification.GetMagnification()
+                               * pitchFactor * request.dpi / OSMTile::tileDPI()),
+                 OSMTile::tileDPI() / pitchFactor,
                  width,
                  height);
 


janbar:
Great !
I will prepare the production release for osmin 1.13 (1.13.1), which will based on the latest libosmscout.

Karry (commented):
Sorry, but I am confused by this change, can you elaborate it more deeply, please? :-/ request.dpi should be the screen DPI (by default). It may be adjusted by user settings, but still, when renderer renders tiles 512x512, app should render it with exactly such size. Rendered tiles should not be up-scaled. I do not understand why are you getting such nasty anti-aliasing....

janbar:
@karry, if you think the pitch is too thin we can adjust the pitchFactor to 1.0.

Karry:
Hi @janbar. 

Sorry for the delayed reply, it is hard for me to find some free time for hacking now.

Tiles are rendered with the same DPI as used for sticking final map canvas. Problem is that during zoom, when you are just "below" next magnification level (for example 16.999), tiles are upscaled almost 2x. And as I found just now, Qt "lies" about canvas resolution on HiDPI screens. When I have 25% upscale on my 4k monitor and render 100px rectangle on the Qt canvas, it is displayed as 125px rectangle on the screen. It kind of makes sense, but libosmscout doesn't take [`QSreen::devicePixelRatio`](https://doc.qt.io/qt-6/qscreen.html#devicePixelRatio-prop) into account. 

As a result, these two factors makes tiled map ugly without antialiasing.

Back to your merge request. I have suspect that resulted map doesn't correspond to required result by stylesheet. In other words, style define that buildings are displayed on zoom level 16, but in your case they will be displayed later. But would need more testing to confirm this hypothesis. Another tricky problem is that stylesheet may define some dimensions in millimeters, so it is necessary to use proper DPI during whole renderign process :-)

Anyway, I will try to prepare alternative solution for this problem.

Karry:
And for online (raster) tiles, they will always ugly, when they are rendered with low DPI. I would suggest to enable anti-aliasing even during animations. Possibly by configuration?

PR #1616, opened 2024-10-13
feat: New DescriptionService

Framstag:
- New DescriptionService that offers structured object information
- Adaption of DumpData to make use of the new service
- Changes to map.ost to make use of more available features


Framstag:
Dropped support for Ubuntu 20.4, GCC 10. Possibly could make it work, but don't want to waste by time for such an old compiler.

Issue #1619, opened 2024-10-27
ThreadTest fails under MSYS


Framstag:
The test currently does fail under MSYS.

Framstag:
Digging a little bit deeper and executing the test in gdb (code with additional std::cout lines):

```
$ meson test -C debug --gdb --print-errorlogs "Check Thread utilities"
ninja: Entering directory `C:/projects/libosmscout/debug'
ninja: no work to do.
1/1 Check Thread utilities RUNNING
>>> PATH=C:/projects/libosmscout/debug/libosmscout;C:/msys64/mingw64/bin;C:\msys64\mingw64\bin;C:\msys64\usr\local\bin;C:\msys64\usr\bin;C:\msys64\usr\bin;C:\Windows\System32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\msys64\usr\bin\site_perl;C:\msys64\usr\bin\vendor_perl;C:\msys64\usr\bin\core_perl;C:\projects\libosmscout\debug\libosmscout MESON_TEST_ITERATION=1 ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 MALLOC_PERTURB_=249 MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 gdb --quiet --args C:/projects/libosmscout/debug/Tests/ThreadTest.exe
――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――
Reading symbols from C:/projects/libosmscout/debug/Tests/ThreadTest.exe...
(gdb) run
Starting program: C:\projects\libosmscout\debug\Tests\ThreadTest.exe
[New Thread 23252.0x66dc]
[New Thread 23252.0x3494]
[New Thread 23252.0xf80]
Randomness seeded to: 2867844404
MainThread id: 1
Starting background thread...
[New Thread 23252.0x2b8c]
Joining background thread...Background thread id:
2
Connect...
Connect...done
Background thread...done

Thread 5 received signal SIGSEGV, Segmentation fault.
[Switching to Thread 23252.0x2b8c]
[...]

PR #1621, opened 2024-11-03
chore: Analyse ThreadTest problems, issue #1619

Framstag:
- Added further debug output
- Skipped test, if running under MINGW


Framstag:
I remove them from the test suite for MSYS. If you look for thread local support in MINGW you finds hints to problems. I have added a call stack in the original issue, perhaps someone has an idea, what is happening.

PR #1622, opened 2024-11-08
HiDPI support in tile map renderer

Karry:
Hi, here is support for HiDPI screens in tiled map renderer. May I ask you @janbar for the review please? 

It is fun on Wayland, when you can have different scale (pixel ratio) for each monitor. Then map is refreshed when window is moved from one monitor to another...

### DIFF
--- a/libosmscout-client-qt/src/osmscoutclientqt/TiledMapRenderer.cpp
+++ b/libosmscout-client-qt/src/osmscoutclientqt/TiledMapRenderer.cpp
@@ -28,6 +28,7 @@
 
 #include <QGuiApplication>
 #include <QScreen>
+#include <QtCore>
 
 namespace osmscout {
 TiledMapRenderer::TiledMapRenderer(QThread *thread,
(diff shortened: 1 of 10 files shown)


libosmscout-client-qt/src/osmscoutclientqt/TiledMapRenderer.cpp
```diff
@@ -418,10 +418,17 @@ void TiledMapRenderer::onLoadJobFinished(QMap<QString,QMap<osmscout::TileKey,osm
             (double)loadXFrom + (double)width/2.0,
             (double)loadYFrom + (double)height/2.0);
 
-    double osmTileDimension = (double)OSMTile::osmTileOriginalWidth() * (mapDpi / OSMTile::tileDPI() ); // pixels
+    /** tiles are up-scaled almost two times, when zoom level is near its next integer value (for example 16.99)
+     * to avoid blured result, we render tiles two times bigger than its minimum visual size.
+     * Another possible upscale (screenPixelRatio) is for HiDPI screens - when there is ratio 2.0, 100px on Qt canvas
+     * is displayed as 200px on the screen. To provide best results on HiDPI screen, we upscale tiles by this pixel ratio.
+     */
+    double finalDpi = mapDpi * 2 * this->screenPixelRatio;
```
Karry, line 426: I am not sure if we should not have this basic upscale configurable. Two is pretty huge performance impact :-/

janbar:
Thanks @karry I will try it as son as possible.

janbar:
Thanks @karry I will try it as soon as possible.

janbar:
@Karry , it is far better than before. No more blurred on zoom animation. Good work.

1 = Before, 2 = Now, 3 = with this PR

Capture d’écran du 2024-11-13 12-04-14


Capture d’écran du 2024-11-13 12-03-52

**Before (1):**
Capture d’écran du 2024-11-13 12-18-39

**Now (2):**
Capture d’écran du 2024-11-13 12-19-16

**This PR (3):**
Capture d’écran du 2024-11-13 12-19-41

Framstag:
Ok, fine that wie have an agreement. I‘ll merge

PR #1625, opened 2024-11-16
Use of std::format in Progress

Framstag:
- Added method variants using std::format for Progress
- Adapted code
- Added last version of WellScoutedRoute
- Added work in progress in OSMScout2 (and fixed search)


Framstag:
@Karry, @janbar I would like to drop support for Ubunu 22.04, too. IT would like work with a custom gcc-13, but I do not want to spent the time to fix builds.

Also, is the newer Android NDK OK?

Framstag:
@Karry , @janbar : Please, give some feedback!

Karry:
Hi Tim. I don't look to specific changes yet. I just tried to compile this branch with Sailfish OS SDK and it fail. It is using GCC 10 right now, without `<format>` support :-(

I am sorry, but I do not see simple way, howto use these changes. Is there boost variant of format library? It is available somehow for gcc 10?

Framstag:
@Karry: There seems to be https://github.com/fmtlib/fmt Do you have time to enhance the pull request and test it. Else I'll see around christms time.

Karry:
I looked into detail. As I do not need Import tools on Sailfish OS, and it is the only place where Progress utility is used, I may easily exclude Progress.cpp [and Progress.h](https://github.com/Framstag/libosmscout/pull/1632) from my build. 

It may work for me as a workaround until [gcc 13 will be available](https://github.com/sailfishos/gcc/pull/8) for SFOS. I would guess that it will be in the middle of 2025.

Framstag:
> I looked into detail. As I do not need Import tools on Sailfish OS, and it is the only place where Progress utility is used, I may easily exclude Progress.cpp [and Progress.h](https://github.com/Framstag/libosmscout/pull/1632) from my build.
> 
> It may work for me as a workaround until [gcc 13 will be available](https://github.com/sailfishos/gcc/pull/8) for SFOS. I would guess that it will be in the middle of 2025.

That sounds reasonable. I mentioned before that the importer in practice has other constraints and execution context than the rest of the libraries, since it will likely not run on the/a device.

PR #1635, opened 2025-01-05
allow ways and areas to have phone and website features

Karry:
I was curious why `amenity_restaurant_building` have no website, when I edited the restaurant month ago... I believe that there is no reason for such exclusion.


Framstag:
I agree. Assigning Features via OST file should be enough filtering. In general no need for further filtering within the Feature implementation.

PR #1647, opened 2025-04-17
fixes huge dpi on hidpi screen

janbar:
With hidpi the pixel size is already taken into account in the request dpi. So do not factorize again with the screen pixel ratio, and avoid OOM on device with limited resources.

### DIFF
--- a/libosmscout-client-qt/include/osmscoutclientqt/MapRenderer.h
+++ b/libosmscout-client-qt/include/osmscoutclientqt/MapRenderer.h
@@ -33,7 +33,6 @@
 #include <QSettings>
 #include <QMutex>
 #include <QPainter>
-#include <QScreen>
 
 namespace osmscout {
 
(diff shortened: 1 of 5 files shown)


janbar:
@Karry , I added a comment in the issue. This doesn't fix the OOM on less powerful devices.

janbar:
The new commits fixes the OOM now.
The up-scaling is completely cleared, and memory footprint is now the same as before PR #1622, or the double (caused by "use 'power of two' tile size"), but no more.
The blur issue is fixed by re-applying the commit "adjust the pixel ratio of tiles for a better visual rendering".

janbar:
As explained in the issue, the best compromise is to no longer up-scale, and so keeping best performances and availability on all devices. The second commit clean up all unused, that was required by the up-scaling feature.
Even reverting, it is an enhancement because we learn what we can and cannot expect. @Karry are you OK ?

PR #1648, opened 2025-04-21
render plane map without alpha channel

Karry:
it is not needed, it will save some memory
and RGB888 format should not be converted
in `QOpenGLTextureCache::bindTexture`


Karry:
Ups. We need alpha channel, when offline tiles are combined with the online one... I didn't realize that.

Framstag:
@Karry, should this still get merged?

Karry:
Yes. Alpha channel is not needed for plane rendering and it will save some memory...

PR #1649, opened 2025-04-22
configurable tile scaling to power of two 

Karry:
Library user may decide howto scale tiles to power of two. It may be
 - upscaled
 - downscaled
 - use nearest "power of two"
 - keep unchanged (there is small performance penalty for scaling the tile before sending to GPU)

When GL_ARB_texture_non_power_of_two extension is available, scaling is disabled.

### DIFF
--- a/libosmscout-client-qt/src/osmscoutclientqt/TiledMapRenderer.cpp
+++ b/libosmscout-client-qt/src/osmscoutclientqt/TiledMapRenderer.cpp
@@ -29,6 +29,8 @@
 #include <QGuiApplication>
 #include <QScreen>
 #include <QtCore>
+#include <QOpenGLContext>
+#include <QOpenGLFunctions>
 
 namespace osmscout {
 TiledMapRenderer::TiledMapRenderer(QThread *thread,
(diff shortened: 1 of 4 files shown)


janbar:
@Karry, trying the PR, i am facing with dead lock as explained below.

In RenderMap, Line 214:    QMutexLocker rendererLocker(&lock)

It could need a dedicated lock to protect the member glPowerOfTwoTexture.

libosmscout-client-qt/src/osmscoutclientqt/TiledMapRenderer.cpp
```diff
@@ -201,6 +204,16 @@ bool TiledMapRenderer::RenderMap(QPainter& painter,
   onlineTileCache.clearPendingRequests();
   offlineTileCache.clearPendingRequests();
 
+  if (QOpenGLContext *glContext = QOpenGLContext::currentContext();
+      glContext
+      && this->glPowerOfTwoTexture != GLPowerOfTwoTexture::NoScaling
+      && glContext->functions()->hasOpenGLFeature(QOpenGLFunctions::NPOTTextures)) {
+
+    // we have GL_ARB_texture_non_power_of_two extension available, there is no need to scale tiles
+    QMutexLocker rendererLocker(&lock);
```
janbar, line 213: There is no clue to lock for write if we don't lock for read. But testing to lock for read make a dead lock ...

janbar:
[fix_locker.txt](https://github.com/user-attachments/files/19965338/fix_locker.txt)

libosmscout-client-qt/src/osmscoutclientqt/TiledMapRenderer.cpp
Karry: but there is lock in onLoadJobFinished method already...

janbar: I tested by forcing "Nearest", and I am facing a dead lock. The dead lock occurs because 2 locks are hold in the function (tileCacheMutex and lock). It was the reason of my comment. The patch "fix2_locker.txt" fixes it. I don't know if the case can occur in an other situation ...

Karry: I see the problem now. Thank you for notice! I will use std::atomic for this value...

Karry:
I see the problem now. Thank you for notice! I will use std::atomic for this value... Can you review the last commit please?

janbar:
@Karry , reviewed and tested. it works !

janbar (approved):

Framstag:
@Karry , @janbar : This can get merged now?

Karry:
> @Karry , @janbar : This can get merged now?

yes, please

PR #1651, opened 2025-04-25
Stop copying QImage and save memory and cpu time

janbar:
Uses smart pointers and avoid copy all back buffers when setting/getting from cache.


Karry:
It is not necessary. QImage, QString, QArray and other Qt objects are behaving differently than objects from standard library. When you do copy of the QImage object, pixel data are referenced twice, from the original object and its copy. So, it is smart pointer by its design...

janbar:
You're right. I didn't realize it was like a container.

PR #1653, opened 2025-05-10
improve lane evaluation on complex junctions

Karry:


Karry:
Can we merge it @Framstag ? There are still few cases where heuristics is not working properly (see TODOs in the test), but I will try to address them in future merge requests...

PR #1659, opened 2025-07-11
feat: Simple MCP Server - Initial version of simple MCP Server

Framstag:


meson.build
```diff
@@ -223,6 +223,12 @@ else
   glfwDep = compiler.find_library('glfw', required: false)
 endif
 
+# cpp-httplib
+httpLibAvailable=compiler.has_header('httplib.h', required: false)
+
+# nlohmann_json
+jsonDep = dependency('nlohmann_json', required: false)
```
Karry, line 230: we have own copy of this library in `libosmscout-client/include/osmscoutclient/json`, I would consider to link MCPServer with libosmscout-client library to re-use asynchronous API ;-)

Karry (commented):
Thank you for kickoff this utility. I was considering to write some experimental MCP server using libosmscout recently :-) But my time for the experiments is limited now.

Anyway, where is CMake build?

meson.build
Karry: hmm, but httplib have no synchronous api, right?

Framstag:
I can merge anytime, if you would like to actively participate and supply further merge requests.

PR #1683, opened 2025-12-16
Update vcpkg Git commit ID for build workflow

Framstag:
Updated vcpkg to 2025.12.12


Framstag:
See https://github.com/microsoft/vcpkg/issues?q=is%3Aissue%20state%3Aopen%20pango

There are various bug tickets regarding build problems with pango. Problem still occurs with current vpkg commit. I have to wait for the bugs to resolve. Will see, if I can add further information to existing tickets.

PR #1684, opened 2025-12-19
use black text in StyleEditor epxlicitly

Karry:
As we are using while text background, text is unreadable with dark system palette (using bright texts).
So, using black color explicitly makes it readable.


Karry:
@janbar , your change is more complete (with style of the buttons), it looks good with dark theme. We may merge it as it.

PR #1705, opened 2026-04-02
Group multi-node junctions in SuggestedLanesPostprocessor

Karry:
Vibe-coded improvement of "suggested lane" heuristics... I am going to test with real word junctions and carefully review it first. 

---

Complex OSM intersections are modeled as multiple nodes connected by short internal segments that maintain the same lane configuration. The lane suggestion algorithm previously processed each node independently, seeing only a subset of exits and producing wrong lane recommendations.

Scan backwards through the route buffer to detect consecutive junction nodes that are close together (<50m) and share the same lane config, then aggregate their exits into a single evaluation. Filter exits pointing back toward the incoming direction to suppress opposite carriageway false positives in grouped junctions.

Fixes the Průmyslová/Černokostecká test case.

### DIFF
--- a/Tests/src/LaneEvaluationCompare.cpp
+++ b/Tests/src/LaneEvaluationCompare.cpp
@@ -26,7 +26,8 @@ struct SuggestionDiff
   std::optional<NodeLaneInfo> lanes;
   std::optional<NodeSuggestedLaneInfo> oldSuggestion;
   std::optional<NodeSuggestedLaneInfo> newSuggestion;
-  std::vector<RouteNodeInfo> routeNodes;
+  std::vector<RouteNodeInfo> oldRouteNodes;
+  std::vector<RouteNodeInfo> newRouteNodes;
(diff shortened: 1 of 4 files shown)


Framstag:
Is this ready for merge? How was your KI experience?

Framstag:
@Karry : Ready to merge?

Karry:
Not yet unfortunately. I found out that situation for some junctions get worse actually. Today, it is easy vibe-code tens of tools, but still you need time for writing prompts and validating the output... :-) 

I will post some testing tool later when I will have some time.

Framstag:
@Karry See https://github.com/Framstag/mesonsbom, where I used a combination of AI tools, to completely design a IMHO helpful tool. I could image to add support that approach to libosmscout, too (if token count for libosmscout stays in the sane area). Feedback appreciated. You can contact my directly, if you need further information.

Karry:
Results looks not better (compared to current master branch), we can merge it I guess.

PR #1708, opened 2026-04-21
StyleEditor: fix visible binding after scrolling

janbar:
An other workaround for TextArea to fix the binding of the cursor visibility after scrolling the area. The goal is to rebind the property when the cursor position is changed, i.e when clicking new position on scrolled area.


janbar:
@Framstag , it is ready to be merged. Thanks.

Framstag:
@janbar : Pull request is still in status "draft"!

janbar:
@Framstag , sorry for the delay. The PR is no longer in draft. You can merge it if you like it.

PR #1709, opened 2026-04-29
tool for evaluation changes in route lane heuristics

Karry:
Tooling to evaluate how SuggestedLanesPostprocessor changes affect
routing: generate random routes as JSON, re-compute after changes,
and compare lane suggestions with optional map image rendering.

---

Disclaimer: These tools are for manual testing. Completely vibe-coded after an hour of brainstorming with LLM. It provided me valuable output when testing https://github.com/Framstag/libosmscout/pull/1705 , so it works somehow. But I do not review the code properly yet, so it is a "draft" from my point of view.

@Framstag are you ok with vibe-coded changes after some level of review?

### DIFF
--- a/Tests/CMakeLists.txt
+++ b/Tests/CMakeLists.txt
@@ -334,3 +334,19 @@ osmscout_test_project(NAME PolygonCenterTest SOURCES src/PolygonCenterTest.cpp)
 #---- MercatorProjection
 osmscout_test_project(NAME MercatorProjectionTest SOURCES src/MercatorProjectionTest.cpp)
 
+#---- LaneEvaluation (manual evaluation tooling, not a CI test)
+set(_lane_eval_sources
+    src/LaneEvaluation.cpp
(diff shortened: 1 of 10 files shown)


Karry:
I am not sure with the copy of nlohmann/json library in the test directory. It is the second place where we are using this simple, but useful library (it is in libosmscout-client). Should not we require it as external dependency?

Framstag:
Yes usage of KI is OK, if the code is checked to be correct.

I also have no problem with checking in KI supporting artifacts as long:
* They are as much standard as possible
* Do not lock-in into a specific tool
* Do not hinder the work with another tool
* Does not require KI to get the software to build, test, release or run.

If this is always possible, has to be discussed.

I also do experiments with KI usage. I currently favor spec-driven approaches, but have not yet decided for a specific tool. Currently, all seem to be a little bit bloated and require high token usage, especially for brownfield development.

PR #1713, opened 2026-05-11
lookup parent admin region while building location index

Karry:
# Issue

In OpenStreetMap database, there is node [#419849333](https://www.openstreetmap.org/node/419849333) with these tags:
```
addr:conscriptionnumber 	45
addr:country 	CZ
addr:housenumber 	45
addr:place 	Číčovice
addr:postcode 	25268
ref:ruian:addr 	6406416
```

It is imported in latest Czech offline database, see:
```
./DumpData/DumpData -n 419849333 ../Maps/czech-republic-26-20260501-1046/

Node {
  OSM id: 419849333

  Id {
    Kind: Node
    Id: 48329145
  }

  Geometry {
    Coordinate: 50.15414 N 14.24125 E
  }

  General {
    Type: address
  }

  Location {
    Address: 45
  }

  lat: 50.15414
  lon: 14.24125
}
```

But the address lookup code do not see it, you can check by:

```
./Demos/LocationLookup --debug ../Maps/czech-republic-26-20260501-1046/ "Číčovice 45"
```

there is entry in `import.log` about this node:
```
cat ../Maps/czech-republic-26-20260501-1046/import.log | grep 48329145
     Node 48329145 - Street 'Číčovice' of address '45' cannot be resolved in region 'Velké Číčovice'
```

For small villages without named street, there is added "virtual" street / place if I remember correctly. It should match to it and be added to the index...

# Claude Code analysis (Opus 4.6 model)

Now I see the full picture. Here's the flow:

  1. Line 2222: The region is determined by geographic coordinates (GetRegionForNode) → this puts node 48329145 into "Velké Číčovice"
  2. Line 2232: AddAddressNodeToRegion is called with that region
  3. In AddAddressToRegion (line 1715), it calls FindLocation to find the location name "Číčovice" within "Velké Číčovice"
  4. FindLocation (line 2103) checks:
    - Existing locations in the postal area → not found
    - Whether location name matches region name ("Číčovice" vs "Velké Číčovice") → no match
    - Region aliases → no match
  5. Returns end() → "cannot be resolved" error

  The bug: when FindLocation fails in the node's geographic region, the code doesn't try parent regions. The virtual location "Číčovice" was created in "Malé Číčovice" (because that region has an alias "Číčovice"), but it doesn't exist in the sibling "Velké Číčovice".

● The hierarchy is:
  Číčovice (admin level 8)
  ├── Velké Číčovice (admin level 10) ← node is geographically here
  └── Malé Číčovice (admin level 10) ← has alias "Číčovice", virtual location created here

The parent region "Číčovice" (admin level 8) has the same name as the addr:place. So a fix could be: when FindLocation fails in the leaf region, try walking up to parent regions.

# Fix

Instead of looking in a single (smallest) region, we now iterate through all enclosing regions from smallest to largest (using the existing spatial index sorting), stopping at admin_level < 6 as a safety bound. The first region where the location name resolves wins.

### DIFF
--- a/libosmscout-import/include/osmscoutimport/GenLocationIndex.h
+++ b/libosmscout-import/include/osmscoutimport/GenLocationIndex.h
@@ -297,6 +297,26 @@ namespace osmscout {
 
     ImportErrorReporterRef errorReporter;
 
+    struct AddressData
+    {
+      ObjectFileRef object;
+      GeoCoord coord;
+      std::string name;
+      std::string location;
+      std::string address;
+      std::string postalCode;
+    };
+
+    struct AreaAddressData: public AddressData
+    {
+      std::vector<Point> nodes;
+      GeoBox boundingBox;
+    };
+
+    struct NodeAddressData: public AddressData
+    {
+    };
+
   private:
     void Write(FileWriter& writer,
                const ObjectFileRef& object) const;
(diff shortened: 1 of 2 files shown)


Karry:
It works now...
```
./Demos/LocationLookup  ../Maps/czech-republic-26-test/ "Číčovice 45"
Database:                ../Maps/czech-republic-26-test/
Search pattern:          Číčovice 45
Search for location:     true
Search for POI:          true
Admin region only match: false
POI only match:          false
Location only match:     false
Address only match:      false
Partial match:           false
Transliterate:           false
Limit:                   30

= Location (Číčovice) = Address (45) ~ PostalArea (25268) = Region (Číčovice)
   * Číčovice/SO POÚ Roztoky/SO ORP Černošice/okres Praha-západ/Středočeský kraj/Střední Čechy/Česko
   - Node 48327155 amenity_restaurant
= Location (Číčovice) = Address (45) ~ PostalArea (25268) = Region (Číčovice)
   * Číčovice/SO POÚ Roztoky/SO ORP Černošice/okres Praha-západ/Středočeský kraj/Střední Čechy/Česko
   - Node 48327232 address

Location search time: 0.083
```

Karry:
It took me two hours of investigation and brainstorming with Claude code, then few hours for import of the new database and one hour for performance optimization and manual refactoring...

PR #1741, opened 2026-07-06
Piper TTS for navigation instructions

Karry:
Navigation is using Voice-of-Marble pre-recorded samples for instruction messages. These instructions sounds natural and implementation is simple. But it lack of variability. There is no way howto add names to instructions or modify instructions somehow. For example for pedestrian navigation. For that reason, I decided to add support for navigation instructions generated by text-to-speech (TTS). (And keep Voice-of-Mable support for compatibility)

As TTS implementation I selected [Piper](https://github.com/OHF-Voice/piper1-gpl) (optional dependency of client-qt library). But abstraction allow to implement other clients easily.


Karry:
It is ready to be merged. I want to improve voice instructions later (add destination names, pedestrian instructions...), in separate merge requests.

PR #1745, opened 2026-07-30
feat: New skia backend

Framstag:
- New skia backend

Since Ubuntu 24.04 does not seam to have a skia package, this will likely not build yet.


Framstag:
The skia backend works and is optical (in my tests) on par with the Cairo and Qt backend. However, skia packaging seems to be a mess. There is an ArchLinux package, but it seems to be missing SVG support. MSYS package does not seem to run in github builds.

PR #1756, opened 2026-08-04
bug: Fixed basemap generation

Framstag:
- Coastline node IDs now use Point::GetId() instead of GeoCoord::GetHash()
- added maxCoastlinePoints=1000 chunking with overlapping endpoints for long shapefile records
- removed old continuation/coordBuffer logic.

WaterIndexProcessor::MergeCoastlines recomputes IDs with Point::GetId(). The old hash-based IDs caused chain merging to stop after the first join, leaving continental coastlines open and dropped. Chunking prevents future oversized records from breaking the merge.

### DIFF
--- a/BasemapImport/src/BasemapImport.cpp
+++ b/BasemapImport/src/BasemapImport.cpp
@@ -38,10 +38,10 @@
 class CoastlineShapeFileVisitor : public osmscout::ShapeFileVisitor
 {
 private:
+  static constexpr size_t                            maxCoastlinePoints=1000;
+
   osmscout::Progress                                 &progress;
   uint32_t                                           coastlineCount;
-  bool                                               continuation;
-  std::vector<osmscout::GeoCoord>                    coordBuffer;
   osmscout::Id                                       currentId;
 public:
   std::list<osmscout::WaterIndexProcessor::CoastRef> coasts;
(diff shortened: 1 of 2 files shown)


Framstag:
@Karry: I had difficulties, to get basemap rendering running in JavaScout. AI fiddled around a lot. This changes were part of the solution process. This - together with prioritized rendering in Javascout - solved the problem.

I'm not that deep in the code. Please review, if the fix is OK.

Note also, that there seems to be no file version 27 basemap on your server?

janbar:
About the basemap: In the osmin/master (map version 27) I updated manually the first byte of the file "types.dat".

```
$ hexdump -C -n 8 types.dat 
00000000  1a 00 00 00 00 0c 00 00
```

Changing 1a to 1b.

As the basemap has no features, I supposed it can be possible, and it avoid to rebuild the whole map database. But probably it is a bad option for later.

note: I know it is crappy ...

Karry:
I am using custom branch in my application where I maintain compatibility with older database formats. So, I had no need to re-generate basemap with 26 -> 27 update. There was added milestone feature in version 27, so it is safe to override types.dat as @janbar suggested, because basemap do not contain milestones :-) But in general, it is not good idea. I will generate new one...

Karry:
BasemapImport tool is just for generating water.dat from coastline shapefile. But the basemap that I was working recently is ordinary map, but imported from planet extract with limited set of types... See description here: https://github.com/Framstag/libosmscout/blob/master/stylesheets/basemap.ost#L7

Sorry for confusing you by using the same word "basemap".

Framstag:
So I can merge the change?

Karry (approved):
yes, the change looks fine.

Karry:
Btw, I just regenerated world "basemap": https://osmscout.karry.cz/basemap/basemap-260803-dist.tar.gz

PR #1769, opened 2026-08-18
feat: Added further natural types and styles

Framstag:


stylesheets/map.ost
```diff
@@ -797,6 +797,159 @@ TYPES
...
+    = NODE AREA ("natural"=="moor")
+      {Name, NameAlt}
+      OPTIMIZE_LOW_ZOOM MERGE_AREAS
+
+  TYPE natural_tundra
+    = NODE AREA ("natural"=="tundra")
+      {Name, NameAlt}
+      OPTIMIZE_LOW_ZOOM MERGE_AREAS
+
+  TYPE natural_coastline
```
Karry, line 832: just curious, what is the purpose to have coastlines in the database as there already is visible border of the sea/land from water index?

stylesheets/map.ost
```diff
@@ -797,6 +797,159 @@ TYPES
...
+
+  TYPE natural_hot_spring
+    = NODE ("natural"=="hot_spring")
+      {Name, NameAlt}
+
+  TYPE natural_geyser
+    = NODE ("natural"=="geyser")
+      {Name, NameAlt}
+
+  TYPE natural_waterfall
```
Karry, line 878: maybe merge with waterway_waterfall, "natural=waterfall" should not be used and it pretty rare...

stylesheets/map.ost
Framstag: That one can be dropped. KI missed the special handling

stylesheets/map.ost
Framstag: Being in holiday I stood at least before one waterfall that was tagged that way. But your are right, both types can get merged

PR #1782, opened 2026-08-28
Further types styles symbols and fixes

Framstag:
- Fix symbol rendering bugs in cairo and Qt
- New types and styles, and symbols for shop, leisure, historic, railway, aeroway, aerialway, military
- Improved existing symbols for existing types/styles


Karry:
When I see such fragmented types, did you test how it affect lookup performance and index size? As I remember, every type have its own lookup tiles in the index...

Framstag:
> When I see such fragmented types, did you test how it affect lookup performance and index size? As I remember, every type have its own lookup tiles in the index...

I tested on the desktop and it looked OK, but it requires further analysis - especially on mobile device. Inpact on database size is another topic.

Trigger are:
* Holiday, which shows that just some types are missing and would be helpful.
* POI/type search that would need differentiating types.

Since types and symbols are AI generated I tried to add all types with some relevance up to a certain limit which depends on the base type. If performance is too low for normal use, one has to solve it using thematic styles, which show only a subset of all types possible (perhaps higher use limit).

If it dos not work fast enough it would be a clear signal to look for new ways to performance improvements.

Framstag:
There are still some type groups missing, so still draft.

PR #1787, opened 2026-09-07
Improv: Improved type config resolution performance

Framstag:


Karry (approved):
I quickly review it, it looks reasonable. Just one question - how it speedup real country database? Does it worth it?

Also asked Claude code with "reviewer" skill for the review. It spend $21 and find just minor findings (suggests few asserts, minor allocation optimizations...).

Framstag:
> I quickly review it, it looks reasonable. Just one question - how it speedup real country database? Does it worth it?
> 
> Also asked Claude code with "reviewer" skill for the review. It spend $21 and find just minor findings (suggests few asserts, minor allocation optimizations...).

The speed improvement for import (initial steps) was visible and - based on KI response - should also be measureable.

I'm using deepseek-v4.1-flex with ollama cloud for 20$/Month. And yes, code looked reasonable for me, too. I'm currently triggering further improvements (target: increase fram rate), you are invited to take a look there, too :-)

PR #1790, opened 2026-09-12
navigation: detect a standing vehicle by net displacement, not a distance floor

Framstag:
Supersedes the earlier revision of this branch (Karry's review, 2026-09-16). The earlier approach zeroed
every segment shorter than 2 m, which is a **7.2 km/h speed floor in disguise** and deletes walking.

## Why the floor was wrong

The position-difference fallback only ever sees segments of at least one second
(`timestamp-lastPosition.time >= seconds(1)`), so a 2 m floor per segment is a 2 m/s = 7.2 km/h floor,
independent of the fix rate:

```
walking   3 km/h @1 Hz = 0.83 m per segment  ->  0 km/h   (deleted)
walking   5 km/h @1 Hz = 1.39 m per segment  ->  0 km/h   (deleted)
walking   7.2 km/h     = 2.00 m per segment  ->  kept
```

## What the phantom speed really is

A standing receiver's fix jitters around its true position, so consecutive fixes differ by about
the jitter amplitude - and the FIFO **sums** those differences (`fifoDistance += s.distance`), so
1 m of jitter per 1 s fix keeps reading as 3.6 km/h for as long as the vehicle stands still. Jitter
does not travel, it stays inside its own radius, while real movement accumulates.

## The fix: net displacement over a window

```
libosmscout/src/osmscout/navigation/SpeedAgent.cpp   +78 -20 (net +58 -20 with gate)
libosmscout/include/osmscout/navigation/SpeedAgent.h +3 -1
```

The stationary decision compares the **net** displacement between the oldest and the newest fix of
a 5 s window, and only when that history spans at least 4 s:

```
standing, 1 m fix jitter                       net <= 2 m        -> 0 km/h
walking at 3 km/h                              4.2 m in 5 s      -> 3 km/h reported
walking at 5 km/h                              6.9 m in 5 s      -> 5 km/h reported
```

Only the stationary case changes: when the vehicle moves, the reported speed is computed exactly as
before (same segment FIFO, same magnitude), so curves, stops and the existing 200 km/h sanity cap are
untouched. `recentFixes` is cleared together with the FIFO on a >10 s GPS gap and when the receiver
reports a speed below 0.5 m/s.

Known limit, documented at the gate: fix jitter above about 1.5 m amplitude is indistinguishable from
a very slow walker. A receiver that reports its own speed is still preferred and the gate is not
consulted then.

## Tests - the missing coverage that let the floor through

```
Tests/src/SpeedAgentTest.cpp   +239 (new)
Tests/CMakeLists.txt, Tests/meson.build  +13 (registration)
```

The agent had no test at all, so no test said what a walker must be reported as. The new test drives
the agent directly with deterministic fixes - a fixed 1 m jitter pattern rather than a random number
generator, so it does not depend on the standard library implementation:

| Test case | Asserts |
|---|---|
| does not report standstill GPS jitter as speed | a standing vehicle with 1 m jitter reports 0 km/h once the gate has its window (the plain fallback reports 3.6 km/h) |
| reports walking speed | 3, 4, 5 km/h at 1 Hz within 0.4 km/h; 4 km/h at 0.5 Hz as well |
| reports walking speed (jitter section) | 4 km/h with 1 m jitter does not collapse to standing |
| prefers the speed reported by the receiver | 4 m/s reported -> 14.4 km/h; 0 m/s reported -> 0 km/h |

Checked against mutated code, so the assertions really bind:

```
floor 0.0 m (gate never fires)  -> "does not report standstill GPS jitter" fails (6.63 km/h)
floor 6.0 m (gate eats walking) -> the three walking assertions fail
```

## Verification

```
cmake --build build                             clean
cd build && ctest -j 4 --output-on-failure      100% tests passed out of 125
meson test -C build-meson "Check speed agent"   1/1 OK
```

## Pre-existing issue found while writing the tests

Recorded in `TODO.md`, not fixed here: the fallback **never runs at fix rates above 1 Hz**, because
the "segment of at least a second" guard is compared against `lastPosition`, which is updated for
every fix - so the difference is always the interval between two consecutive fixes. Measured with the
new test driven at 4 Hz: the agent publishes no speed at all, for a moving vehicle as well as for a
standing one, while the same walk at 1 Hz reports 4 km/h.

### DIFF
--- a/libosmscout/src/osmscout/navigation/SpeedAgent.cpp
+++ b/libosmscout/src/osmscout/navigation/SpeedAgent.cpp
@@ -79,8 +79,16 @@ std::list<NavigationMessageRef> SpeedAgent::Process(const NavigationMessageRef &
           segmentFifo.clear();
         }
 
-        segmentFifo.push_back({GetEllipsoidalDistance(lastPosition.coord,gpsUpdateMsg->currentPosition),
-                               gpsUpdateMsg->timestamp-lastPosition.time});
+        // Standstill guard: GPS position jitter while the vehicle is stationary
+        // (typically 0.5-2 m per 1 s segment) must not be reported as speed.
+        // Segments below the displacement floor contribute zero distance, so
+        // drift-derived speed collapses to 0; real movement >= 2 m/s still
+        // computes normally (spec: gps-speed-priority).
+        Distance segmentDistance = GetEllipsoidalDistance(lastPosition.coord, gpsUpdateMsg->currentPosition);
+        if (segmentDistance.AsMeter() < 2.0) {
+          segmentDistance = Distance::Zero();
+        }
+        segmentFifo.push_back({segmentDistance, gpsUpdateMsg->timestamp-lastPosition.time});
         Timestamp::duration fifoDuration{Timestamp::duration::zero()};
         Distance fifoDistance;
         for (const auto &s:segmentFifo){


Karry (commented):
The gps jitter may really cause phantom speed reports, but it will discard walking speed (less than 7.2 km/h), right? Not sure if it is better than occasional speed jumps. 

The segment fifo should be solving exactly this. Speed is computed from three (of four) second window.

These SpeedAgent changes would need some testing in real conditions, but from the desk looks fishy.

Framstag:
Fixed - and you were right on both counts, with the number exact.

The fallback only ever sees segments of at least one second, so a 2 m floor of "segment distance" was a 7.2 km/h floor in disguise: walking at 3 km/h is 0.83 m per fix and read as 0.

The phantom speed has a different cause than a short segment. A standing receiver's fix jitters around its true position, so consecutive fixes differ by roughly the jitter amplitude - and the FIFO *sums* those differences, so 1 m per 1 s fix reads as 3.6 km/h for as long as the vehicle stands still. That is why the FIFO as written cannot tell the two apart: it measures path length, not displacement. The stationary decision now compares the *net* displacement between the oldest and the newest fix of a 5 s window:

    standing, 1 m jitter     net <= 2 m       -> 0 km/h
    walking at 3 km/h        4.2 m in 5 s     -> 3 km/h
    walking at 5 km/h        6.9 m in 5 s     -> 5 km/h

Only the stationary case changes; when moving, the reported speed is computed exactly as before. `Tests/src/SpeedAgentTest.cpp` now states the contract (0 km/h for jitter, 3/4/5 km/h within 0.4 km/h, at 1 Hz and 0.5 Hz) - the agent had no test at all, which is how the floor got through.

One more finding from writing those tests, recorded in `TODO.md`: the fallback never runs above 1 Hz, because the "segment of at least a second" guard is compared against `lastPosition`, which is updated for every fix. At 4 Hz the agent publishes no speed at all, for a moving vehicle as well as a standing one.

Framstag:
Branch brought up to date by **merging master into it** (no history rewrite), because it was `CONFLICTING`/`DIRTY` and could not be merged.

What the branch carries now:

- the two original commits unchanged - `SpeedAgent.h` and `SpeedAgent.cpp` are byte-identical to the version you reviewed;
- a new case in `Tests/src/SpeedAgentTest.cpp` asserting the boundary the gate cannot see through: movement that accumulates less displacement than the floor over the window (2 km/h at 1 Hz) is reported as standing. It documents the limit at the gate and fails if the floor is removed. **Revert-checked**: with the displacement floor set to zero, three cases fail with a jitter-derived speed (`6.63 == Approx(0.0)`); with the gate, 4 cases / 21 assertions pass;
- `openspec/changes/fix-navigation-standstill-gate/` - proposal, the new `navigation-speed-agent` capability (receiver speed preferred, position-difference fallback, the net-displacement decision and its documented limit), the design (which records the alternative you rejected - the per-segment distance floor as a 7.2 km/h floor in disguise - and why net displacement was chosen) and the verified task list. `openspec validate "fix-navigation-standstill-gate" --strict` passes.

The tests also uncovered a second defect in the same fallback, recorded in `TODO.md` and **not** part of this PR: a speed is published only when the interval since the previous fix spans a second, but that fix is updated for **every** accepted fix, so at 4 Hz the fallback never runs and the agent publishes no speed at all - moving or standing. Prepared as a stacked follow-up branch `fix-navigation-fallback-fix-rate`, which moves the minimum to the accumulated history and is revert-checked by 4 Hz and 10 Hz cases.

Verification: `SpeedAgentTest` 4 cases / 21 assertions; full suite `ctest -j 4 --exclude-regex PerformanceTest` 91/91 under `xvfb-run`; `meson test "Check speed agent"` passes; no compiler warnings from the touched files.

PR #1797, opened 2026-09-14
perf: reuse label measurements, glyphs and scratch storage of the label stage

Framstag:
The label stage is the allocation budget of a rendered frame: on a zoom 16 view of 16 tiles (Dortmund, Cairo) the five label steps spent 10344 of the frame's 10519 allocations per tile-frame (98%) and half of its frame time, and every frame measured every registered label again although a label's measurement depends only on its text, font, font size and wrapping, which stay the same while a view is panned, zoomed or redrawn. The same steps are where the label-dominated frame time of the z16-z17 band sits.

- remember a label measurement per measurement key (text, font size, proposed width, wrapping, contour flag and the four line wrapping parameters the backends read inside `Layout()` through `MapPainter::GetProposedLabelWidth`) in the label layouter, bounded by `defaultMeasurementCount` (4096) entries, and hold a label's glyph data in its measurement entry, so it is derived once and dropped with the label it was derived from
- let a backend declare the state its measurements depend on (font, resolution, magnification, drawing target settings) as a measurement environment once per frame; a change of that environment drops the remembered measurements. Cairo derives it from the font name, the font size factor, the DPI, the magnification, the font options and the device scale of its drawing target, Qt adds the DPI and the device pixel ratio of its painter device, the other backends (Skia, SVG, AGG, GDI, DirectX, iOS) use the shared `BuildMeasurementEnvironment` prefix
- reuse the frame's label state (`LayoutJob` with its three overlap canvases, the `ScreenMask`/`ScreenRectMask` bitmasks, the per-instance mask and canvas stores) instead of allocating it per frame, and build the resolved label instances directly in the output store, without the visible-element list and the instance copy
- reuse the per-object label list of `LayoutPointLabels`, the `LabelPath` of a path label (`LabelPath::Clear`) and the grid points of a shielded way (sorted unique buffer instead of a `std::set` per way)
- add `MapPainterLabelReuseTest`: measurement and glyph reuse, reuse independent of drawing, changed measurement inputs and environment, unchanged label set/placement/order, constant allocation per frame, honoured cache bound; plus cases for `ScreenMask`/`ScreenRectMask` reset, `LabelPath` clear and the dependence of the Cairo measurement on the projection resolution

Verified: cmake build clean with no warning in the changed files, meson build warning-free and its three target tests pass, ctest 118/118 passed, the label placement dump of the fixed view is byte-identical to the pre-change dump (36 entries, 0 diff), the new test file is uncrustify clean and the changed library files are at or below the uncrustify drift they had before the change, clang-tidy without a new finding category. Measured on the fixed view (Dortmund, 1 tile zoom 15, Cairo): allocations per frame 4637 -> 1740, `DrawLabels` 2006 -> 361, `DrawWayContourLabels` 777 -> 51, `CalculateWayShields` 860 -> 286, `PrepareNodeLabels` 651 -> 173, `PrepareAreaLabels` 150 -> 41; on the zoom 16 view 10537 -> 4486 allocations per tile-frame.

Review fixes (`d79471650`, in response to the review below):

- **`LabelLayouter.h:1134` - glyphs were kept in a table keyed by the address of a label that only the measurement cache kept alive.** With the reuse switched off (bound 0) the table grew without bound *and* could serve the glyphs of an earlier label to a label that the allocator placed at the same address. The glyph data is a member of the measurement entry now (`LabelMeasurement::glyphs` with its derivation flag), the second table is gone, and a layouter without a cache holds the measurement of the current step in one slot that `Reset()` empties.
- **`LabelLayouter.h:254` - the key missed the line wrapping parameters.** `GetProposedLabelWidth` reads `LabelLineMinCharCount`, `LabelLineMaxCharCount`, `LabelLineFitToArea` and `LabelLineFitToWidth` *inside* `Layout()` (Qt `MapPainterQt.cpp:431`, Skia `:814`, iOS `:383`). All four are part of the key now, filled from the `MapParameter` the layouter already receives. The other backends use only `objectWidth` and the font state that the measurement environment covers.
- **`LabelLayouter.h:1089` - eviction followed the order of insertion.** The order list holds the order of use now and a hit splices its entry to the back (`std::list::splice`, without an allocation), so a working set above the bound does not thrash.
- `MapPainterLabelReuseTest` gains four cases for these (LRU, changed wrapping parameter, disabled reuse, glyph ownership); each fix fails a new case when it is reverted. ctest 118/118 passed again, uncrustify drift of `LabelLayouter.h` 284 -> 273 lines, and an A/B of the fixed view measures 1740 -> 1738 allocations per frame with the frame time inside the spread of the run. The placement dump and the per-step probe buckets of the numbers above were not re-run, because the probe was a temporary instrument that is not part of the tree; `verification.md`, section 14, records what was.

Notes for the reviewer:

- AGG, GDI, DirectX and the iOS backend are not part of the local CMake configuration; their environment call follows the same pattern as the compiled backends and is verified by the msys/vs2025/ios CI jobs.
- The clang-tidy report of the new test file carries more `avoid-magic-numbers` and `pro-bounds-avoid-unchecked-container-access` findings than the existing painter tests, in the same check categories; `BuildMeasurementEnvironment` carries one `bugprone-easily-swappable-parameters` (two adjacent `double` parameters).
- `TODO.md` gains the GDI device-resolution gap, the `MapPainterCairo::fonts` cache with its incomplete key, the remaining two pointer vectors of the layouter's draw step, the second copy of the measurement key in the cache's order list, and a correction: the allocations of the `DrawLabels` step were the layouter's overlap resolution, not the symbol drawing.
- The branch was rebased onto `master` (`9f799a819`, the pango and area-culling merges). The conflicts in `TODO.md`, `Tests/CMakeLists.txt` and `Tests/meson.build` (both sides added a test registration, and a TODO section) are resolved by keeping both sides; the full suite was re-run afterwards (121/121) and is recorded in `verification.md`, section 14.
- The numbers of this description are from desktop x86_64 (gcc, Release) with the Cairo backend; no mobile device is available in this environment.

openspec: add map-painter-label-reuse


libosmscout-map/include/osmscoutmap/LabelLayouter.h
```diff
@@ -928,8 +1036,108 @@ constexpr bool debugLabelLayouter = false;
...
+     */
+    const std::vector<Glyph<NativeGlyph>>& GetLabelGlyphs(const LabelPtr& label)
+    {
+      auto entry=glyphs.find(label.get());
+
+      if (entry!=glyphs.end()) {
+        return entry->second;
+      }
+
+      auto inserted=glyphs.emplace(label.get(),
```
Karry, line 1134: this insert is never flushed, when `maxMeasurementCount==0`

libosmscout-map/include/osmscoutmap/LabelLayouter.h
```diff
@@ -236,6 +240,44 @@ constexpr bool debugLabelLayouter = false;
...
+   * with the same key measure the same, so the second one can reuse the first measurement.
+   */
+  class LabelMeasurementKey
+  {
+  public:
+    std::string text;
+    double      fontSize{0.0};
+    double      objectWidth{0.0};
+    bool        enableWrapping{false};
+    bool        contourLabel{false};
```
Karry, line 254: minor, but the key is missing few parameters, that may affect layouting: LabelLineMinCharCount/MaxCharCount/FitToWidth/FitToArea

libosmscout-map/include/osmscoutmap/LabelLayouter.h
```diff
@@ -928,8 +1036,108 @@ constexpr bool debugLabelLayouter = false;
...
+      key.text=text;
+      key.fontSize=fontSize;
+      key.objectWidth=objectWidth;
+      key.enableWrapping=enableWrapping;
+      key.contourLabel=contourLabel;
+
+      auto entry=measurements.find(key);
+
+      if (entry!=measurements.end()) {
+        return entry->second;
```
Karry, line 1089: when the entry is used, should not be moved to the front of `measurementOrder` to behave as LRU ?

Karry (changes requested):
TLDR ;-) just make a fast review by eyes and like the idea. I would really like to see some benchmark numbers, how it improves rendering on the mobile device. 

After manual review, I spend few $$ for AI review and it found few problems...

libosmscout-map/include/osmscoutmap/LabelLayouter.h
Framstag: You were right, and it was worse than a leak. The table was keyed by the address of a label that only the measurement cache kept alive, so with `maxMeasurementCount==0` nothing ever took an entry out of it: it grew by one entry per label and frame, and the glyphs it held could be served to a *different* label - a later label placed at the same address by the allocator returns the earlier label's glyph data.

Fixed in `c5eeb8302`: the glyph data is a member of the measurement entry now (`LabelMeasurement::glyphs` plus the flag that says whether it has been derived), the pointer-keyed table is gone, and nothing can be keyed on a label the layouter does not own. A layouter without a cache keeps the measurement of its current step in one slot, `notRememberedMeasurement`, which `Reset()` empties at the end of the frame, so nothing at all survives the frame in that mode.

Two new cases cover it: "A bound of 0 remembers neither measurements nor glyph data" (one measurement and one glyph derivation per frame, no remembered glyph data) and "Glyph data does not outlive the measurement it belongs to". Both compare the frame's contour labels with those of a fresh layouter, so they do not depend on which address the allocator happens to hand out.

libosmscout-map/include/osmscoutmap/LabelLayouter.h
Framstag: Right, and thank you - a stale measurement of a different wrapping configuration is exactly what it was.

I audited every `Layout()` implementation for inputs beyond its arguments. The ones that read those parameters are Qt (`MapPainterQt.cpp:431`), Skia (`:814`) and iOS (`:383`), all three through `MapPainter::GetProposedLabelWidth`, which is where `LabelLineMinCharCount`, `LabelLineMaxCharCount`, `LabelLineFitToArea` and `LabelLineFitToWidth` are read. The other backends (Cairo, AGG, SVG, GDI, DirectX) use `objectWidth` alone, and the font name, font size factor, DPI and magnification they also read are covered by the measurement environment.

Fixed in `c5eeb8302`: all four are part of `LabelMeasurementKey` now (hash and equality included) and are filled from the `MapParameter` the layouter already receives, so no backend has to report them and the key stays complete for a backend that starts reading another one of them. I kept them in the key rather than in the environment, because a client that toggles a wrapping setting keeps both variants reusable instead of dropping every measurement.

New case: "A changed line wrapping parameter is measured again" - four parameter changes, one measurement each, and a repeated frame in between that measures nothing. I verified it is load bearing by removing the four fields from the key again: the case fails.

libosmscout-map/include/osmscoutmap/LabelLayouter.h
Framstag: Yes, it should - my reasoning was the O(1) cost of the touch, but that was wrong here: the order list and the hash lookup already exist, so least-recently-used costs one `std::list::splice` per hit and nothing per miss (no allocation, no search of the list).

Fixed in `c5eeb8302`: each entry knows its position in the order list, a hit splices it to the back (`std::list::splice(measurementOrder.end(), measurementOrder, entry->second.order)`), and eviction takes the front. So a label that every frame uses is not dropped once the bound is reached, which is what would have happened to a working set above 4096 measurements.

New case: "A measurement that is used again outlives the one that is not" - with a bound of 2, a frame that uses A and C after A and B must still have A remembered in the next frame, so that frame performs one measurement instead of two. Verified load bearing by disabling the splice: the case fails at `REQUIRE(fake.measurementCount==4)`.

Framstag:
All three findings of the review are fixed in `c5eeb8302` (replies on the individual comments), each with a test case that fails when the fix is reverted, and `verification.md` section 14 records the evidence. Summary: the measurement key carries the four line wrapping parameters now, the glyph data is a member of the measurement entry instead of a second table keyed by a label address, and eviction drops the least recently used measurement instead of the oldest measured one.

On the benchmarks: I cannot answer the mobile question from here - this environment has no device or emulator, so all numbers in the description are desktop x86_64 gcc Release with the Cairo backend (single-threaded, one tile at zoom 15 and 16 tiles at zoom 16). A mobile measurement needs an Android or iOS build of the client with a fixed view and a fixed number of repeated draws; if you tell me which device and which view would be representative, or hand me a device, I will produce the same table for it. What suggests the effect carries over is that the removed work is per registered label and per frame (shaping, glyph derivation, scratch allocation), not per pixel, and the remaining label work is the drawing itself.

The AI review was worth its money - two of the three findings were real defects that my own test suite did not cover, and the third was right about the eviction policy. Both defects came from the same mistake on my side: I keyed or bounded state on assumptions about ownership and about which parameters a backend reads, instead of on the arguments the layouter is handed. The fixes move that state into the measurement entry and enumerate the parameters explicitly; the four new cases make the assumptions explicit.

One note for the merge: the branch is behind `master` and conflicts in `MapPainter.cpp`, `Tests/CMakeLists.txt`, `Tests/meson.build`, `Tests/src/TextMetricsCairoTest.cpp` and `TODO.md` (from the pango and area-culling merges). I will rebase next and re-run the full suite afterwards.

Framstag:
Hi @Karry , I let the AI reply ;-)

Framstag:
The branch is rebased onto `master` (`9f799a819`, the pango and area-culling merges), so the two commits are now `fd437f680` (the change) and `d79471650` (the review fixes) - the hashes in my replies above refer to the pre-rebase commits and to the same content.

Conflicts: `TODO.md`, `Tests/CMakeLists.txt` and `Tests/meson.build` (both sides added a test registration; master added a TODO section), plus `MapPainter.cpp` and `Tests/src/TextMetricsCairoTest.cpp`, which merged on their own. The two build files keep both registrations, `TODO.md` keeps the rewritten list of this branch plus the `map-painter-area-visibility-cull` section from master.

Re-verified after the rebase: `cmake --build` without a warning in the changed files, `ctest -j 4` 121/121 passed, meson reconfigured and `meson test "Check MapPainterLabelReuse compilation"` OK, uncrustify drift unchanged (273 lines for `LabelLayouter.h`, 0 for the test file). Recorded in `verification.md`, section 14.

PR #1815, opened 2026-09-19
Propagate lane suggestions backward across preceding junctions

Karry:
Another round of improvements for suggested lanes heuristics. It started with the wrong suggestion on the real junction where I was riding the bike :-) I split ground-truth routes to smaller segments for better reproducibility, then found few another issues and let AI to fix them...

As a next step, I want to add lane suggestions to spoken instructions...


Framstag:
I'll merge as soon as build succeeds. Using https://github.com/Framstag/NaviVeylin I now have a way t test things (more or less directly) in real live (as often as I drive) under Android Phone and Android Auto (and in principle Android Automobile).

PR #1828, opened 2026-09-21
fix: keep the active style when a stylesheet fails to load

Framstag:
## What

A stylesheet that cannot be loaded no longer reaches the map painter.

**Bug:** `DBInstance::LoadStyle` discarded the candidate configuration on a parse error and left the database with *no* style configuration. Only the explicit runtime switch (`loadStyleSheet`) noticed the failure and restored the previous style; the session-start load, a style-flag change, the basemap stylesheet and the stylesheet refresh did not. A malformed bundled stylesheet therefore left the painter running without a configuration and the process died with `Fatal signal 11 (SIGSEGV)` in the style lookup, preceded only by a warning.

## How

- Adopt a candidate configuration **only after a clean parse**; on a rejection the configuration installed before the attempt stays in effect.
- Install the already existing empty configuration (`emptyStyleConfig`, previously constructed and never used) for a database whose stylesheet has never loaded successfully, so every database always has a usable configuration. A later success replaces it, a later failure keeps the working one.
- Report the outcome: `DBThread::WasLastStyleLoadSuccessful()` / `GetActiveStyleSheetFilename()`, the JNI entry point `wasLastStyleLoadSuccessful()`, and `getActiveStyleSheet()` now returns the stylesheet that is actually installed.
- `loadStyleSheet()` decides on the new flag instead of comparing the error count before and after.
- The basemap follows the same rule: a rejected basemap stylesheet drops the basemap layer and keeps the map rendering.
- The render batch keeps its `GetStyleConfig()` guard as a documented safety net and no longer silently skips a database without saying which one.

## BREAKING

None: no API is removed or renamed, and no successful load path changes behaviour.

Both signature changes are additive and source-compatible:

- `DBInstance::LoadStyle(..., std::list<StyleError> &errors)` → gains a **defaulted** `const osmscout::StyleConfigRef &fallback=nullptr` parameter.
- `DBThread` gains two read-locked accessors, `GetActiveStyleSheetFilename()` and `WasLastStyleLoadSuccessful()`.

## Tests

- New `Tests/src/StyleLoadResilienceTest.cpp` (4 cases): a rejected stylesheet keeps the previously active configuration; a first failed load installs the fallback; valid → rejected → valid recovers; a batch with one fallback-configured and one healthy database paints without a fault.
- Registered in both build systems (CMake + Meson).
- Revert check: mutating `if (!styleConfig && fallback)` to `if (fallback)` makes the first case fail (`GetStyleConfig() == active`, 1 failed / 35 assertions).
- CMake `ctest -j4`: 125/126 (the one failure, `MapPainterAreaVisibilityCullTest`, is pre-existing on `master` and is what PR #1825 fixes; this change touches no file under `libosmscout-map/`).
- Meson `meson test`: 126/126. Builds clean in both configurations, no new warnings.

## Spec

OpenSpec change `openspec/changes/client-style-load-resilience/` (proposal, spec delta for `client-java-style-switching`, design, tasks). The spec generalizes the existing "switching to an unloadable stylesheet fails" contract to every load path and adds "a database always has a usable style configuration" and "rendering never uses a style configuration from a failed load".


Karry (approved):
Not sure if the increased complexity worth it. In my application I added stylesheet checks with [OSTAndOSSTest](https://github.com/Framstag/libosmscout/blob/master/Tests/src/OSTAndOSSTest.cpp) as part of the CI/CD pipeline, so application package should always contain valid stylesheets...

PR #1831, opened 2026-09-21
Fix contour label rendering in MapPainterIOS

vyskocil:
Two independent defects in the iOS painter made contour labels render wrong characters and made the label layouter work in the wrong coordinate space.

DrawGlyphs() took the CTRun and its font from glyphs[0] and indexed the run with the glyph's position in the vector it was given. Neither holds: a label may be laid out as several runs, and the layouter is free to drop glyphs that fall outside the layout viewport, so the nth entry of the vector is not the nth glyph of the run. The painter therefore drew the first n characters of the name instead of the ones that had been kept. Each glyph already carries its own run and its index within it, so use those.

BeforeDrawingCallback() passed CGBitmapContextGetWidth()/GetHeight() to SetViewport(). Those return the backing store in device pixels, while glyph positions, the projection and every other length the layouter works with are in the context's user space. On a retina context the layouter was told the drawn area was 2 or 3 times wider and taller than it is, depending on the device. Use projection.GetWidth()/GetHeight(), which is what the agg, gdi and cairo painters already do.


Framstag:
I have knowwhere to test this. The Mac Book now has declining support. So I merge everyting you have tested :-) I wait for rebase to see if the build is fixed than will merge.

